PR #15, opened 2020-02-20
add detail comment (part 2)

severinbeauvais:
*Issue #:* /bcgov/entity#2719

*Description of changes:*
- added Add Comment Dialog unit tests
- added Detail Comment unit tests
- updated Staff Payment unit tests
- refactored Directors unit tests to not use setTimeout()
- fixed filing save tests (missing 'get tasks' stub)
- cleaned up other unit tests
- added Detail Comment autofocus prop
- ~~display app version on startup to console~~ deleted
- removed obsolete hard-coded entity type
- ensured comments are sorted by timestamp
- reload just the new comments, not the whole dashboard

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -64,6 +64,7 @@
                 </template>
                 <v-list dense>
                   <v-list-item-group color="primary">
+                    <!-- NB: this menu item is disabled in current release -->
                     <v-list-item disabled>
                       <v-list-item-icon>
                         <v-icon>mdi-file-document-edit-outline</v-icon>
@@ -162,8 +163,8 @@
                 v-if="isRoleStaff"
                 @click.stop="showCommentDialog(item.filingId)"
               >
-                  <span>Add Detail</span>
-                </v-btn>
+                <span>Add Detail</span>
+              </v-btn>
             </div>
             <div>
               <!-- the details list-->
@@ -344,7 +345,7 @@ export default {
             }],
             paperOnly: false,
             isCorrected: filing.header.isCorrected || false,
-            comments: this.flattenComments(filing.header.comments)
+            comments: this.flattenAndSortComments(filing.header.comments)
           }
           this.filedItems.push(item)
         } else {
@@ -388,7 +389,7 @@ export default {
           status: filing.header.status,
           paperOnly: false,
           isCorrected: filing.header.isCorrected || false,
-          comments: this.flattenComments(filing.header.comments)
+          comments: this.flattenAndSortComments(filing.header.comments)
         }
         this.filedItems.push(item)
       } else {
@@ -429,13 +430,20 @@ export default {
         }],
         paperOnly: true,
         isCorrected: filing.header.isCorrected || false,
-        comments: this.flattenComments(filing.header.comments)
+        comments: this.flattenAndSortComments(filing.header.comments)
       }
       this.filedItems.push(item)
     },
 
-    flattenComments (comments: any): Array<any> {
-      return (comments && comments.length > 0) ? comments.map(c => c.comment) : []
+    flattenAndSortComments (comments: any): Array<any> {
+      if (comments && comments.length > 0) {
+        // first use map to change comment.comment to comment
+        const flattened: Array<any> = comments.map(c => c.comment)
+        // then sort newest to oldest
+        const sorted = flattened.sort((a, b) => new Date(a.timestamp) < new Date(b.timestamp) ? 1 : -1)
+        return sorted
+      }
+      return []
     },
 
     typeToTitle (type: string, agmYear: string = null): string {
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -201,7 +201,7 @@
 <script lang="ts">
 // Libraries
 import axios from '@/axios-auth'
-import { mapGetters, mapState, mapActions } from 'vuex'
+import { mapGetters, mapState } from 'vuex'
 
 // Dialogs
 import { AddCommentDialog, DownloadErrorDialog } from '@/components/dialogs'
@@ -254,8 +254,6 @@ export default {
   },
 
   methods: {
-    ...mapActions(['setTriggerDashboardReload']),
-
     loadData () {
       this.filedItems = []
 
@@ -314,7 +312,7 @@ export default {
 
     // Method to extract date from a local datetime string
     // Returns "yyyy-mm-dd"
-    formatDate (dateString): string {
+    formatDate (dateString: string): string {
       if (!dateString) return null // safety check
       return dateString.split(' ')[0]
     },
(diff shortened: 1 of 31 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -64,6 +64,7 @@
                 </template>
                 <v-list dense>
                   <v-list-item-group color="primary">
+                    <!-- NB: this menu item is disabled in current release -->
                     <v-list-item disabled>
```
cameron-freshworks: Was this partly why you  were inquiring about Launch darkly??

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -429,13 +428,20 @@ export default {
...
         isCorrected: filing.header.isCorrected || false,
-        comments: this.flattenComments(filing.header.comments)
+        comments: this.flattenAndSortComments(filing.header.comments)
       }
       this.filedItems.push(item)
     },
 
-    flattenComments (comments: any): Array<any> {
-      return (comments && comments.length > 0) ? comments.map(c => c.comment) : []
+    flattenAndSortComments (comments: any): Array<any> {
```
cameron-freshworks: Are the comments also an array?

src/utils/sleep.ts
```diff
@@ -1,10 +1,8 @@
...
-  sleep (ms: number): Promise<void> {
-    return new Promise(resolve => setTimeout(resolve, ms))
-  }
+/**
+ * Function that "sleeps" for specified timeout. Must be called from async method.
+ * @param ms Delay to sleep, in milliseconds.
+ * @returns A promise to await upon.
+ */
+export function sleep (ms: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, ms))
```
cameron-freshworks: Just curious here, as far as i can tell this is just used in jest spec files. Did you not solve a similar issue with `flushPromises()` or was something unrelated?

src/views/Correction.vue
```diff
@@ -606,9 +606,7 @@ export default {
         }
       }
       if (setting === 'add' && !added) {
-        // TODO: remove hard-coded entity type when fee codes are available from Pay team
-        // this.filingData.push({ filingTypeCode: filing, entityType: this.entityType })
-        this.filingData.push({ filingTypeCode: filing, entityType: EntityTypes.BCOMP })
+        this.filingData.push({ filingTypeCode: filing, entityType: this.entityType })
```
cameron-freshworks: Nice!

src/views/Correction.vue
```diff
@@ -329,7 +329,7 @@ export default {
       }
     }
     // NB: this is the id of the filing to correct
-    const origFilingId = this.$route.params.id
+    const origFilingId = +this.$route.params.id // number
```
cameron-freshworks: What does the prefixed + do again? Cast it to a specific type?

tests/unit/Directors.spec.ts
```diff
@@ -736,152 +739,145 @@ describe('Appoint New Director tests', () => {
...
 
-    const Constructor = Vue.extend(Directors)
-    const instance = new Constructor({ store, vuetify })
-    vm = instance.$mount()
-
-    // set props
-    vm.asOfDate = '2019-04-01'
-    vm.componentEnabled = true
+    wrapper = mount(Directors, {
+      sync: false,
```
cameron-freshworks: What does this sync property do in this case?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Yes, this would have been a good candidate for LD. Instead I just disabled it for the last release.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Maybe, or they could be null. I didn't want to go crazy with typing in this component because it's not fully TypeScript yet.

src/utils/sleep.ts
severinbeauvais: Yes, and I do use it in the unit test to wait for the debounce interval to pass.

Other than this, this was a pretty handy function for prototype actions (like save), as we did in the Business Create UI, remember?

src/views/Correction.vue
severinbeauvais: Yes, the Correction filing code is valid for both BCOMP and COOP (I checked).

src/views/Correction.vue
severinbeauvais: It's called the unary plus operator. It tries to convert the operand into a number (or NaN).
https://www.google.com/search?q=javascript+unary+operator

tests/unit/Directors.spec.ts
severinbeauvais: It's a test framework "workaround" to allow things to happen synchronously, but it's admittedly "dangerous" because we're still using a beta version of the test-utils (and I think things break if we upgrade to the latest version, sigh).

Ref: https://github.com/vuejs/vue-test-utils/issues/1137

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Definitely, i would like to work with LD a bit as well!

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: 100%, totally understand, every little bit helps!

src/utils/sleep.ts
cameron-freshworks: I do! It is very handy indeed, i'm not opposed to it whatsoever 👍

src/views/Correction.vue
cameron-freshworks: ahhh right, thanks :) This is at least the second time i've asked you now, sorry and thanks!

tests/unit/Directors.spec.ts
cameron-freshworks: Very interesting, thanks for clarifying!

src/components/Dashboard/FilingHistoryList.vue
thorwolpert: @rarmitag or @JohnamLane can setup the LD Flag for you. Could you work with Richard to do this?

cameron-freshworks: Will discuss with them, thanks!

src/utils/config-helper.ts
```diff
@@ -7,6 +7,9 @@ export default {
    * fetch config from environment and API
    */
   fetchConfig () {
+    const { name, version } = require('../../package.json')
```
thorwolpert: Does this work in build time?
Something like JSON.parse(unescape(process.env.PACKAGE_JSON || '%7Bversion%3A0%7D')).version
there is also some concern about loading all of the package.json
you could add something like this to vue.config.js
const fs = require('fs')
const packageJson = fs.readFileSync('./package.json')
const version = JSON.parse(packageJson).version || 0
module.exports = {
    configureWebpack: {
        plugins: [
            new webpack.DefinePlugin({
                'process.env': {
                    PACKAGE_VERSION: '"' + version + '"'
                }
            })
        ]
    }
}
Which doesn't expose the package.json in the built applicatioin

severinbeauvais: As discussed, I have removed the subject code and we can revisit this in /bcgov/entity#2555.

PR #21, opened 2020-02-24
fixed code smells

severinbeauvais:
*Issue #:* /bcgov/entity#2719

- "vm" was already declared in the upper scope so I just created a new describe section for the subject tests


cameron-freshworks:
Just curious about the SonarCloud checks. How heavily do we weigh this against files such as Spec files? Does a failing test mean a failing PR?

severinbeauvais:
> Just curious about the SonarCloud checks. How heavily do we weigh this against files such as Spec files? Does a failing test mean a failing PR?

We should discuss this as a team. IMHO we should try to fix SC issues but with lots of exceptions. For example, SC told me "::v-deep" was invalid, which it isn't in Vue. Also, SC reports a lot of duplicated code, which I don't want to worry about atm.

PR #22, opened 2020-02-26
correction filing (part 1)

severinbeauvais:
*Issue #:* /bcgov/entity#2654

*Description of changes:*
This is Part 1 of the the Correction Filing task, which includes:
- moved shared components to common folder
- enabled correction filing in Filing History List menu
- updated Correction filing page
- updated Detail Comment to support maxLength prop and to use placeholder instead of label
- updated Staff Payment component (with Priority and No Fee)
- replaced &lt;b&gt; with &lt;strong&gt; per SonarCloud

Part 2 will implement:
- updates to Summary Staff Payment
- updates to other filings (re: StaffPayment)
- updates to unit tests


src/components/common/Certify.vue
```diff
@@ -45,37 +45,35 @@ import { mapState } from 'vuex'
...
   private message: string
 
+  /** Entity Display prop. */
   @Prop({ default: '' })
   private entityDisplay: string
 
-  /**
-   * Called when component is created.
-   */
+  /** Called when component is created. */
```
cameron-freshworks: Don't mind this at all for one line DocBlock Comments

src/components/common/StaffPayment.vue
```diff
@@ -0,0 +1,152 @@
...
+  private isWaiveFees: boolean
+
+  /** Form model property. */
+  private formValid: boolean = false
+
+  /** Array of validation rules for Routing Slip Number. */
+  private readonly rules = [
+    v => !!v || 'Routing Slip Number is required',
+    v => /^\d{9}$/.test(v) || 'Routing Slip Number must be 9 digits'
+  ]
```
cameron-freshworks: Do these rules exist elsewhere as well? I can't recall?

src/components/common/Certify.vue
severinbeauvais: Yeah, they display the same when you mouse over the property/method.

src/components/common/StaffPayment.vue
severinbeauvais: Yes, different components have different rules for different properties. Also the validation error messages are different for different fields.

src/components/common/StaffPayment.vue
```diff
@@ -0,0 +1,152 @@
+<template>
```
cameron-freshworks: Quick Question about this component. Is this a new feature for staff on all maintenance filings or is this specific to just Corrections?

src/components/common/DetailComment.vue
```diff
@@ -39,27 +37,36 @@ export default class DetailComment extends Vue {
...
+    // inform parent of initial validity
+    this.emitValid(this.value)
+  }
+
   /**
    * Called when prop changes (ie, v-model is updated by parent).
    * This method is debounced to prevent excessive validation.
    */
   @Watch('value')
   @Debounce(300)
```
cameron-freshworks: I know this was not implemented in this PR, but this makes me think. Should we have implemented the Debounce decorator on all our listeners? Just a little bit of reading on the topic makes me think it is certainly not a bad thing when trying to optimize performance. No Action required here, just starting a conversation :)

src/components/common/StaffPayment.vue
cameron-freshworks: Sorry, i got confused. This was a relocation of a component from AR to common. Not something i had worked with prior, i see now this is for all filing types. Resolved!

src/components/common/StaffPayment.vue
cameron-freshworks: Ahh i see, yes the msg's are closely tied to the rules and are very specific. Makes sense they are where they are!

src/components/common/StaffPayment.vue
severinbeauvais: The enhancements to Staff Payment apply to all filings... I am working on that in Part 2!

src/components/common/DetailComment.vue
severinbeauvais: I implemented it here because it needed it (it validated a bunch of rules on every keystroke).

I think we should debounce when needed, and I haven't found any other candidates. Do you have something in mind?

cameron-freshworks: No nothing specific, i just know that some of the watchers are triggered constantly and was just thinking ( theoretically ) this is a cool way of optimizing. It's just something i'll keep in mind going forward, if a watcher is a good candidate for Debouncing.

PR #27, opened 2020-02-27
correction filing (part 3)

severinbeauvais:
*Issue #:* /bcgov/entity#2654

*Description of changes:*
This is Part 3 of the the Correction Filing task, which includes:
- implemented loading draft correction
- renamed some route params to correctedFilingId
- blocked correcting a correction (for now)
- enabled Save for corrections
- changed Save buttons to always be enabled (in all filings, per Scott)
- changed some strings to Filing Types
- added some explicit types

Part 4 will implement:
- updates to unit tests

### DIFF
--- a/src/views/AnnualReport.vue
+++ b/src/views/AnnualReport.vue
@@ -208,7 +208,7 @@
       <div class="buttons-left">
         <v-btn id="ar-save-btn" large
           v-if="isAnnualReportEditable"
-          :disabled="!isSaveButtonEnabled || busySaving"
+          :disabled="busySaving"
           :loading="saving"
           @click="onClickSave()"
         >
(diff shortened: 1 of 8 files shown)


src/views/AnnualReport.vue
```diff
@@ -208,15 +208,15 @@
       <div class="buttons-left">
         <v-btn id="ar-save-btn" large
           v-if="isAnnualReportEditable"
-          :disabled="!isSaveButtonEnabled || busySaving"
```
cameron-freshworks: Just curious how this ties in here?

severinbeauvais: I had to think about this for Correction filing so I asked Scott what the requirement is and whether it applies to all filings. He said you should always be allows to save a draft, for any filing. So I updated it.

severinbeauvais: I will be sure to add test notes to the ticket so that QA knows to look at it.

thorwolpert:
Will merge with the unit test temporarily failing.

PR #30, opened 2020-02-28
correction filing (part 4)

severinbeauvais:
*Issue #:* /bcgov/entity#2654

*Description of changes:*
This is Part 4 of the Correction Filing task, which inclues:
- moved some methods to Filing Mixin
- moved Filing Data to global store
- init Filing Data on filing creation
- expanded use of types
- updated version number

Part 5 will implement:
- updates to Summary Staff Filing component
- updates to unit tests

### DIFF
--- a/src/mixins/entityFilter-mixin.ts
+++ b/src/mixins/entityFilter-mixin.ts
@@ -1,5 +1,6 @@
 import { Component, Vue } from 'vue-property-decorator'
 import { mapState } from 'vuex'
+import { EntityTypes } from '@/enums'
 
 /**
  * Mixin that provides an entity filter utility.
@@ -10,15 +11,15 @@ import { mapState } from 'vuex'
   }
 })
 export default class EntityFilterMixin extends Vue {
-  readonly entityType: string
+  readonly entityType!: EntityTypes
 
   /**
    * Compares the conditional entity to the entityType defined from the Store.
    *
-   * @param entity The entity type of the component.
-   * @return boolean A boolean indicating if the entityType given matches the entityType assigned to the component.
+   * @param entity the entity type of the component
+   * @return True if the entityType given matches the entityType assigned to the component, else False
    */
-  entityFilter (entityType: string): boolean {
+  entityFilter (entityType: EntityTypes): boolean {
     return this.entityType === entityType
   }
 }
--- a/src/mixins/filing-mixin.ts
+++ b/src/mixins/filing-mixin.ts
@@ -1,0 +1,111 @@
+import { Component, Vue } from 'vue-property-decorator'
+import { mapActions, mapState } from 'vuex'
+import { FilingData } from '@/interfaces'
+import { EntityTypes, FilingCodes, FilingTypes, FilingNames } from '@/enums'
+
+/**
+ * Mixin that provides some useful filing utilities.
+ */
+@Component({
+  computed: {
+    ...mapState(['filingData', 'entityType'])
+  },
+  methods: {
+    ...mapActions(['setFilingData'])
+  }
+})
+export default class FilingMixin extends Vue {
+  // store actions
+  readonly setFilingData!: (x: any) => void
+
+  // store getters
+  readonly filingData!: Array<FilingData>
+  readonly entityType!: EntityTypes
+
+  /**
+   * Flattens and sorts an array of comments.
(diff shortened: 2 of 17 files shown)


src/mixins/entityFilter-mixin.ts
```diff
@@ -10,15 +11,15 @@ import { mapState } from 'vuex'
...
   /**
    * Compares the conditional entity to the entityType defined from the Store.
    *
-   * @param entity The entity type of the component.
-   * @return boolean A boolean indicating if the entityType given matches the entityType assigned to the component.
+   * @param entity the entity type of the component
+   * @return True if the entityType given matches the entityType assigned to the component, else False
    */
-  entityFilter (entityType: string): boolean {
+  entityFilter (entityType: EntityTypes): boolean {
```
cameron-freshworks: Nice!

src/mixins/filing-mixin.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+@Component({
+  computed: {
+    ...mapState(['filingData', 'entityType'])
+  },
+  methods: {
+    ...mapActions(['setFilingData'])
+  }
+})
+export default class FilingMixin extends Vue {
```
cameron-freshworks: Well done, i considered doing this as well but perhaps i was feeling the time crunch :(. 
Good job, thanks for doing this!

src/views/Correction.vue
```diff
@@ -331,20 +330,20 @@ export default {
     }
 
     // this is the id of THIS correction filing
-    // if falsy, this is a new correction filing
+    // if 0, this is a new correction filing
     // otherwise it's a draft correction filing
-    this.filingId = +this.$route.params.id // number (may be NaN)
+    this.filingId = +this.$route.params.id || 0 // number
```
cameron-freshworks: Don't fully understand explicitly setting to 0 here. My assumption is because filingId is of type Number and but you still want to represent a falsy value if there is no id present?

src/views/Correction.vue
```diff
@@ -331,20 +330,20 @@ export default {
...
 
     // this is the id of the original filing to correct
     this.correctedFilingId = +this.$route.params.correctedFilingId // number (may be NaN)
 
     // if required data isn't set, route to home
-    if (!this.entityIncNo || !this.correctedFilingId) {
+    if (!this.entityIncNo || isNaN(this.correctedFilingId)) {
```
cameron-freshworks: Was `!this.correctedFilingId)` still returning true in some instances because it was a string, albeit sometimes an empty one?

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -402,10 +404,10 @@ export default {
 
     // NB: filing id of 0 means "new"
     // otherwise it's a draft filing id
-    this.filingId = +this.$route.params.id // number
+    this.filingId = +this.$route.params.id // number (may be NaN)
```
cameron-freshworks: Just curious how this differs from the previous filing where it sets the id to 0?

src/views/Correction.vue
severinbeauvais: This one is a little different than the other filings in which Filing ID is required. In this case I give it a nice fallback value if it isn't passed in as a param (ie, optional)... and the value 0 is what other code uses to denote that this is a "new filing" (instead of resuming a draft).

One alternative would have been to force the callers to include param id=0 to denote that this is a new correction.

You can also contrast this with correctedFilingId which is required. I could have used 0 for that one too but instead I stayed consistent with the other filings that use NaN.

src/views/StandaloneDirectorsFiling.vue
severinbeauvais: 0 means new filing

> 0 means resume draft

Unspecified (ie, NaN) means you got to this page without setting the route param (eg, by returning to a bookmark) so you get rerouted to the dashboard.

src/views/Correction.vue
cameron-freshworks: Fair enough, thanks for clarifying!

src/views/StandaloneDirectorsFiling.vue
cameron-freshworks: ahhh right right, i saw that before. Right on, thanks for clarifying!

PR #39, opened 2020-03-04
misc small UI fixes

severinbeauvais:
*Issue #:* /bcgov/entity#2888

*Description of changes:*
- fixed typos (in code comments)
- updated app max-width (per Jeremy, ie, same as auth-web)
- fixed AR fee summary scrolling
- fixed Staff Payment not valid when props are set
- tweaked some text fields to prevent element from overflowing its container
- added maxlength to RSN input for better usability

### DIFF
--- a/src/components/common/StaffPayment.vue
+++ b/src/components/common/StaffPayment.vue
@@ -3,7 +3,7 @@
     <div class="payment-container">
       <label>Payment</label>
       <v-radio-group
-        class="value mt-0 pt-0"
+        class="value payments"
         :value="paymentOption"
         @change="onPaymentOptionChanged($event)"
       >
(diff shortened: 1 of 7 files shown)


src/components/common/StaffPayment.vue
```diff
@@ -106,7 +117,7 @@ export default class StaffPayment extends Vue {
   /** Emit an event indicating whether or not the form is valid. */
   @Emit('valid')
   private emitValid (): boolean {
-    return (this.formValid || this.isWaiveFees)
+    return Boolean(this.formValid || this.isWaiveFees)
```
cameron-freshworks: `!!` could also be used in place of `Boolean`, but i'm starting to prefer this approach more. Clarity > all, i think. No action required here at all, just thinking out loud :)

PR #46, opened 2020-03-22
implemented Business ID in URL route + Keycloak signin and signout

severinbeauvais:
*Issue #:* /bcgov/entity#2949

*Description of changes:*
- get Business ID from initial URL route
- set Vue Router Base to include Business ID
- refactored main.ts to fetch config before creating Vue Router
- refactored App.vue to fetch data only when authenticated
- integrated Keycloak Signin and Signout components
- changed root path (/) to be Dashboard
- updated to sbc-common-components 2.1.x
- changed route param "id" to "filingId"
- changed store and router to functions
- fixed error handling in fetchConfig
- fixed unit tests that weren't working correctly
- updated unit tests per changes
- misc cleanup


src/views/AnnualReport.vue
```diff
@@ -415,7 +417,7 @@ export default {
       return this.certifyText(FilingCodes.ANNUAL_REPORT_OT)
     },
 
-    payAPIURL () {
+    payApiUrl () {
```
ochiu: Return type string for payApiUrl, saw this updated everywhere else.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -255,7 +257,7 @@ export default {
       return (this.officeAddressFormValid && filingDataValid)
     },
 
-    payAPIURL () {
+    payApiUrl () {
```
ochiu: same as the other comment for string return type

src/views/auth/Signin.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <sbc-signin @sync-user-profile-ready="redirectOnLogin()" />
```
ochiu: When updating to 2.1.18 of the sbc common components on create business UI I had to change the listening event to keycloak-session-ready or I was experience a periodic refresh. No action required, just FYI.

src/views/AnnualReport.vue
severinbeauvais: I'll fix this in an upcoming PR.

src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: Yes.

src/views/auth/Signin.vue
severinbeauvais: Thanks!

src/App.vue
```diff
@@ -80,17 +83,44 @@ export default {
...
+    },
+
+    /** Is True if loading container should be shown, else False. */
+    showLoadingContainer (): boolean {
+      return (!this.dataLoaded && !this.dashboardUnavailableDialog && !this.accountAuthorizationDialog)
+    },
+
+    /** Is True if route is Signin, else False. */
+    isSigninRoute (): boolean {
+      return Boolean(this.$route.name === 'signin')
```
cameron-freshworks: Is this used elsewhere as well? Potential constant or if there is multiple route names, perhaps an enum?

src/App.vue
```diff
@@ -80,17 +83,44 @@ export default {
...
+    },
+
+    /** Is True if user is not authenticated, else False. */
+    isNotAuthenticated (): boolean {
+      // FUTURE: also check that token isn't expired!
+      return Boolean(!sessionStorage.getItem('KEYCLOAK_TOKEN'))
```
cameron-freshworks: Not 100% sure on the context of the method, but would it make more sense to be `isAuthenticated` and return true if they are?

src/App.vue
```diff
@@ -104,36 +134,51 @@ export default {
...
+      const now = new Date()
+      const date = this.dateToUsableString(now)
+      this.setCurrentDate(date)
+
+      // set timeout to run this again at midnight
+      const hoursToMidnight = 23 - now.getHours()
+      const minutesToMidnight = 59 - now.getMinutes()
+      const secondsToMidnight = 59 - now.getSeconds()
+      const timeout = ((((hoursToMidnight * 60) + minutesToMidnight) * 60) + secondsToMidnight) * 1000
+      this.currentDateTimerId = setTimeout(this.updateCurrentDate, timeout)
```
cameron-freshworks: Is all of this in the event a client leaves the browser open? Just curious as to the purpose of implementation!

cameron-freshworks: * Edit, i see this was all existing prior.

src/store/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  const store = new Vuex.Store({
+    actions,
+    getters,
+    mutations,
+    state
+  })
+
+  return store
+}
```
cameron-freshworks: Nice job!

src/App.vue
severinbeauvais: It's used in 3 or 4 places. I'll change it to a constant in my next PR.

src/App.vue
severinbeauvais: I originally did this to simplify where this is used (ie, check for a condition instead of !condition) but I have since changed my mind and changed this to "isAuthenticated". Great minds think alike :)

src/App.vue
severinbeauvais: Yes, just moved this up. Also added a safety check to ensure we don't get multiple timers.

PR #51, opened 2020-03-26
on startup error, try to redirect to correct environment

severinbeauvais:
*Issue #:* /bcgov/entity#3133

*Description of changes:*
- moved Business ID code after config.json code so we can redirect to the proper Businesses Registry home page if Business ID is missing
- deleted obsolete config file


src/utils/fetch-config.ts
```diff
@@ -12,34 +12,13 @@ export async function fetchConfig (): Promise<void> {
   //
   const processEnvVueAppPath: string = process.env.VUE_APP_PATH // eg, cooperatives
   const processEnvBaseUrl: string = process.env.BASE_URL // eg, /cooperatives/
-  const windowLocationPathname: string = window.location.pathname // eg, /cooperatives/CP0000841/...
-  const windowLocationOrigin: string = window.location.origin // eg, http://localhost:8080
```
severinbeauvais: Typing here is optional because the assigned value is already of type "string". Some Typescript linters will actually report a "unnecessary type annotation" error.

PR #52, opened 2020-03-30
initial Create/Draft Incorporation UI changes

severinbeauvais:
*Issue #:* /bcgov/entity#3103

*Description of changes:*
- added check for NR Number in URL
- load NR data instead of business data
- display NR data instead of business data
- added Namex mixin (temporary)
- created NR Todo Task (temporary)
- handle Name Request vs Incorporation Application
- added CREATE_URL config key
- updated Todo List functionality
- redirect to Create UI to incorporate / resume
- removed "businesses" from LEGAL_API_URL
- created EnumMixin for common helpers
- cleaned up some mixins
- hide some dashboard component when we are "NR mode" (temporary until future tickets are implemented)
- changed Account Auth dialog to Business Auth Error dialog
- created new Name Request Auth Error dialog
- created new Name Request Invalid dialog
- fixed Confirm dialog text style
- fixed existing unit tests
- created new App unit tests
- changed NamexApiMixin axios.request() to axios.post() to make it stubbable
- some cleanup


src/App.vue
```diff
@@ -324,23 +389,115 @@ export default {
...
+      if (!this.isNrValid(data)) {
+        this.nameRequestInvalidDialog = true
+        throw new Error('Invalid NR data')
+      }
+      if (!this.isNrConsumable(data)) {
+        this.nameRequestInvalidDialog = true
+        throw new Error('NR not consumable')
+      }
+
+      // FOR DEBUGGING ONLY
```
severinbeauvais: FYI, this is because the expiration data in our test NRs is several months ago :]

src/components/common/SummaryOfficeAddresses.vue
```diff
@@ -120,8 +120,8 @@ export default class SummaryOfficeAddresses extends Mixins(CommonMixin, EntityFi
   @Prop({ default: null })
   private recordsAddress: BaseAddressObjIF
 
-  // Entity Enum
-  private EntityTypes: {} = EntityTypes
+  // Enum definition for use in template.
+  private EntityTypes = EntityTypes
```
severinbeauvais: FYI, this is so we don't lose type information. (Previously type was `{}`.)

src/App.vue
```diff
@@ -324,23 +389,117 @@ export default {
...
+      data.expirationDate = new Date() // today
```
severinbeauvais: FYI, this is because the expiration data in our test NRs is several months ago :]

src/components/common/SummaryOfficeAddresses.vue
severinbeauvais: FYI, this is so we don't lose type information. (Previously type was {}.)

src/App.vue
```diff
@@ -324,23 +389,119 @@ export default {
...
+      }
+      if (!this.isNrConsumable(data)) {
+        this.nameRequestInvalidDialog = true
+        throw new Error('NR not consumable')
+      }
+
+      // FOR DEBUGGING ONLY
+      if (new Date(data.expirationDate) < new Date()) {
+        data.expirationDate = new Date() // today
+      }
```
severinbeauvais: FYI, this is because the expiration data in our test NRs is several months ago :]

(Now updated to work with unit tests/hard-coded data.)

src/mixins/namex-api-mixin.ts
```diff
@@ -0,0 +1,97 @@
...
+    const config = {
+      headers: {
+        'content-type': 'application/x-www-form-urlencoded'
+      },
+      auth: {
+        username: this.accountId,
+        password: this.accountSecret
+      }
+    }
+    return axios.post(this.tokenUrl, 'grant_type=client_credentials', config).then(res => {
```
severinbeauvais: NB: I changed this from `axios.request` to `axios.post` so I could stub it out in the App unit tests.

src/App.vue
```diff
@@ -96,9 +124,19 @@ export default {
       return sessionStorage.getItem('BUSINESS_ID')
     },
 
+    /** The NR Number string. */
+    nrNumber (): string {
+      // change _ to %20 for query purposes
+      return sessionStorage.getItem('NR_NUMBER')?.replace('_', '%20')
```
cameron-freshworks: Interesting! Doesn't it apply this automatically in the browser if entered with a space?

src/App.vue
```diff
@@ -170,43 +208,69 @@ export default {
...
-          this.storeFilings(data[3])
-          this.storeAddresses(data[4])
-          this.storeDirectors(data[5])
-          this.dataLoaded = true
-        }).catch(error => {
-          // eslint-disable-next-line no-console
-          console.error(error)
-          this.dashboardUnavailableDialog = true
-        })
+        // good so far ... fetch the business data
```
cameron-freshworks: I like the comment, like a little code-pep-talk😃

src/App.vue
cameron-freshworks: Excellent!

cameron-freshworks: Curious, do we know how long the expiration date will be in the end?

src/App.vue
```diff
@@ -324,23 +389,119 @@ export default {
...
+        } else {
+          throw new Error('Invalid tasks')
+        }
       }
     },
 
     /** Gets filings list from Legal API. */
     getFilings (): Promise<any> {
-      const url = this.businessId + '/filings'
+      const id = this.businessId || this.nrNumber
```
cameron-freshworks: Interesting change! It may be uncharted waters but what happens after a user completes their incorporation application. At what point do we stop using the NR for the identifier and start using the BusinessID?

src/components/AnnualReport/AGMDate.vue
```diff
@@ -69,7 +69,7 @@
 <script lang="ts">
 import { Component, Mixins, Vue, Prop, Watch, Emit } from 'vue-property-decorator'
 import { mapState, mapGetters } from 'vuex'
-import DateMixin from '@/mixins/date-mixin'
+import { DateMixin } from '@/mixins'
```
cameron-freshworks: nice clean up!

src/components/Dashboard/AddressListSm.vue
```diff
@@ -68,9 +68,9 @@
         </v-expansion-panel-content>
       </v-expansion-panel>
 
-      <!--Records Office-->
+      <!-- Records Office -->
       <v-expansion-panel id="records-office-panel"
-        v-if="entityFilter(EntityTypes.BCOMP)"
+        v-if="isBComp()"
```
cameron-freshworks: I like it! The entityFilter might be a better choice if we have had to differentiate betweens 10's of entity types but this is simple and explicit 👍

src/components/Dashboard/AddressListSm.vue
```diff
@@ -167,8 +167,8 @@ export default class AddressListSm extends Mixins(EntityFilterMixin, CommonMixin
   @Prop({ default: false })
   private coaPending: boolean
 
-  // EntityTypes Enum
-  readonly EntityTypes: {} = EntityTypes
+  // Enum definition for use in template.
+  readonly EntityTypes = EntityTypes
```
cameron-freshworks: This looks much better, i think we started doing this elsewhere as well, since you mentioned this wasn't required in the latest version of TS.

src/components/common/SummaryOfficeAddresses.vue
cameron-freshworks: Definitely cleaner! If only there was a way of binding the enum directly to the class without having to define a property of the same name on the class.

src/components/dialogs/NameRequestAuthErrorDialog.vue
```diff
@@ -0,0 +1,64 @@
...
+
+  // Pass click events to parent.
+  @Emit() private exit () { }
+  @Emit() private retry () { }
+}
+</script>
+
+<style lang="scss" scoped>
+// @import '@/assets/styles/theme.scss';
+</style>
```
cameron-freshworks: I'm actually glad you made these dialogs separate. I kind of prefer small and simple dialogs. Imo better for testability.

src/mixins/common-mixin.ts
```diff
@@ -1,27 +1,48 @@
...
+  /** Returns True if entity is a Cooperative. */
+  isCoop (): boolean {
+    return (this.entityType === EntityTypes.COOP)
+  }
+
+  /** Returns True if entity is a Corporation. */
+  isCorp (): boolean {
+    return (this.entityType === EntityTypes.CORP)
+  }
+
```
cameron-freshworks: Good change 👍 Large refactoring tho, your brave!

src/mixins/date-mixin.ts
```diff
@@ -101,13 +103,27 @@ export default class DateMixin extends Vue {
...
+   * @returns +1 for tomorrow
+   * @returns NaN in case of error
+   */
+  daysFromToday (date: string): number {
+    if (!date) return NaN
+    // calculate difference between start of "today" and start of "date" (in local time)
+    const todayLocalMs = new Date().setHours(0, 0, 0, 0)
+    const dateLocalMs = new Date(date).setHours(0, 0, 0, 0)
+    return Math.round((dateLocalMs - todayLocalMs) / this.MS_IN_A_DAY)
+  }
```
cameron-freshworks: What is the application of this method?

src/mixins/filing-mixin.ts
```diff
@@ -25,7 +25,7 @@ export default class FilingMixin extends Vue {
   /**
    * Flattens and sorts an array of comments.
    * @param comments the array of comments to sort and deconstruct
-   * @return the sorted and flattened array of comments
+   * @returns the sorted and flattened array of comments
```
cameron-freshworks: I've seen this! What is the difference?

src/views/Dashboard.vue
```diff
@@ -162,7 +162,12 @@ export default {
   },
 
   computed: {
-    ...mapState(['entityIncNo'])
+    ...mapState(['entityIncNo']),
+
+    /** The NR Number string. */
+    nrNumber (): string {
+      return sessionStorage.getItem('NR_NUMBER')
+    }
```
cameron-freshworks: Hmmm, there is something very similar in `App.vue` but with a str.replace. How are these two methods applied differently?

src/components/common/SummaryOfficeAddresses.vue
severinbeauvais: This is only needed for the template to see the types. The script doesn't need it. Just a Vue thing I guess.

src/App.vue
severinbeauvais: I think it's 60 days.

What needs to be clarified is how the object's 'expiration date' field relates to its 'state' field. For example, our test NRs are expired (eg, "expirationDate":"Thu, 21 Nov 2019 08:00:00 GMT") but their state is still APPROVED instead EXPIRED. Maybe the Namex API will update this when it responds with the data, or maybe there's a nightly job updating data... I will follow up with Lorna on this.

src/App.vue
severinbeauvais: The Auth API, for one, doesn't respond with the requested data if I enter "NR xxxxxxx" instead of "NR%20xxxxxxx".

In the browser URL, the space is automatically converted to %20. (And also when  you cut and paste to/from the browser URL.)

src/App.vue
severinbeauvais: Another gem for you:
`// just let signin page do its thing`

src/App.vue
severinbeauvais: Um... (thinking)

We will do whatever Scott designs UX-wise, and this also depends on what BE processing needs to happen for a draft incorporation to become a business, but it would be nice if we could redirect the user from `...cooperatives/NRxxxxxxx` to `...cooperatives/CPxxxxxxx`.

src/mixins/common-mixin.ts
severinbeauvais: What can I say? I like clean code! Makes it easier to review, too.

src/mixins/date-mixin.ts
severinbeauvais:
```
    expiresText (task): string {
      const expireDays = this.daysFromToday(task.nameRequest?.expirationDate)
      if (isNaN(expireDays) || expireDays < 0) {
        return 'Expired'
      } else if (expireDays < 1) {
        return 'Expires today'
      } else if (expireDays < 2) {
        return 'Expires tomorrow'
      } else {
        return `Expires in ${expireDays} days`
      }
    },
```

src/mixins/filing-mixin.ts
severinbeauvais: Both are allowed, but when I hover over the function I prefer to see "returns XXX" instead of "return XXX" :)

src/views/Dashboard.vue
severinbeauvais: In the dashboard I don't care what format the string is, just whether it truthy or falsy.
In the other modules I may actually use the string, either for display (with a space) or for querying (with a %20).
So I kept it simple here.

src/mixins/date-mixin.ts
severinbeauvais: Basically return 0 if date is today, -X if date is X days in the past, or +X if date is X days in the future.

PR #55, opened 2020-03-31
added token refresh service

severinbeauvais:
*Issue #:* /bcgov/entity#3163

*Description of changes:*
- added TokenService to handle refreshes
- initialized after signin success (ie, once tokens are populated)

### DIFF
--- a/src/views/auth/Signin.vue
+++ b/src/views/auth/Signin.vue
@@ -1,10 +1,11 @@
 <template>
-  <sbc-signin @keycloak-session-ready="navigateToDashboard()" />
+  <sbc-signin @keycloak-session-ready="onReady()" />
 </template>
 
 <script lang="ts">
 // Libraries
 import { Component, Vue } from 'vue-property-decorator'
+import TokenService from 'sbc-common-components/src/services/token.services'
 
 // Components
 import SbcSignin from 'sbc-common-components/src/components/SbcSignin.vue'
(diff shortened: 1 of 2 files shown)


src/views/auth/Signin.vue
```diff
@@ -18,8 +19,16 @@ import { DASHBOARD } from '@/constants'
...
-  /** Called when signin has succeeded. */
-  private navigateToDashboard () {
+  private tokenService = new TokenService()
+
+  /** Called when Keycloak session is ready. */
+  private async onReady () {
+    // start token service to refresh KC token periodically
+    console.info('Starting token refresh service...') // eslint-disable-line no-console
+    await this.tokenService.init()
+    this.tokenService.scheduleRefreshTimer()
```
ochiu: Does this cover the case where someone is already authenticated and doesn't get forwarded to the sign in component? Do we need to handle that separately so the token service code handles that?

severinbeauvais: I just tested and no, it doesn't! I need to rework this.

ochiu: haha great! thanks for checking.

severinbeauvais: A related question would be, if there are multiple tabs each with the token refresh service, could the token be refreshed by each tab?

@sutherlanda

severinbeauvais: > Does this cover the case where someone is already authenticated and doesn't get forwarded to the sign in component? Do we need to handle that separately so the token service code handles that?

The full reply is, the app always redirects to signin (since there are no KC keys in session storage). The difference is that the signin functionality won't prompt the user to login if the KC cookie already exists. So our Signin page is always in play upon return from signin.

sutherlanda: I believe it would start an instance of the TokenService for each each tab, although I'm not sure if this is something we've tested outright.

PR #56, opened 2020-04-02
updated NR base functionality

severinbeauvais:
*Issue #:* /bcgov/entity#3293

*Description of changes:*
- updated NR query to use new legal-api endpoint
- set debugging expiration date to tomorrow
- added handling in case of tasks/filings 404
- use only approved NR name
- replaced Namex Mixin with Name Request mixin
- don't display loading message when signing out
- fixed NR display by simplifying its usage (now always stored with a space)
- minor refactoring and cleanup
- updated unit tests
- updated app version to 2.9.0


src/mixins/namex-request-mixin.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    // If the NR's root state is not APPROVED, it is not consumable.
+    if (nr.state !== NameRequestStates.APPROVED) {
+      return NameRequestStates.NOT_APPROVED
+    }
+
+    // NB: from here down, the NR is APPROVED
+
+    // If the NR has already been consumed, it is not consumable.
+    if (nr.names.some(name => name.consumptionDate)) {
```
ochiu, line 45: Nice. I like this update, I forget about some().

src/App.vue
```diff
@@ -317,7 +323,7 @@ export default {
 
     storeAuthorizations (response: any): void {
       // NB: roles array may contain 'view', 'edit' or nothing
-      const authRoles = response && response.data && response.data.roles
+      const authRoles = response?.data?.roles
```
cameron-freshworks, line 326: nice clean up. I read should read the docs a little further on optional chaining and how it works. I sure do like doing away with the short circuiting!

severinbeauvais: It stops evaluating when the token is null. E.g.,
null?.data.roles is null
response.null?.roles is null
response.data.null is null

severinbeauvais:
> Looks good! Feel free to merge after your updates or request another review if you think it's needed👌

Yes, please re-review (or at least just look at the 2 new commits), as I think you should know about the things I changed.

Second reviewer can now merge.

 Thanks!

ochiu:
> Yes, please re-review (or at least just look at the 2 new commits), as I think you should know about the things I changed.
> 
> Second reviewer can now merge.
> 
> Thanks!

Looks good!

PR #57, opened 2020-04-02
Director Name Change Reset

cameron-freshworks:
*Issue #:* /bcgov/entity#1613

*Description of changes:*
* Implemented director name change reset btn.
* Fixed the cancel btn functionality.
* Updated unit testing.

### DIFF
--- a/src/components/common/Directors.vue
+++ b/src/components/common/Directors.vue
@@ -456,6 +456,14 @@
                     >
                       <span>Remove</span>
                     </v-btn>
+                    <v-btn color="error"
+                     class="reset-btn"
+                     v-show="!isNew(director) && editFormShowHide.showName"
+                     @click="restoreDirName(director.id, true)"
+                     :disabled="!isNameChanged(director)"
+                    >
+                      <span>Reset</span>
+                    </v-btn>
                     <v-btn color="primary"
                       class="form-primary-btn done-edit-btn"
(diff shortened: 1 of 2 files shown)


src/components/common/Directors.vue
```diff
@@ -456,6 +456,14 @@
                     >
                       <span>Remove</span>
                     </v-btn>
+                    <v-btn color="error"
+                     class="reset-edit-btn"
+                     v-show="!isNew(director) && editFormShowHide.showName"
```
severinbeauvais, line 461: I think we want to show the button for a new director but leave it disabled. This is to show users where the button lives if they edit a director.

cameron-freshworks: Interesting... that seems weird to me because the UI is different for editing a new director and a director name change. The new director has a 'Remove' btn, i don't understand showing a btn there that they would never use.

cameron-freshworks:
> Just a functionality question. The rest looks great. Love the tests!

Awesome thanks sir, perhaps we should discuss the design a little more :)

src/components/common/Directors.vue
cameron-freshworks: Lets talk about this offline, i think it might be a design question!

severinbeauvais: Maybe you're right. But this is a Scott question.

PR #63, opened 2020-04-21
3107, 3108

ochiu:
*Issue #:* /bcgov/entity#3107, /bcgov/entity#3108

*Description of changes:*
- Updated dashboard to show address and directors list for NRs
- Update dashboard to disable change address / change directors buttons for NRs without a completed filing
- Update Address list component to show 'Complete your filing to display'
- Update Directors list component to show 'Complete your filing to display'
- Updated Dashboard unit tests
- Updated Address list unit tests
- Updated Director list unit tests

### DIFF
--- a/src/components/Dashboard/AddressListSm.vue
+++ b/src/components/Dashboard/AddressListSm.vue
@@ -9,8 +9,40 @@
         <v-expansion-panel-header id="registered-office-panel-toggle">
           <div class="list-item__title">Registered Office</div>
         </v-expansion-panel-header>
+        <!-- A completed filing is required; Don't show address information -->
+        <v-expansion-panel-content class="panel-wrapper pt-0 pb-0" v-if="completedFilingRequired">
+          <v-list class="pt-0 pb-0">
+            <v-list-item class="delivery-address-list-item">
+              <v-list-item-icon class="address-icon mr-0">
+                <v-icon color="primary">mdi-truck</v-icon>
+              </v-list-item-icon>
+              <v-list-item-content>
+                <v-list-item-title class="mb-2 address-title">Delivery Address</v-list-item-title>
+                <v-list-item-subtitle>
+                  <div>
+                    <span class="complete-filing">Complete your filing to display</span>
+                  </div>
+                </v-list-item-subtitle>
+              </v-list-item-content>
+            </v-list-item>
 
-        <v-expansion-panel-content class="panel-wrapper pt-0 pb-0">
+            <v-list-item class="mailing-address-list-item pt-10">
+              <v-list-item-icon class="address-icon mr-0">
+                <v-icon color="primary">mdi-email-outline</v-icon>
+              </v-list-item-icon>
+              <v-list-item-content>
+                <v-list-item-title class="mb-2 address-title">Mailing Address</v-list-item-title>
+                <v-list-item-subtitle>
+                  <div>
+                    <span class="complete-filing">Complete your filing to display</span>
+                  </div>
+                </v-list-item-subtitle>
+              </v-list-item-content>
+            </v-list-item>
+          </v-list>
+        </v-expansion-panel-content>
+        <!-- A completed filing is NOT required - Show address information -->
+        <v-expansion-panel-content class="panel-wrapper pt-0 pb-0" v-else>
           <v-list class="pt-0 pb-0" v-if="registeredAddress">
             <v-list-item class="delivery-address-list-item"
               v-if="registeredAddress.deliveryAddress"
@@ -77,8 +109,40 @@
         <v-expansion-panel-header id="records-office-panel-toggle">
           <div class="list-item__title">Records Office</div>
         </v-expansion-panel-header>
+        <!-- A completed filing is required; Don't show address information -->
+        <v-expansion-panel-content class="panel-wrapper" v-if="completedFilingRequired">
+          <v-list class="pt-0 pb-0">
+            <v-list-item class="delivery-address-list-item">
+              <v-list-item-icon class="address-icon mr-0">
+                <v-icon color="primary">mdi-truck</v-icon>
+              </v-list-item-icon>
+              <v-list-item-content>
+                <v-list-item-title class="mb-2 address-title">Delivery Address</v-list-item-title>
+                <v-list-item-subtitle>
+                  <div>
+                    <span class="complete-filing">Complete your filing to display</span>
+                  </div>
+                </v-list-item-subtitle>
+              </v-list-item-content>
+            </v-list-item>
 
-        <v-expansion-panel-content class="panel-wrapper">
+            <v-list-item class="mailing-address-list-item pt-10">
+              <v-list-item-icon class="address-icon mr-0">
+                <v-icon color="primary">mdi-email-outline</v-icon>
+              </v-list-item-icon>
+              <v-list-item-content>
+                <v-list-item-title class="mb-2 address-title">Mailing Address</v-list-item-title>
+                <v-list-item-subtitle>
+                  <div>
+                    <span class="complete-filing">Complete your filing to display</span>
+                  </div>
+                </v-list-item-subtitle>
+              </v-list-item-content>
+            </v-list-item>
+          </v-list>
+        </v-expansion-panel-content>
+        <!-- A completed filing is NOT required - Show address information -->
+        <v-expansion-panel-content class="panel-wrapper" v-else>
           <v-list class="pt-0 pb-0" v-if="recordsAddress">
             <v-list-item class="delivery-address-list-item"
               v-if="recordsAddress.deliveryAddress"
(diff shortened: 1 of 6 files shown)


src/components/Dashboard/AddressListSm.vue
```diff
@@ -9,8 +9,40 @@
...
+            <v-list-item class="delivery-address-list-item">
+              <v-list-item-icon class="address-icon mr-0">
+                <v-icon color="primary">mdi-truck</v-icon>
+              </v-list-item-icon>
+              <v-list-item-content>
+                <v-list-item-title class="mb-2 address-title">Delivery Address</v-list-item-title>
+                <v-list-item-subtitle>
+                  <div>
+                    <span class="complete-filing">Complete your filing to display</span>
+                  </div>
```
cameron-freshworks, line 24: Just curious, was wrapping the spans in divs for spacing?

src/views/Dashboard.vue
```diff
@@ -167,6 +171,20 @@ export default {
...
     nrNumber (): string {
       return sessionStorage.getItem('NR_NUMBER')
+    },
+
+    /** Checks if changes need to be disabled */
+    disableChanges () : boolean {
+      if (this.nrNumber) {
+        return !this.hasCompletedIncorporationFiling
+      } else {
+        return this.hasBlockerFiling
```
cameron-freshworks, line 181: This could be compacted it a bit, but i'm just nitpicking :) 

`return this.nrNumber ? !this.hasCompletedIncorporationFiling : this.hasBlockerFiling`

src/components/Dashboard/AddressListSm.vue
ochiu: Just a habit of mine to more easily style a group of elements within the div as well as the individual children. In this case it isn't used, but I just structured it this way.

Cleaner to remove?

src/views/Dashboard.vue
ochiu: I like it more compact for this :) good call. 

Updated.

src/components/Dashboard/AddressListSm.vue
cameron-freshworks: I'm totally fine with it, was just curious!

severinbeauvais: My 2 cents...

If text won't fit on 1 line between a start and end tag then I'll wrap it in a span, so that, during testing, the exact text can be selected... otherwise you sometimes get extra whitespace in the result.

src/components/Dashboard/AddressListSm.vue
```diff
@@ -9,8 +9,40 @@
         <v-expansion-panel-header id="registered-office-panel-toggle">
           <div class="list-item__title">Registered Office</div>
         </v-expansion-panel-header>
+        <!-- A completed filing is required; Don't show address information -->
```
severinbeauvais, line 12: It's too bad about all this duplicated markup. Was it too messy to have "v-if"s at the "v-list-item-subtitle" element?

src/components/Dashboard/DirectorListSm.vue
```diff
@@ -85,6 +92,14 @@ export default class DirectorListSm extends Mixins(CountriesProvincesMixin, Comm
 // Variables
 $avatar-width: 2.75rem;
 
+// Complete filing required styling
+.complete-filing {
+  padding: 2rem;
```
severinbeauvais, line 97: You don't need padding here and in the div's class, do you?

src/views/Dashboard.vue
```diff
@@ -293,6 +327,9 @@ export default {
       if (this.isBComp()) {
         this.checkPendingFilings(this.historyFilings)
       }
+      // check for completed filing on an incoporation application
+      this.checkCompletedIncorporationFilings(this.historyFilings)
```
severinbeauvais, line 331: I suggest adding `if (this.nrNumber)` to this which will do less work (if not NR) and remove the need for the NR check on line 183 (and probably remove the need for the computed property "completedFilingRequired").

src/components/Dashboard/AddressListSm.vue
ochiu: I tried that at the beginning, but some of the elements further up had v-ifs as well as specific styling based on the coaPending. I did my best to preserve the existing behavior so I duplicated some of the mark up and removed some of the logic.

tests/unit/AddressListSm.spec.ts
```diff
@@ -259,4 +259,64 @@ describe('AddressListSm', () => {
...
+    expect(vm.$el.querySelector('#registered-office-panel .delivery-address-list-item .complete-filing')
+      .textContent).toBe(expectedMessage)
+
+    expect(vm.$el.querySelector('#registered-office-panel .mailing-address-list-item .complete-filing')
+      .textContent).toBe(expectedMessage)
+
+    expect(vm.$el.querySelector('#records-office-panel .delivery-address-list-item .complete-filing')
+      .textContent).toBe(expectedMessage)
+
+    expect(vm.$el.querySelector('#records-office-panel .mailing-address-list-item .complete-filing')
```
severinbeauvais, line 289: Good tests for BCOMP 👍 

If you're in this file again, I think we should move as much as possible to `wrapper.find()` to select elements instead of using `vm.$el.querySelector()`. It's cleaner and has richer methods :)

tests/unit/Dashboard.spec.ts
```diff
@@ -263,4 +263,46 @@ describe('Dashboard - Click Tests', () => {
...
+        store,
+        vuetify,
+        computed: {
+          disableChanges () {
+            return true
+          }
+        }
+      })
+
+    Vue.nextTick(async () => {
```
severinbeauvais, line 279: Please use `await nextTick()` instead. I guess you copied existing code but the problem with it is that the body of the nextTick executes in a different scope and you get no detail on any errors (you only get "test didn't complete within timeout").

src/components/Dashboard/DirectorListSm.vue
ochiu: Nope! Actually re-tested this and found my styling was incorrect, so fixed that.
I kept the vertical padding to the div as it wasn't quite working on the span.

src/views/Dashboard.vue
```diff
@@ -267,6 +281,26 @@ export default {
...
+
+      const completedFiling = filings.find(filing => {
+        return filing.name === FilingNames.INCORPORATION_APPLICATION && filing.status === FilingStatus.COMPLETED
+      })
+
+      if (completedFiling) {
+        this.hasCompletedIncorporationFiling = true
+      } else {
+        this.hasCompletedIncorporationFiling = false
+      }
```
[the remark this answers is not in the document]
ochiu: Done! Thanks!

tests/unit/Dashboard.spec.ts
ochiu: Updated my tests and the other tests in this file.

src/views/Dashboard.vue
ochiu: added the NR check, but after taking a look. I feel like I can do a bit more clean up based on your feed back about not needing the computed property.

Will commit another update in a bit.

ochiu: Updated.
- removed computed completedFilingRequired
- renamed variable hasCompletedIncorporationFIling to completedFilingRequired as we are more interested in the fact that is is required for the list components

tests/unit/AddressListSm.spec.ts
severinbeauvais: Fixed, thanks!

src/components/Dashboard/DirectorListSm.vue
```diff
@@ -85,6 +92,14 @@ export default class DirectorListSm extends Mixins(CountriesProvincesMixin, Comm
 // Variables
 $avatar-width: 2.75rem;
 
+// Complete filing required styling
+.complete-filing {
+  padding: 0rem 2rem;
```
severinbeauvais, line 97: I don't get it. Why do you need padding for both the div ("class="pt-4 pb-4") and the span in it (this line)?

src/views/Dashboard.vue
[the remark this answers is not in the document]
severinbeauvais: Cleaner and getting rid of another property makes it easier to understand :)

tests/unit/Dashboard.spec.ts
severinbeauvais: Way nicer (and we'll get proper errors if anything fails). 👍

src/components/Dashboard/DirectorListSm.vue
severinbeauvais: I'm not going to worry about this one.

ochiu: Default span behavior doesn't let me influence the vertical spacing.

Updated display attribute on `.complete-filing` so I can keep the padding to the single style.

PR #64, opened 2020-04-23
Todo List for draft incorp app

severinbeauvais:
*Issue #:* /bcgov/entity#3101

*Description of changes:*
- changed pending status icon to alert
- applied colours to pending, error and paid status icons
- added missing payment token code to incorp apps and corrections
- added incorp app unit test
- updated sbc-common-components to 2.1.27 for AUTH_URL fix
- updated app version to 2.10.0
- added unit tests for NR and Incorp App actions

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -68,7 +68,7 @@
               </div>
 
               <div class="list-item__subtitle">
-                <div v-if="(isCoop() || isCorp()) && task.subtitle" class="todo-status">
+                <div v-if="task.subtitle" class="todo-status">
                   <span>{{task.subtitle}}</span>
                 </div>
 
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -442,6 +444,7 @@ export default {
           case FilingTypes.ANNUAL_REPORT: {
             const ARFilingYear = todo.header.ARFilingYear
             this.taskItems.push({
+              id: -1, // not falsy
```
severinbeauvais, line 447: This is needed so that the conditional on line 146 isn't "true" during unit tests. (Because "inProcessFiling" prop defaults to null and "id" was previously undefined. Those should not be strictly equal but when Jest is running, apparently they are.)

src/components/Dashboard/TodoList.vue
```diff
@@ -601,6 +605,7 @@ export default {
           status: filing.header.status,
           enabled: Boolean(task.enabled),
           order: task.order,
+          paymentToken: filing.header.paymentToken || null,
```
severinbeauvais, line 608: This won't do anything today, but in the future correction payments will need to be resumed/retried.

src/components/Dashboard/TodoList.vue
```diff
@@ -619,7 +624,8 @@ export default {
           draftTitle: 'Incorporation Application',
           status: filing.header.status,
           enabled: Boolean(task.enabled),
-          order: task.order
+          order: task.order,
+          paymentToken: filing.header.paymentToken || null
```
severinbeauvais, line 628: This was a bug for resume/retry payment. The pay url was incorrect because it needs the payment token id.

tests/unit/TodoList.spec.ts
```diff
@@ -1606,6 +1606,60 @@ describe('TodoList - Click Tests - BCOMPs', () => {
     })
   })
 
+  it('routes to Annual Report page when \'Resume\' is clicked', done => {
```
severinbeauvais, line 1609: Added this missing test for BCOMP.

tests/unit/TodoList.spec.ts
```diff
@@ -1709,6 +1763,103 @@ describe('TodoList - Click Tests - BCOMPs', () => {
   })
 })
 
+describe('TodoList - Click Tests - NRs and Incorp Apps', () => {
+  const { assign } = window.location
+
+  beforeAll(() => {
+    // init store
+    sessionStorage.setItem('CREATE_URL', `${process.env.VUE_APP_PATH}/create/`)
```
severinbeauvais, line 1771: Not quite the correct URL but doesn't matter for the unit tests.
This will probably be more correct in the future when we clean up our URLs :)

severinbeauvais:
test comment for email test (reviewers please ignore this)

PR #65, opened 2020-04-23
3107,3108

ochiu:
*Issue #:* /bcgov/entity#3107, /bcgov/entity#3108

*Description of changes:*
- bug fix on completed filings fix

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -277,23 +277,6 @@ export default {
     },
 
     /**
-     * Searches the filings history for a completed incorporation application
-     * Used to disable the ability to change directors and addresses for NRs
-     *
-     * @param filings The array of filings in history
-     */
-    checkCompletedIncorporationFilings (filings: Array<any>) {
-      if (!filings || !filings.length) {
-        // No filings, set completed filing required to true
-        this.completedFilingRequired = true
-      } else {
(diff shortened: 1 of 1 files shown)


ochiu:
Discussed with @severinbeauvais to simplify this check as once a filing is complete the real flow will transition to a non NR identifier.

src/views/Dashboard.vue
```diff
@@ -316,9 +302,10 @@ export default {
       if (this.isBComp()) {
         this.checkPendingFilings(this.historyFilings)
       }
-      // check for completed filing on an incoporation application
+      // Set completed filing required to show address since this is an NR
+      // once transitioned to the business id, addresses/directors will be available
       if (this.nrNumber) {
-        this.checkCompletedIncorporationFilings(this.historyFilings)
+        this.completedFilingRequired = true
```
severinbeauvais, line 308: Will this get set if this.historyFilings doesn't change (ie, there aren't any filings)?

ochiu: It is being triggered on my local build, so it looks like it does.

ochiu: Perhaps I should move it to mounted() ?

severinbeauvais: Sure, if you want to keep using that property. Or use the nrNumber check directly where needed?

Either is OK with me.

PR #67, opened 2020-04-27
Fix for alignment issue in dev

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3106

*Description of changes:*
The layout in dev seems to be a bit different from what I see in local for many components. For todo list NR in particular, the icon position looks a bit off. After trying the styles in dev, the one in this PR looks ok.

### DIFF
--- a/src/components/common/NameRequestInfo.vue
+++ b/src/components/common/NameRequestInfo.vue
@@ -156,9 +156,11 @@ export default class NameRequestInfo extends Mixins(CommonMixin, DateMixin, Enum
 
   #name-request-summary {
     font-size: 0.9rem;
+    margin-left: 0.875rem
   }
 
   .nr-status-icon {
-    margin-left:-2rem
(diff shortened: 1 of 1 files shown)


src/components/common/NameRequestInfo.vue
```diff
@@ -156,9 +156,11 @@ export default class NameRequestInfo extends Mixins(CommonMixin, DateMixin, Enum
 
   #name-request-summary {
     font-size: 0.9rem;
+    margin-left: 0.875rem
   }
 
   .nr-status-icon {
-    margin-left:-2rem
+    margin-left:-2rem;
+    margin-right:0.2 rem
```
severinbeauvais, line 164: In the future, keep an eye on spacing and punctuation.
- normally CSS properties should end with a semi-colon
- normally there should be a space between colon and value
- normally there is no space between value and "rem"... does it actually work here?

PR #68, opened 2020-04-28
updated Todo List buttons

severinbeauvais:
*Issue #:* /bcgov/entity#3101

*Description of changes:*
- added View/Hide Details to some buttons (per discussion with Scott)
- consolidated button styles

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -48,7 +48,7 @@
             <div class="todo-label">
               <h3 class="list-item__title">{{task.title}}
                 <div v-if="isTypeCorrection(task) && isStatusDraft(task)">
-                  <v-btn small icon color="red" class="info-btn">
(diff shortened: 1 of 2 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -139,8 +154,15 @@
                   </div>
                   <div class="payment-status" v-else>
                     <span>PAID</span>
-                    <v-btn small icon color="black" class="info-btn">
-                      <v-icon color="red darken-4">mdi-information-outline</v-icon>
+                    <v-btn
+                      class="expand-btn"
+                      outlined
+                      color="red darken-4"
+                      :ripple=false
```
cameron-freshworks, line 161: Is ripple a little graphic when clicked?

severinbeauvais: It's like of like a wave of colour starting from the middle and going to the edges. It's passe, frankly :)  And I'm just following the previous design.

PR #70, opened 2020-04-30
2596 - launchdarkly clean up and basic logic

ochiu:
*Issue #:* /bcgov/entity#2596

*Description of changes:*
- removed old library ld-vue
- updated to use launchdarkly-js-client-sdk (same as create ui)
- added FeatureFlags class
- added basic logic to block the app from loading up if it is an NR

### DIFF
--- a/public/config/configuration.json
+++ b/public/config/configuration.json
@@ -7,5 +7,5 @@
   "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "KEYCLOAK_CONFIG_PATH": "/local-keycloak-config-url/keycloak.json",
   "ADDRESS_COMPLETE_KEY": "RH99-RN99-GB72-FW86",
-  "LD_CLIENT_ID": "5dc1c5fd1d213d087f9e1024"
+  "LD_CLIENT_ID": "5e5d5af4574f7e09ab53d03b"
 }
--- a/src/utils/fetch-config.ts
+++ b/src/utils/fetch-config.ts
@@ -1,7 +1,6 @@
 /* eslint-disable no-console */
 
 import axios from '@/axios-auth'
-import { initLDClient, featureFlags } from '@/common/FeatureFlags'
 
 /**
  * Fetches config from environment and API.
@@ -69,9 +68,6 @@ export async function fetchConfig (): Promise<void> {
   window['ldClientId'] = ldClientId
   console.info('Set Launch Darkly Client ID.')
 
-  // initialize Launch Darkly
-  await initLDClient()
-
   // get Business ID / NR Number and validate that it looks OK
   // it should be first token after Base URL in Pathname
   // also change %20 to space for display and routing purposes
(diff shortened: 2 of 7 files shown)


public/config/configuration.json
```diff
@@ -7,5 +7,5 @@
   "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "KEYCLOAK_CONFIG_PATH": "/local-keycloak-config-url/keycloak.json",
   "ADDRESS_COMPLETE_KEY": "RH99-RN99-GB72-FW86",
-  "LD_CLIENT_ID": "5dc1c5fd1d213d087f9e1024"
+  "LD_CLIENT_ID": "5e5d5af4574f7e09ab53d03b"
```
ochiu, line 10: Discussed with @lekshmimallika-aot and we think we probably should use a single LD_CLIENT_ID so we can disable the features in one location. Thoughts?

src/utils/fetch-config.ts
```diff
@@ -68,6 +69,9 @@ export async function fetchConfig (): Promise<void> {
   window['ldClientId'] = ldClientId
   console.info('Set Launch Darkly Client ID.')
 
+  // initialize Launch Darkly
+  await initLDClient()
+
```
severinbeauvais, line 74: I'd prefer to see this init visible in `main.ts` (same as in Create UI).

src/views/Dashboard.vue
```diff
@@ -293,7 +290,8 @@ export default {
 
   mounted (): void {
     // Launch Darkly flag usage example:
-    // console.log('coopsVersion =', this.flags.coopsVersion)
+    // First import common class FeatureFlags
+    // console.log('bcrs-create-ui-enabled =', featureFlags.getFlag('bcrs-create-ui-enabled'))
   },
```
severinbeauvais, line 295: Looks like all of `mounted()` can be deleted :)

public/config/configuration.json
severinbeauvais: Sure, but what about the different environments?

And if you change it here, do you also have to change the config map keys for Dev and/or Test and/or Prod?

src/utils/fetch-config.ts
ochiu: The validation piece is right after this line of code and we need the ldClientId to be set for the `initLDClient()`

Should I move the feature flag check elsewhere?

public/config/configuration.json
ochiu: I am not sure if we have different client ids per environment, @lekshmimallika-aot ?

But yes, if we change it here we need to change it for dev, test, and prod so it shares the same set of properties

src/utils/fetch-config.ts
severinbeauvais: No, I had in mind the FF check here, but I forgot that you need to load the config before init (catch-22).

What do you think about moving this FF check to main.ts and wrapping it around the app start (same as Create UI)? You could then log a message if the flag is false (and/or throw an error that the existing catch logic would handle).

src/views/Dashboard.vue
ochiu: Done.

src/utils/fetch-config.ts
ochiu: Sounds good, will do that

public/config/configuration.json
lekshmimallika-aot: Different environments will have different client ID. That will have to be updated in the config map of the respective environments. Here we can keep this one

PR #75, opened 2020-05-07
initial commit

peter-freshworks:
*Issue #:* /bcgov/entity#2764

*Description of changes:*

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -403,7 +403,7 @@ import { DetailsList, NameRequestInfo } from '@/components/common'
 import { AddCommentDialog, ConfirmDialog, DeleteErrorDialog, CancelPaymentErrorDialog } from '@/components/dialogs'
 
 // Mixins
-import { CommonMixin, DateMixin, EnumMixin, FilingMixin } from '@/mixins'
+import { BcolMixin, CommonMixin, DateMixin, EnumMixin, FilingMixin } from '@/mixins'
 
 // Enums and Constants
 import { EntityTypes, FilingStatus, FilingTypes } from '@/enums'
@@ -421,7 +421,7 @@ export default {
     NameRequestInfo
   },
 
-  mixins: [CommonMixin, DateMixin, EnumMixin, FilingMixin, Vue2Filters.mixin],
+  mixins: [BcolMixin, CommonMixin, DateMixin, EnumMixin, FilingMixin, Vue2Filters.mixin],
 
   data () {
     return {
@@ -436,6 +436,7 @@ export default {
       confirmEnabled: false,
       currentFilingId: null,
       panel: null as number, // currently expanded panel
+      payAPiUrl: sessionStorage.getItem('PAY_API_URL'),
 
       // enums
       EntityTypes,
@@ -599,7 +600,7 @@ export default {
         const bcolErr = filing.header.paymentErrorType || null
         let bcolObj = null
         if (bcolErr) {
-          bcolObj = await this.getBcolMessage(bcolErr)
+          bcolObj = await this.getErrorObj(this.payApiUrl, bcolErr)
         }
         if (date) {
           const ARFilingYear = +date.substring(0, 4)
@@ -632,7 +633,7 @@ export default {
         const bcolErr = filing.header.paymentErrorType || null
         let bcolObj = null
         if (bcolErr) {
-          bcolObj = await this.getBcolMessage(bcolErr)
+          bcolObj = await this.getErrorObj(this.payAPiUrl, bcolErr)
         }
         this.taskItems.push({
           type: FilingTypes.CHANGE_OF_DIRECTORS,
@@ -658,7 +659,7 @@ export default {
         const bcolErr = filing.header.paymentErrorType || null
         let bcolObj = null
         if (bcolErr) {
-          bcolObj = await this.getBcolMessage(bcolErr)
+          bcolObj = await this.getErrorObj(this.payApiUrl, bcolErr)
         }
         this.taskItems.push({
           type: FilingTypes.CHANGE_OF_ADDRESS,
@@ -684,7 +685,7 @@ export default {
         const bcolErr = filing.header.paymentErrorType || null
         let bcolObj = null
         if (bcolErr) {
-          bcolObj = await this.getBcolMessage(bcolErr)
+          bcolObj = await this.getErrorObj(this.payApiUrl, bcolErr)
         }
         this.taskItems.push({
           type: FilingTypes.CORRECTION,
@@ -713,7 +714,7 @@ export default {
         const bcolErr = filing.header.paymentErrorType || null
         let bcolObj = null
         if (bcolErr) {
-          bcolObj = await this.getBcolMessage(bcolErr)
+          bcolObj = await this.getErrorObj(this.payApiUrl, bcolErr)
         }
         this.taskItems.push({
           type: FilingTypes.INCORPORATION_APPLICATION,
@@ -936,16 +937,6 @@ export default {
           this.cancelPaymentErrorDialog = true
         }
       })
-    },
-
-    async getBcolMessage (errCode: string): Promise<any> {
-      const fetchUrl = sessionStorage.getItem('PAY_API_URL') + 'codes/errors/' + errCode
-      const errObj = await axios.get(fetchUrl).catch(() => { return null })
-      if (errObj && errObj.data) {
-        return errObj.data
-      } else {
-        return null
-      }
     },
 
     showCommentDialog (filingId: number): void {
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -570,7 +570,7 @@ export default {
             await this.loadChangeOfAddress(task)
             break
           case FilingTypes.CORRECTION:
-            await this.loadCorrection(task)
+            this.loadCorrection(task)
             break
           case FilingTypes.INCORPORATION_APPLICATION:
             await this.loadIncorporationApplication(task)
@@ -670,7 +670,7 @@ export default {
       }
     },
 
-    async loadCorrection (task) {
+    loadCorrection (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.correction) {
         this.taskItems.push({
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -1132,7 +1132,7 @@ export default {
   border-top: solid #a94442 3px;
 }
 
-::v-deep .bcol-todo-list-detail {
+.bcol-todo-list-detail {
   background-color: #f1f1f1 !important;
 }
 </style>
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -456,10 +456,6 @@ export default {
       return sessionStorage.getItem('BUSINESS_ID')
     },
 
-    payAPiUrl (): string {
-      return sessionStorage.getItem('PAY_API_URL')
-    },
-
     /** The NR Number string. */
     nrNumber (): string {
       return sessionStorage.getItem('NR_NUMBER')
@@ -677,11 +673,6 @@ export default {
     async loadCorrection (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.correction) {
-        const bcolErr = filing.header.paymentErrorType || null
-        let bcolObj = null
-        if (bcolErr) {
-          bcolObj = await this.getErrorObj(bcolErr)
-        }
         this.taskItems.push({
           type: FilingTypes.CORRECTION,
           id: filing.header.filingId,
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -597,7 +597,10 @@ export default {
           ? date = filing.annualReport.annualReportDate
           : date = filing.annualReport.nextARDate
         const bcolErr = filing.header.paymentErrorType || null
-        const bcolObj = await this.getBcolMessage(bcolErr)
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getBcolMessage(bcolErr)
+        }
         if (date) {
           const ARFilingYear = +date.substring(0, 4)
           this.taskItems.push({
@@ -627,7 +630,10 @@ export default {
       const filing = task.task.filing
       if (filing && filing.header && filing.changeOfDirectors) {
         const bcolErr = filing.header.paymentErrorType || null
-        const bcolObj = await this.getBcolMessage(bcolErr)
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getBcolMessage(bcolErr)
+        }
         this.taskItems.push({
           type: FilingTypes.CHANGE_OF_DIRECTORS,
           id: filing.header.filingId,
@@ -650,7 +656,10 @@ export default {
       const filing = task.task.filing
       if (filing && filing.header && filing.changeOfAddress) {
         const bcolErr = filing.header.paymentErrorType || null
-        const bcolObj = await this.getBcolMessage(bcolErr)
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getBcolMessage(bcolErr)
+        }
         this.taskItems.push({
           type: FilingTypes.CHANGE_OF_ADDRESS,
           id: filing.header.filingId,
@@ -673,7 +682,10 @@ export default {
       const filing = task.task.filing
       if (filing && filing.header && filing.correction) {
         const bcolErr = filing.header.paymentErrorType || null
-        const bcolObj = await this.getBcolMessage(bcolErr)
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getBcolMessage(bcolErr)
+        }
         this.taskItems.push({
           type: FilingTypes.CORRECTION,
           id: filing.header.filingId,
@@ -699,7 +711,10 @@ export default {
       const filing = task.task.filing
       if (filing && filing.header && filing.incorporationApplication) {
         const bcolErr = filing.header.paymentErrorType || null
-        const bcolObj = await this.getBcolMessage(bcolErr)
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getBcolMessage(bcolErr)
+        }
         this.taskItems.push({
           type: FilingTypes.INCORPORATION_APPLICATION,
           id: filing.header.filingId,
@@ -925,8 +940,12 @@ export default {
 
     async getBcolMessage (errCode: string): Promise<any> {
       const fetchUrl = sessionStorage.getItem('PAY_API_URL') + 'codes/errors/' + errCode
-      const errObj = await axios.get(fetchUrl)
-      return errObj.data
+      const errObj = await axios.get(fetchUrl).catch(() => { return null })
+      if (errObj && errObj.data) {
+        return errObj.data
+      } else {
+        return null
+      }
     },
 
     showCommentDialog (filingId: number): void {
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -39,7 +39,8 @@
         :key="index"
         :class="{
           'disabled': !task.enabled,
-          'draft': isStatusDraft(task) && !isTypeCorrection(task)
+          'draft': isStatusDraft(task) && !isTypeCorrection(task) && !isBcolError(task),
+          'bcol-error': isBcolError(task)
         }"
       >
         <v-expansion-panel-header class="todo-item-toggle no-dropdown-icon">
@@ -83,12 +84,33 @@
                   </div>
                 </div>
 
-                <div v-if="isTypeCorrection(task) && isStatusDraft(task)" class="todo-subtitle">
+                <div v-else-if="isTypeCorrection(task) && isStatusDraft(task)" class="todo-subtitle">
                   <div>DRAFT</div>
                   <v-btn x-small icon class="expand-btn">
                     <v-icon>mdi-message-reply</v-icon>
                   </v-btn>
                   Detail{{task.comments.length > 1 ? "s" : ""}} ({{task.comments.length}})
+                </div>
+
+                 <div v-else-if="isBcolError(task)" class="todo-subtitle">
+                  <div>FILING PENDING</div>
+                  <div class="vert-pipe"></div>
+                  <div class="payment-status" v-if="inProcessFiling === task.id">
+                    PROCESSING...
+                  </div>
+                  <div class="payment-status" v-else>
+                    <span>PAYMENT UNSUCCESSFUL</span>
+                    <v-btn
+                      class="expand-btn"
+                      outlined
+                      color="red darken-4"
+                      :ripple=false
+                      @click.stop="togglePanel(index)"
+                    >
+                      <v-icon left>mdi-information-outline</v-icon>
+                      {{ (panel === index) ? "Hide Details" : "View Details" }}
+                    </v-btn>
+                  </div>
                 </div>
 
                 <div v-else-if="isStatusDraft(task)" class="todo-subtitle">
@@ -294,7 +316,13 @@
         </v-expansion-panel-header>
 
         <v-expansion-panel-content>
-          <div v-if="isTypeCorrection(task)">
+          <v-card v-if="isBcolError(task)" class="todo-list-detail bcol-todo-list-detail">
+            <v-card-text>
+            <p class="font-weight-bold black--text">Payment Incomplete - {{task.bcolErrObj.title}}</p>
+            <span v-html="task.bcolErrObj.detail"/>
+            </v-card-text>
+          </v-card>
+          <div v-else-if="isTypeCorrection(task)">
             <div v-if="isStatusDraft(task)" data-test-class="correction-draft" class="todo-list-detail">
               <p class="list-item__subtitle">
                 This filing is in review and has been saved as a draft.<br />
@@ -330,7 +358,7 @@
             </v-card-text>
           </v-card>
 
-          <v-card v-else-if="isStatusError(task)" data-test-class="payment-unsuccessful">
+          <v-card v-else-if="isBcolError(task)" data-test-class="payment-unsuccessful">
             <v-card-text>
               <p class="font-weight-bold black--text">Payment Unsuccessful</p>
               <p>This filing is pending payment. The payment appears to have been unsuccessful for some
@@ -445,17 +473,17 @@ export default {
   methods: {
     ...mapActions(['setARFilingYear', 'setCurrentFilingStatus']),
 
-    loadData () {
+    async loadData () {
       this.taskItems = []
       // If the Entity is a COOP, Enable the 'FileNow' Button without any user validation
       if (this.isCoop()) this.confirmCheckbox = true
 
       // create task items
-      this.tasks.forEach(task => {
+      this.tasks.forEach(async task => {
         if (task && task.task && task.task.todo) {
           this.loadTodoItem(task)
         } else if (task && task.task && task.task.filing) {
-          this.loadFilingItem(task)
+          await this.loadFilingItem(task)
         } else {
           // eslint-disable-next-line no-console
           console.log('ERROR - got unknown task =', task)
@@ -531,24 +559,24 @@ export default {
       }
     },
 
-    loadFilingItem (task) {
+    async loadFilingItem (task) {
       const filing = task.task.filing
       if (filing && filing.header) {
         switch (filing.header.name) {
           case FilingTypes.ANNUAL_REPORT:
-            this.loadAnnualReport(task)
+            await this.loadAnnualReport(task)
             break
           case FilingTypes.CHANGE_OF_DIRECTORS:
-            this.loadChangeOfDirectors(task)
+            await this.loadChangeOfDirectors(task)
             break
           case FilingTypes.CHANGE_OF_ADDRESS:
-            this.loadChangeOfAddress(task)
+            await this.loadChangeOfAddress(task)
             break
           case FilingTypes.CORRECTION:
-            this.loadCorrection(task)
+            await this.loadCorrection(task)
             break
           case FilingTypes.INCORPORATION_APPLICATION:
-            this.loadIncorporationApplication(task)
+            await this.loadIncorporationApplication(task)
             break
           default:
             // eslint-disable-next-line no-console
@@ -561,13 +589,15 @@ export default {
       }
     },
 
-    loadAnnualReport (task) {
+    async loadAnnualReport (task) {
       let date
       const filing = task.task.filing
       if (filing && filing.header && filing.annualReport) {
         filing.annualReport.annualReportDate
           ? date = filing.annualReport.annualReportDate
           : date = filing.annualReport.nextARDate
+        const bcolErr = filing.header.paymentErrorType || null
+        const bcolObj = await this.getBcolMessage(bcolErr)
         if (date) {
           const ARFilingYear = +date.substring(0, 4)
           this.taskItems.push({
@@ -579,7 +609,9 @@ export default {
             status: filing.header.status || FilingStatus.NEW,
             enabled: Boolean(task.enabled),
             order: task.order,
-            paymentToken: filing.header.paymentToken || null
+            paymentToken: filing.header.paymentToken || null,
+            paymentErrorType: bcolErr,
+            bcolErrObj: bcolObj
           })
         } else {
           // eslint-disable-next-line no-console
@@ -591,9 +623,11 @@ export default {
       }
     },
 
-    loadChangeOfDirectors (task) {
+    async loadChangeOfDirectors (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.changeOfDirectors) {
+        const bcolErr = filing.header.paymentErrorType || null
+        const bcolObj = await this.getBcolMessage(bcolErr)
         this.taskItems.push({
           type: FilingTypes.CHANGE_OF_DIRECTORS,
           id: filing.header.filingId,
@@ -602,7 +636,9 @@ export default {
           status: filing.header.status || FilingStatus.NEW,
           enabled: Boolean(task.enabled),
           order: task.order,
-          paymentToken: filing.header.paymentToken || null
+          paymentToken: filing.header.paymentToken || null,
+          paymentErrorType: bcolErr,
+          bcolErrObj: bcolObj
         })
       } else {
         // eslint-disable-next-line no-console
@@ -610,9 +646,11 @@ export default {
       }
     },
 
-    loadChangeOfAddress (task) {
+    async loadChangeOfAddress (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.changeOfAddress) {
+        const bcolErr = filing.header.paymentErrorType || null
+        const bcolObj = await this.getBcolMessage(bcolErr)
         this.taskItems.push({
           type: FilingTypes.CHANGE_OF_ADDRESS,
           id: filing.header.filingId,
@@ -621,7 +659,9 @@ export default {
           status: filing.header.status || FilingStatus.NEW,
           enabled: Boolean(task.enabled),
           order: task.order,
-          paymentToken: filing.header.paymentToken || null
+          paymentToken: filing.header.paymentToken || null,
+          paymentErrorType: bcolErr,
+          bcolErrObj: bcolObj
         })
       } else {
         // eslint-disable-next-line no-console
@@ -629,9 +669,11 @@ export default {
       }
     },
 
-    loadCorrection (task) {
+    async loadCorrection (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.correction) {
+        const bcolErr = filing.header.paymentErrorType || null
+        const bcolObj = await this.getBcolMessage(bcolErr)
         this.taskItems.push({
           type: FilingTypes.CORRECTION,
           id: filing.header.filingId,
@@ -653,9 +695,11 @@ export default {
       }
     },
 
-    loadIncorporationApplication (task) {
+    async loadIncorporationApplication (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.incorporationApplication) {
+        const bcolErr = filing.header.paymentErrorType || null
+        const bcolObj = await this.getBcolMessage(bcolErr)
         this.taskItems.push({
           type: FilingTypes.INCORPORATION_APPLICATION,
           id: filing.header.filingId,
@@ -664,7 +708,9 @@ export default {
           status: filing.header.status,
           enabled: Boolean(task.enabled),
           order: task.order,
-          paymentToken: filing.header.paymentToken || null
+          paymentToken: filing.header.paymentToken || null,
+          paymentErrorType: bcolErr,
+          bcolErrObj: bcolObj
         })
       } else {
         // eslint-disable-next-line no-console
@@ -877,6 +923,12 @@ export default {
       })
     },
 
+    async getBcolMessage (errCode: string): Promise<any> {
+      const fetchUrl = sessionStorage.getItem('PAY_API_URL') + 'codes/errors/' + errCode
+      const errObj = await axios.get(fetchUrl)
+      return errObj.data
+    },
+
     showCommentDialog (filingId: number): void {
       this.currentFilingId = filingId
       this.addCommentDialog = true
@@ -888,6 +940,10 @@ export default {
         // emit dashboard reload trigger event
         this.$root.$emit('triggerDashboardReload')
       }
+    },
+
+    isBcolError (task: any): boolean {
+      return task.paymentErrorType && task.paymentErrorType.startsWith('BCOL')
     },
 
     isPriority (priority: boolean): string {
@@ -1070,5 +1126,13 @@ export default {
 
 .todo-item-toggle h3 {
   margin-bottom: 0.25rem;
+}
+
+.bcol-error {
+  border-top: solid #a94442 3px;
+}
+
+::v-deep .bcol-todo-list-detail {
+  background-color: #f1f1f1 !important;
 }
 </style>
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -470,7 +470,7 @@ export default {
   methods: {
     ...mapActions(['setARFilingYear', 'setCurrentFilingStatus']),
 
-    async loadData () {
+    loadData () {
       this.taskItems = []
       // If the Entity is a COOP, Enable the 'FileNow' Button without any user validation
       if (this.isCoop()) this.confirmCheckbox = true
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -600,7 +600,7 @@ export default {
         const bcolErr = filing.header.paymentErrorType || null
         let bcolObj = null
         if (bcolErr) {
-          bcolObj = await this.getErrorObj(this.payApiUrl, bcolErr)
+          bcolObj = await this.getErrorObj(bcolErr)
         }
         if (date) {
           const ARFilingYear = +date.substring(0, 4)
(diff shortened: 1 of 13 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -877,6 +923,12 @@ export default {
       })
     },
 
+    async getBcolMessage (errCode: string): Promise<any> {
+      const fetchUrl = sessionStorage.getItem('PAY_API_URL') + 'codes/errors/' + errCode
+      const errObj = await axios.get(fetchUrl)
+      return errObj.data
```
ochiu, line 929: Might consider putting this in a mixin so we can re-use the mixin class between apps easily. Also probably should add some error handling and maybe present some kind of generic message if fetching the error message fails.

Might need to ask Scott what he wants ( 'Details currently unavailable.'?).

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -625,9 +637,19 @@ export default {
...
           filing = res.data.filing
           this.haveChanges = false
-        }).catch(error => {
+        }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
-            this.paymentErrorDialog = true
+            if (error.response.data && error.response.data.errors) {
+              const msgCode = error.response.data.errors.find(x => x.payment_error_type.startsWith('BCOL'))
+              if (msgCode) {
+                const errObj = await this.getErrorObj(msgCode.payment_error_type)
```
ochiu, line 645: Looks like might be some potential for mixin logic as it is duplicated across components

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -673,6 +705,13 @@ export default {
       }
     },
 
+    async getErrorObj (errCode) {
+      const fetchUrl = this.payApiUrl + 'codes/errors/BCOL_ERROR'
+      const errObj = await axios.get(fetchUrl)
+      console.log(errObj)
+      return errObj.data
+    },
```
ochiu, line 713: Could use the mixin call instead to grab an error code so you wouldn't have to duplicate this function.
Also, it looks like BCOL_ERROR is hardcoded

src/views/AnnualReport.vue
```diff
@@ -859,9 +869,19 @@ export default {
           }
           filing = res.data.filing
           this.haveChanges = false
-        }).catch(error => {
+        }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
-            this.paymentErrorDialog = true
+            if (error.response.data && error.response.data.errors) {
+              const msgCode = error.response.data.errors.find(x => x.payment_error_type.startsWith('BCOL'))
```
ochiu, line 875: Just a question, will we always looking for BCOL? is it possible we have other payment error types and we want to just display it if there is a payment error?

ochiu:
Initial draft review looks good, just a few comments to potentially consolidate some code for re-usability.

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,26 @@
+import { Component, Vue } from 'vue-property-decorator'
+import { AlertMessageIF } from '@/interfaces'
+import axios from '@/axios-auth'
+
+@Component({})
+export default class BcolMixin extends Vue {
+  async getErrorObj (url, errCode) {
```
ochiu, line 7: probably don't need the url parameter, you are already fetching it from session storage.

src/views/AnnualReport.vue
peter-freshworks: Yup, these are BCOL specific errors, and they all start with BCOL. I don't know if we have a future plan to store/retrieve all payment error HTML on the Pay API side

src/mixins/bcol-mixin.ts
peter-freshworks: removed

src/components/Dashboard/TodoList.vue
peter-freshworks: done

src/components/Dashboard/TodoList.vue
```diff
@@ -83,14 +84,35 @@
                   </div>
                 </div>
 
-                <div v-if="isTypeCorrection(task) && isStatusDraft(task)" class="todo-subtitle">
+                <div v-else-if="isTypeCorrection(task) && isStatusDraft(task)" class="todo-subtitle">
```
severinbeauvais, line 87: No. The v-if above this is checks for a subtitle, so THIS line is not an else case of that.

src/components/Dashboard/TodoList.vue
```diff
@@ -83,14 +84,35 @@
...
                     <v-icon>mdi-message-reply</v-icon>
                   </v-btn>
                   Detail{{task.comments.length > 1 ? "s" : ""}} ({{task.comments.length}})
                 </div>
 
+                 <div v-else-if="isBcolError(task)" class="todo-subtitle">
+                  <div>FILING PENDING</div>
+                  <div class="vert-pipe"></div>
+                  <div class="payment-status" v-if="inProcessFiling === task.id">
+                    PROCESSING...
```
severinbeauvais, line 99: Although this is labelled "pending", I think it's not the same as a true pending, where the BE may still be doing work, so the FE retries 10X for an update (which is where the "PROCESSING..." comes in. In this case I think there is no processing state.

src/components/Dashboard/TodoList.vue
```diff
@@ -294,7 +316,13 @@
         </v-expansion-panel-header>
 
         <v-expansion-panel-content>
-          <div v-if="isTypeCorrection(task)">
+          <v-card v-if="isBcolError(task)" class="todo-list-detail bcol-todo-list-detail">
```
severinbeauvais, line 319: I'm working on this code atm (in #3306) so you will have to make some of my changes here when you merge after me. Basically the expansion content is now a div instead of a v-card and uses some classes for styling. Your best bet will be to see what the other expansion blocks look like and copy that.

src/components/Dashboard/TodoList.vue
```diff
@@ -294,7 +316,13 @@
...
+            <v-card-text>
+            <p class="font-weight-bold black--text">Payment Incomplete - {{task.bcolErrObj.title}}</p>
```
severinbeauvais, line 321: And this I've changed to a h4.

src/components/Dashboard/TodoList.vue
```diff
@@ -330,7 +358,7 @@
             </v-card-text>
           </v-card>
 
-          <v-card v-else-if="isStatusError(task)" data-test-class="payment-unsuccessful">
+          <v-card v-else-if="isBcolError(task)" data-test-class="payment-unsuccessful">
```
severinbeauvais, line 361: This is a duplicate "if" case of line 319. I think this line needs to be changed back to check for isStatusError(), to handle non-BCOL errors.

src/components/Dashboard/TodoList.vue
```diff
@@ -408,6 +436,7 @@ export default {
       confirmEnabled: false,
       currentFilingId: null,
       panel: null as number, // currently expanded panel
+      payAPiUrl: sessionStorage.getItem('PAY_API_URL'),
```
severinbeauvais, line 439: I'd rather see this as a getter method, as we do elsewhere.

I believe by declaring this here, it's assigned when the component is created, so the session item needs to exist before that. Instead of making that assumption, by making this a getter, it becomes reactive, which theoretically means it's updated where its content changes, so it's always up to date.

src/components/Dashboard/TodoList.vue
```diff
@@ -561,13 +590,18 @@ export default {
...
       const filing = task.task.filing
       if (filing && filing.header && filing.annualReport) {
         filing.annualReport.annualReportDate
           ? date = filing.annualReport.annualReportDate
           : date = filing.annualReport.nextARDate
+        const bcolErr = filing.header.paymentErrorType || null
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getErrorObj(bcolErr)
+        }
```
severinbeauvais, line 604: If you don't like the let and the separate if statement, in Javascript you can do cool things like this:
`const bcolObj = bcolErr && await this.getErrorObj(bcolErr)`

src/components/Dashboard/TodoList.vue
```diff
@@ -890,6 +952,10 @@ export default {
       }
     },
 
+    isBcolError (task: any): boolean {
+      return task.paymentErrorType && task.paymentErrorType.startsWith('BCOL')
+    },
```
severinbeauvais, line 957: FYI, you can also now write this as
`return task.paymentErrorType?.startsWith('BCOL')`

src/components/Dashboard/TodoList.vue
```diff
@@ -294,7 +316,13 @@
         </v-expansion-panel-header>
 
         <v-expansion-panel-content>
-          <div v-if="isTypeCorrection(task)">
+          <v-card v-if="isBcolError(task)" class="todo-list-detail bcol-todo-list-detail">
+            <v-card-text>
+            <p class="font-weight-bold black--text">Payment Incomplete - {{task.bcolErrObj.title}}</p>
+            <span v-html="task.bcolErrObj.detail"/>
```
severinbeauvais, line 322: As there are no checks anywhere else that bcolErrorObj is not null, please add a check here before de-referencing detail.

src/components/Dashboard/TodoList.vue
```diff
@@ -579,7 +613,9 @@ export default {
             status: filing.header.status || FilingStatus.NEW,
             enabled: Boolean(task.enabled),
             order: task.order,
-            paymentToken: filing.header.paymentToken || null
+            paymentToken: filing.header.paymentToken || null,
+            paymentErrorType: bcolErr,
+            bcolErrObj: bcolObj
```
severinbeauvais, line 618: Because we only care about payment error type "BCOL..." at the moment, did you consider merging the 2 new properties into a single one? For instance, "if bcolErrObj isn't null then this is a BCOL error". Then you don't need both a flag and the object. Just a thought.

src/components/dialogs/BcolErrorDialog.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <v-dialog v-model="dialog" width="45rem" persistent :attach="attach" content-class="payment-error-dialog">
```
severinbeauvais, line 2: Update the content class.

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, Vue } from 'vue-property-decorator'
+import axios from '@/axios-auth'
+
+@Component({})
+export default class BcolMixin extends Vue {
+  async getErrorObj (errCode) {
+    const fetchUrl = sessionStorage.getItem('PAY_API_URL') + 'codes/errors/' + errCode
```
severinbeauvais, line 7: The Pay API URL should be a getter so you don't have to access session storage every time. (Getters are reactive and cached.)

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,24 @@
...
+    const errObj = await axios.get(fetchUrl)
```
severinbeauvais, line 8: You need to handle network error exceptions. You could use `.catch()`, or wrap this in a try..catch block, or handle this in the calling code.

For an example of this, see https://github.com/bcgov/bcrs-business-create-ui/blob/7a405e7c1a14fa821a8cf4df8dca34ffd55248ba/src/components/common/Actions.vue#L118

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (errObj && errObj.data) {
+      return errObj.data
+    } 
+    return null
+  }
+
+  getErrorCode (error) {
+    if (error && error.response && error.response.data && error.response.data.errors) {
```
severinbeauvais, line 16: Use optional chaining :)

src/views/AnnualReport.vue
```diff
@@ -859,9 +869,19 @@ export default {
...
+        }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
-            this.paymentErrorDialog = true
+            const errCode = this.getErrorCode(error)
+            if (errCode) {
+              const errObj = await this.getErrorObj(errCode.payment_error_type)
+              if (errObj) {
+                this.bcolErrMsg = errObj.detail
+                this.bcolTitle = errObj.title
+                this.bcolErrorDialog = true
```
severinbeauvais, line 880: Instead of having 3 different properties for this dialog, have you considered merging them? For instance "if errorObj isn't null then display the dialog". Then you only have to pass a single prop to the dialog. Then, inside the dialog, since you won't be able to use the boolean as a v-model, instead create a boolean getter (true if errorObj isn't null) and use that as a value prop to display the dialog (see https://vuetifyjs.com/en/components/dialogs/). Ask if you want a code example.

src/components/Dashboard/TodoList.vue
```diff
@@ -591,9 +627,14 @@ export default {
...
     },
 
-    loadChangeOfDirectors (task) {
+    async loadChangeOfDirectors (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.changeOfDirectors) {
+        const bcolErr = filing.header.paymentErrorType || null
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getErrorObj(bcolErr)
```
severinbeauvais, line 636: Did you consider returning the error object in the filing object so we don't have to query it separately here?

src/views/AnnualReport.vue
```diff
@@ -29,6 +29,13 @@
       attach="#annual-report"
     />
 
+    <bcol-error-dialog
+      :dialog="bcolErrorDialog"
+      :title="bcolTitle"
+      :errMsg="bcolErrMsg"
```
severinbeauvais, line 35: See my other comment about merging these properties into 1.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -594,9 +607,19 @@ export default {
           }
           filing = res.data.filing
           this.haveChanges = false
-        }).catch(error => {
+        }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
-            this.paymentErrorDialog = true
+            const errCode = this.getErrorCode(error)
```
severinbeauvais, line 612: Since you're not awaiting getErrorCode(), this catch arrow function doesn't need to be async.

tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -1385,3 +1385,141 @@ describe('Standalone Directors Filing - Part 6 - Error/Warning Dialogs', () => {
...
+      detail: 'An Error has occured',
+      title: 'Error'
+    }})))
+    // click the File & Pay button
+    await button.trigger('click')
+    await flushPromises()
+    await vm.onClickFilePay()
+    // work-around because click trigger isn't working
+
+    // verify redirection
```
severinbeauvais, line 1516: Update this comment.

src/components/Dashboard/TodoList.vue
peter-freshworks: ah like C# :D

src/views/AnnualReport.vue
peter-freshworks: I actually did consider this, but I didn't see too many components taking advantage of passing an object/dto as a parameter so I figured I should keep them separate. Will update

src/components/Dashboard/TodoList.vue
peter-freshworks: eek, good catch

src/components/Dashboard/TodoList.vue
peter-freshworks: done

src/components/Dashboard/TodoList.vue
peter-freshworks: I like it!

src/components/Dashboard/TodoList.vue
peter-freshworks: done

src/components/Dashboard/TodoList.vue
peter-freshworks: done

src/components/dialogs/BcolErrorDialog.vue
peter-freshworks: done

src/mixins/bcol-mixin.ts
peter-freshworks: done

src/mixins/bcol-mixin.ts
peter-freshworks: done

src/mixins/bcol-mixin.ts
peter-freshworks: done

src/components/dialogs/BcolErrorDialog.vue
peter-freshworks: done

src/mixins/bcol-mixin.ts
peter-freshworks: done

src/components/Dashboard/TodoList.vue
peter-freshworks: TBD in the future

severinbeauvais: As discussed on RocketChat, although often it's better for the API to gather up and package up all the necessary data (and not require the UI to make additional network calls), in this case the additional calls are only if there is a payment error code, so it's not a performance issue worth worrying about.

No change needed here. I'm resolving this conversation.

src/components/Dashboard/TodoList.vue
severinbeauvais: Yes, like that. After all, they're both from Microsoft :)

src/views/StandaloneOfficeAddressFiling.vue
peter-freshworks: the async is necessary if there is an error code, since getErrorOBj is awaited (if it exists), is it not? (line 611)

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -179,7 +186,7 @@ import axios from '@/axios-auth'
 import { mapActions, mapState, mapGetters } from 'vuex'
 
 // Dialogs
-import { ConfirmDialog, PaymentErrorDialog, ResumeErrorDialog, SaveErrorDialog } from '@/components/dialogs'
+import { ConfirmDialog, PaymentErrorDialog, ResumeErrorDialog, SaveErrorDialog, BcolErrorDialog } from '@/components/dialogs'
```
severinbeauvais, line 189: GHA line length warning here (see files in GH).

src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: Oops, I missed the getErrorObj() call a couple of lines later. You are correct! I'll resolve this comment now :]

src/components/Dashboard/TodoList.vue
severinbeauvais: Since Ody is away this week, I confirm that this code was satisfactorily moved to a mixin.

However, it looks like there is no fallback if the fetch returns no errorObj. Is this a conscious decision? Or does Scott suggest something here?

src/components/Dashboard/TodoList.vue
```diff
@@ -561,13 +590,17 @@ export default {
...
+    async loadAnnualReport (task) {
       let date
       const filing = task.task.filing
       if (filing && filing.header && filing.annualReport) {
         filing.annualReport.annualReportDate
           ? date = filing.annualReport.annualReportDate
           : date = filing.annualReport.nextARDate
+
+        const bcolErr = filing.header.paymentErrorType || null
+        const bcolObj = bcolErr && await this.getErrorObj(bcolErr)
```
severinbeauvais, line 602: If getObjError() fails (network error) then an exception is generated but is not caught here, so the caller (loadAnnualReport) will exit and all callers upstream until the exception is caught or Javascript exits. Please add `.catch()` or wrap this in `try..catch` block.

src/views/AnnualReport.vue
severinbeauvais: Since Ody is away this week, I'll say this looks satisfactorily handled and am closing this conversation.

src/mixins/bcol-mixin.ts
severinbeauvais: Since Ody is away this week, I'll say this looks satisfactorily handled and am closing this conversation.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
severinbeauvais: Last person to merge will have to handle this.

src/components/Dashboard/TodoList.vue
severinbeauvais: Last person to merge will have to handle this.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
severinbeauvais: Code was changed. Comment no longer applies.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed, although I think we should have a fallback if this is a BCOL error but we don't have the error object or title or details -- see Ody's comment above.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed by refactor.

src/components/dialogs/BcolErrorDialog.vue
severinbeauvais: Fixed.

src/mixins/bcol-mixin.ts
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
severinbeauvais: Never mind. I see you're now catching errors in getErrorObj().

src/mixins/bcol-mixin.ts
severinbeauvais: I thought you might handle the error instead of ignoring it. If the intent is truly to ignore the error (and I don't have a problem with this) then please add a comment that you're explicitly ignoring the error, so it doesn't look like an oversight. Thanks.

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,30 @@
...
+    const fetchUrl = this.payApi + 'codes/errors/' + errCode
+    const errObj = await axios.get(fetchUrl).catch()
+    if (errObj && errObj.data) {
+      return errObj.data
+    }
+    return null
+  }
+
+  getErrorCode (error) {
+    if (error.response?.data?.errors) {
```
severinbeauvais, line 18: Code looks good but what if `error` itself is null?

src/mixins/bcol-mixin.ts
severinbeauvais: Fixed.

src/components/dialogs/BcolErrorDialog.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <v-dialog v-model="bcolDialog" width="45rem" persistent :attach="attach" content-class="bcol-error-dialog">
```
severinbeauvais, line 2: When I wrote similar code, I used `:value=` instead of v-model. This may not work here... I'm guessing when you close the dialog, it will try to set bcolDialog (to false), but that property is readonly (getter). I recomment you just use `:value` here. And also test this :)

src/views/AnnualReport.vue
severinbeauvais: Fixed. Looks way cleaner!

src/views/AnnualReport.vue
severinbeauvais: Fixed.

tests/unit/StandaloneDirectorsFiling.spec.ts
severinbeauvais: 2 out of 3 were fixed. Please also fix line 1867 in StandaloneOfficeAddressFiling.spec.ts.

src/views/AnnualReport.vue
```diff
@@ -311,10 +317,10 @@ import SbcFeeSummary from 'sbc-common-components/src/components/SbcFeeSummary.vu
 import { Certify, OfficeAddresses, StaffPayment, SummaryDirectors, SummaryOfficeAddresses } from '@/components/common'
 
 // Dialogs
-import { ConfirmDialog, PaymentErrorDialog, ResumeErrorDialog, SaveErrorDialog } from '@/components/dialogs'
+import { ConfirmDialog, PaymentErrorDialog, ResumeErrorDialog, SaveErrorDialog, BcolErrorDialog } from '@/components/dialogs'
```
severinbeauvais, line 320: GHA line length warning here (see file in GH).

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -313,10 +319,10 @@ import SbcFeeSummary from 'sbc-common-components/src/components/SbcFeeSummary.vu
 import { Certify, StaffPayment, SummaryDirectors, SummaryCertify, SummaryStaffPayment } from '@/components/common'
 
 // Dialog Components
-import { ConfirmDialog, PaymentErrorDialog, ResumeErrorDialog, SaveErrorDialog } from '@/components/dialogs'
+import { ConfirmDialog, PaymentErrorDialog, ResumeErrorDialog, SaveErrorDialog, BcolErrorDialog } from '@/components/dialogs'
```
severinbeauvais, line 322: GHA line length warning here (see file in GH).

src/components/Dashboard/TodoList.vue
```diff
@@ -431,6 +456,10 @@ export default {
       return sessionStorage.getItem('BUSINESS_ID')
     },
 
+    payAPiUrl (): string {
```
severinbeauvais, line 459: This computed method is now obsolete.

src/components/Dashboard/TodoList.vue
```diff
@@ -621,17 +664,24 @@ export default {
...
 
-    loadCorrection (task) {
+    async loadCorrection (task) {
       const filing = task.task.filing
       if (filing && filing.header && filing.correction) {
+        const bcolErr = filing.header.paymentErrorType || null
+        let bcolObj = null
+        if (bcolErr) {
+          bcolObj = await this.getErrorObj(bcolErr)
+        }
```
severinbeauvais, line 684: It looks like this code is not used for anything.

src/components/dialogs/BcolErrorDialog.vue
```diff
@@ -0,0 +1,62 @@
...
+})
+export default class BcolErrorDialog extends Vue {
+  // Getter definition for static type checking.
+  readonly isRoleStaff!: boolean
+
+  @Prop() private filingType: string
+
+  @Prop({ default: () => { return null } }) private bcolObject: object
+
+  @Prop() title: string
```
severinbeauvais, line 46: Looks like this prop is now obsolete.

src/components/Dashboard/TodoList.vue
peter-freshworks: it's also named incorrectly :p will remove

src/views/AnnualReport.vue
```diff
@@ -29,6 +29,12 @@
       attach="#annual-report"
     />
 
+    <bcol-error-dialog
+      :bcolOBject="bcolObj"
+      filingType="Annual Report"
```
severinbeauvais, line 34: You're not actually passing a "type" here; you're passing a name. There's a mixin that can help you clean this up! Search for `filingTypeToName` method.

peter-freshworks: I thought something like this existed, couldn't track it down...thanks!

src/views/AnnualReport.vue
```diff
@@ -859,9 +866,17 @@ export default {
...
           this.haveChanges = false
-        }).catch(error => {
+        }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
-            this.paymentErrorDialog = true
+            const errCode = this.getErrorCode(error)
+            if (errCode) {
+              const errObj = await this.getErrorObj(errCode.payment_error_type)
+              if (errObj) {
+                this.bcolObj = errObj
```
severinbeauvais, line 875: This "if" is not needed. It looks like it assigns errorObj if it's not null, which is the same thing that would happen without the if... unless you want to set bcolObj=null in an else block?

src/views/AnnualReport.vue
```diff
@@ -914,6 +937,7 @@ export default {
 
     resetErrors () {
       this.saveErrorDialog = false
+      this.bcolErrorDialog = false
```
severinbeauvais, line 940: Obsolete code.

Also in other 2 filings.

peter-freshworks: done

tests/unit/AnnualReport.spec.ts
```diff
@@ -2103,3 +2103,149 @@ describe('AnnualReport - Part 7 - Concurrent Saves', () => {
...
+      title: 'Error'
+    }})))
+    // click the File & Pay button
+    await button.trigger('click')
+    await flushPromises()
+    await vm.onClickFilePay()
+    // work-around because click trigger isn't working
+
+    // verify an error has been received
+    expect(vm.bcolObj?.detail?.length).toBeGreaterThan(0)
```
severinbeauvais, line 2245: Please create a new unit test file for your new dialog. The tests should verify that the different props generate the expected function/output.

src/mixins/bcol-mixin.ts
severinbeauvais: Looks like nothing is returned by this method if this statement is false.

Actually, if you move the optional chaining inside line 19 (const msgCode...) then you probably don't need the "if" and it will all work out.

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,30 @@
+import { Component, Vue } from 'vue-property-decorator'
+import axios from '@/axios-auth'
+
+@Component({})
+export default class BcolMixin extends Vue {
+  // readonly payApiUrl: string
+
+  async getErrorObj (errCode) {
```
severinbeauvais, line 8: Please add return types to all methods in this mixin.

peter-freshworks: I considered this, but I didn't see a test file for paymenterrordialog. Will do

tests/unit/AnnualReport.spec.ts
peter-freshworks: I considered this, but I didn't see a test file for paymenterrordialog. Will do

src/views/AnnualReport.vue
peter-freshworks: done

tests/unit/TodoList.spec.ts
```diff
@@ -1516,6 +1516,61 @@ describe('TodoList - Click Tests', () => {
...
+
+    await flushPromises()
+
+    await Vue.nextTick()
+
+    const todoItem = vm.$el.querySelector('.bcol-error')
+
+    const button = wrapper.find('.expand-btn')
+
+    button.trigger('click')
```
severinbeauvais, line 1565: Great unit test! Could you please add a few comments to narrate what the test is doing (as with other unit tests)? Thx.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/dialogs/BcolErrorDialog.vue
peter-freshworks: done

src/mixins/bcol-mixin.ts
peter-freshworks: done

tests/unit/AnnualReport.spec.ts
severinbeauvais: There are other files for other dialogs. Check them out for guidance.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/dialogs/BcolErrorDialog.vue
severinbeauvais: Fixed.

src/views/AnnualReport.vue
severinbeauvais: Fixed.

src/mixins/bcol-mixin.ts
severinbeauvais: Check for null `error` is now implemented but method still has no return value if check fails.

src/mixins/bcol-mixin.ts
severinbeauvais: Fixed.

src/views/AnnualReport.vue
severinbeauvais: Fixed.

src/views/AnnualReport.vue
severinbeauvais: Fixed.

src/views/AnnualReport.vue
severinbeauvais: Fixed.

src/views/StandaloneDirectorsFiling.vue
severinbeauvais: Fixed.

src/components/dialogs/BcolErrorDialog.vue
severinbeauvais: Fixed.

src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: Fixed.

src/components/Dashboard/TodoList.vue
```diff
@@ -621,15 +660,17 @@ export default {
...
+          bcolErrObj: bcolObj
         })
       } else {
         // eslint-disable-next-line no-console
         console.log('ERROR - invalid filing or header or changeOfAddress in task =', task)
       }
     },
 
-    loadCorrection (task) {
+    async loadCorrection (task) {
```
severinbeauvais, line 673: Isn't async any more.

peter-freshworks: whoops

src/components/dialogs/BcolErrorDialog.vue
```diff
@@ -0,0 +1,62 @@
...
+    ...mapGetters(['isRoleStaff'])
+  },
+  mixins: [EnumMixin],
+  components: { ErrorContact }
+})
+export default class BcolErrorDialog extends Vue {
+  // Getter definition for static type checking.
+  readonly isRoleStaff!: boolean
+
+  @Prop() private filingType: string
```
severinbeauvais, line 44: Type is now an enum, not string. You'll have to import it.

src/views/AnnualReport.vue
```diff
@@ -888,9 +901,14 @@ export default {
...
           this.haveChanges = false
-        }).catch(error => {
+        }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
-            this.paymentErrorDialog = true
+            const errCode = this.getErrorCode(error)
+            if (errCode) {
+              this.bcolObj = await this.getErrorObj(errCode.payment_error_type)
+            } else {
+              this.paymentErrorDialog = true
```
severinbeauvais, line 910: In `resetErrors()`, should clear this.bcolObj and this.paymentErrorDialog.

Same in other 2 filings.

peter-freshworks: whoops

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed.

src/components/dialogs/BcolErrorDialog.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <v-dialog :value="bcolDialog" width="45rem" persistent :attach="attach" content-class="bcol-error-dialog">
+    <v-card v-if="bcolObject">
+      <v-card-title id="dialog-title">Payment Incomplete - {{bcolObject.title}}</v-card-title>
+
+      <v-card-text>
+        <p class="genErr" id="dialog-header">This {{filingTypeToName(filingType)}} could not be filed for the following reason:</p>
```
severinbeauvais, line 7: Line too long.

In the future, have a look at your own files (https://github.com/bcgov/business-filings-ui/pull/75/files) and watch for these. These should be triggering lint errors but I don't know why they're not.

src/mixins/bcol-mixin.ts
severinbeauvais: 2 out of 3 are done. Please also do `get payApi ()`.

severinbeauvais: Do you use VS plugins Vetur and Vue Peek? Those might not identify that a return type is missing, but they should identify errors if you specify one type and use it differently.

src/components/dialogs/BcolErrorDialog.vue
severinbeauvais: Fixed.

src/views/AnnualReport.vue
severinbeauvais: Fixed.

tests/unit/AnnualReport.spec.ts
severinbeauvais: Done, great!

tests/unit/StandaloneDirectorsFiling.spec.ts
severinbeauvais: Fixed.

src/mixins/bcol-mixin.ts
peter-freshworks: the second return statement should cover that case, no?

src/mixins/bcol-mixin.ts
peter-freshworks: I do not use either of those :/ I will install them...and whoops I guess I missed this one

tests/unit/TodoList.spec.ts
peter-freshworks: done

src/components/dialogs/BcolErrorDialog.vue
peter-freshworks: done...and yeah I didn't have any linting errors

severinbeauvais: Maybe those warnings show up if you do `npm run build` ?

severinbeauvais: Fixed.

src/mixins/bcol-mixin.ts
severinbeauvais: The second return statement is inside "if (error?...)" block. So if this "if" fails, you don't return anything.

tests/unit/TodoList.spec.ts
severinbeauvais: Fixed.

src/mixins/bcol-mixin.ts
peter-freshworks: oh shoot I was looking at the other method (getErrorObj)...similar names, sorry. Will fix!

src/components/Dashboard/TodoList.vue
```diff
@@ -445,17 +470,17 @@ export default {
   methods: {
     ...mapActions(['setARFilingYear', 'setCurrentFilingStatus']),
 
-    loadData () {
+    async loadData () {
```
cameron-freshworks, line 473: I'm not 100% sure if you need this async, when you've applied it to the scope of the callback below?

src/components/Dashboard/TodoList.vue
```diff
@@ -1071,4 +1121,12 @@ export default {
 .todo-item-toggle h3 {
   margin-bottom: 0.25rem;
 }
+
+.bcol-error {
+  border-top: solid #a94442 3px;
```
cameron-freshworks, line 1126: If this colour is something that matches one of the theme colours in the assets folder, then perhaps use of of those? I haven't seen the designs, but if it's something new/custom for the BCOL messages and we use it more than once, perhaps we could add it to the assets for re-use!

src/components/Dashboard/TodoList.vue
```diff
@@ -1071,4 +1121,12 @@ export default {
...
+}
+
+.bcol-todo-list-detail {
+  background-color: #f1f1f1 !important;
```
cameron-freshworks, line 1130: Same as above 👍

src/components/dialogs/BcolErrorDialog.vue
```diff
@@ -0,0 +1,66 @@
...
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+import { mapGetters } from 'vuex'
+import { ErrorContact } from '@/components/common'
+import { EnumMixin } from '@/mixins'
+import { FilingTypes } from '@/enums'
+@Component({
```
cameron-freshworks, line 36: Might want a space between the imports and the component decorator.

cameron-freshworks: Component looks good though, nice and clean! Good job 👍

src/mixins/bcol-mixin.ts
```diff
@@ -0,0 +1,30 @@
+import { Component, Vue } from 'vue-property-decorator'
+import axios from '@/axios-auth'
+
+@Component({})
+export default class BcolMixin extends Vue {
+  async getErrorObj (errCode): Promise<any> {
```
cameron-freshworks, line 6: The methods seem pretty straight forward so i'm not sure if you absolutely NEED doc bloc comments, but i find in a mixin for example, where the sole purpose is to export these methods to use elsewhere, it never hurts :)

src/components/Dashboard/TodoList.vue
peter-freshworks: you are correct! removed

src/components/Dashboard/TodoList.vue
peter-freshworks: right now, it's pretty isolated to this page and this scenario, so I don't know how much value there would be?

src/components/Dashboard/TodoList.vue
peter-freshworks: also same as above :D

src/components/dialogs/BcolErrorDialog.vue
peter-freshworks: will do!

src/mixins/bcol-mixin.ts
peter-freshworks: Severin wants them, so in they go!

src/components/Dashboard/TodoList.vue
severinbeauvais: As merged before mine, I will update this.

src/components/Dashboard/TodoList.vue
severinbeauvais: As this PR was merged before mine, I will update this.

src/mixins/bcol-mixin.ts
severinbeauvais: @cameron-freshworks I wanted the comments to explain the `.catch()` (so it didn't look like an oversight).

PR #76, opened 2020-05-14
updated dashboard for FE filings + other tweaks and fixes

severinbeauvais:
*Issue #:* /bcgov/entity#3306

*Description of changes:*
- added safety check for 1 task xor 1 filing
- cleaned up task loading for NRs/IAs
- cleaned up filing loading for pending IAs
- set temporary addresses and directors for pending IAs
- conditionally grayed out/disabled dashboard office addresses list (per Scott)
- tightened up spacing in office addresses list (OK'ed by Scott)
- displayed Records Office for Corps (work-around for current test NRs and also correct in future per Scott)
- conditionally grayed out/disabled dashboard current directors list (per Scott)
- cleaned up filing history list layout (classes, etc)
- cleaned up filing history list icon/button displays (now look like buttons with hover style)
- added support for Future Effective Incorporations in filing history list
- tightened up all vertical spacing in Filing History List
- cleaned up document (output) code for easier future refactor (ie, list of reports)
- moved v-dividers out of sub-components and into parents
- cleaned up existing filing code and assumptions (ie, defensive coding)
- converted future effective date from UTC to local
- fixed future effective COA unit test (UTC-local issue)
- updated download button states
- updated IA report file names
- added sub-component for FE Incorporation filing
- cleaned up misc sub-component layouts
- cleaned up todo list layout (classes, etc)
- cleaned up todo list icon/button displays (now look like buttons with hover style)
- added missing Entity Status for pending IA filing
- cleaned up name request info (ie, reuse common util)
- updated dashboard to manage new properties
- updated some unit tests to async usage
- updated Address List unit tests
- updated Director List unit tests
- updated and added Filing History List unit tests (for FE)
- added unit tests for Future Effective Filing sub-component


src/App.vue
```diff
@@ -464,11 +464,17 @@ export default {
...
-          } else {
-            this.setEntityStatus(EntityStatus.INCORPORATION_APPLICATION)
+          switch (tasks[0].task?.todo?.header?.name) {
+            case FilingTypes.NAME_REQUEST:
+              // this is a Name Request
+              this.setEntityStatus(EntityStatus.NAME_REQUEST)
+              break
+            case FilingTypes.INCORPORATION_APPLICATION:
+              // this is a Draft Incorporation Application
+              this.setEntityStatus(EntityStatus.DRAFT_INCORP_APP)
```
cameron-freshworks, line 476: Just curious, what is the difference between a Name Request and Incorporation application. I thought a name request was a draft incorporation application?

src/components/Dashboard/AddressListSm.vue
```diff
@@ -1,16 +1,23 @@
...
+      :value="(showGrayedOut || showCompleteYourFilingMessage) ? [0,1] : [0]"
+      :disabled="(showGrayedOut || showCompleteYourFilingMessage)"
+    >
       <!-- Registered Office -->
       <v-expansion-panel id="registered-office-panel"
         class="align-items-top"
-        :class="{ 'address-overlay': coaPending }"
+        :class="{
+          'address-overlay': coaPending,
+          'disabled': showGrayedOut
```
cameron-freshworks, line 13: Cool! Does the CoaPending state still work when a bcomps coa is pending? 😆  I only ask, because there is a ticket remaining to look into that :)

src/components/Dashboard/FutureEffectiveFiling.vue
```diff
@@ -0,0 +1,56 @@
...
+<script lang="ts">
+import { Component, Vue, Prop } from 'vue-property-decorator'
+import { mapState } from 'vuex'
+import { ErrorContact } from '@/components/common'
+
+@Component({
+  computed: { ...mapState(['entityName']) },
+  components: { ErrorContact }
+})
+export default class FutureEffectiveFiling extends Vue {
```
cameron-freshworks, line 25: Very cool, is this little component what ties in with a Future effective filing in the Filing History List?

src/components/Dashboard/PaperFiling.vue
```diff
@@ -4,7 +4,8 @@
       as paper documents.</p>
     <p>To request copies of paper documents, contact BC Registry Staff with the document you require and
       the name and incorporation number of your association:</p>
-    <ErrorContact class="mt-5" />
+
+    <ErrorContact class="pt-3" />
```
cameron-freshworks, line 8: I finally adopted this approach of using helper classes for padding and margins after seeing you use them, so useful! Thanks for using them consistently 👍

src/components/Dashboard/TodoList.vue
```diff
@@ -85,10 +90,16 @@
...
-                  <v-btn x-small icon class="expand-btn">
-                    <v-icon>mdi-message-reply</v-icon>
+                  <v-btn
+                    v-if="task.comments.length > 0"
+                    class="expand-btn"
+                    outlined
+                    color="primary"
+                    :ripple=false
+                  >
+                    <v-icon small left style="padding-top: 2px">mdi-message-reply</v-icon>
```
cameron-freshworks, line 100: would a helper class work here? I'm not sure if class="pt-2" would work the same on a v-icon tho.. 🤔

src/App.vue
severinbeauvais: At the moment, a Name Request does not have a draft (hence "Incorporate Now" vs "Resume"). However, this will shortly change when using a NR creates a draft that we will fetch using Temp Reg Number... this code will need a bit of refactoring and there's a ticket for that.

src/components/Dashboard/AddressListSm.vue
severinbeauvais: BCOMP FE COA filings _should_ work. They are mutually exclusive with IA filings. That said, I haven't tested them explicitly. As I refactored much of the layout, it may be possible to improve their display as well (ie, button and text alignment).

src/components/Dashboard/FutureEffectiveFiling.vue
severinbeauvais: Yes. ATM it's specific to IAs. It really cleaned up the Filings History List code:
```
          <template v-else-if="filing.isIaFutureEffective">
            <future-effective-filing :filing=filing />
            <v-divider class="mt-7 mb-5"></v-divider>
          </template>
```

src/components/Dashboard/PaperFiling.vue
severinbeauvais: Cool. My approach is:
- I think it usually makes sense to use spacing/layout helper classes in the template, where you can see how the different components are laid out (instead of hiding these styles in a CSS block).
- However, elements that need common (or specific) styling should use specific class names.

In general I think it's good to try to keep "layout" separate from "styling".

src/components/Dashboard/TodoList.vue
severinbeauvais: There isn't a Vuetify class that gives 2px padding/margin.

In this case it's very specific to this icon to push it down a bit for best display in this button so I think this inline style is acceptable.

src/views/AnnualReport.vue
```diff
@@ -30,7 +30,7 @@
     />
 
     <bcol-error-dialog
-      :bcolOBject="bcolObj"
+      :bcolObject="bcolObj"
```
peter-freshworks, line 33: woah, how did this work before with the case being wrong :s Good catch!

severinbeauvais: Apparently, props aren't case sensitive. See https://www.google.com/search?q=vue+prop+name+casing .

There's also some magic with event names ... something like ImAnEvent gets transformed to im-an-event ... not sure. I just follow our existing code style, which works :)

severinbeauvais:
> The checks also indicate a duplicate class in the css for the ToDo list. Class ".todo-list-detail" or something like that. Not sure what can be done about the ::v-deep though, that is throwing the bugs...

I updated the CSS to solve the first SonarCloud warning.

src/components/Dashboard/AddressListSm.vue
cameron-freshworks: Excellent, will give it a quick revisit in the follow up ticket, sounds promising though :)

src/components/Dashboard/FutureEffectiveFiling.vue
cameron-freshworks: Awesome!

src/components/Dashboard/PaperFiling.vue
cameron-freshworks: Great points and a good rule of thumb to follow 👍

PR #78, opened 2020-05-20
fixed BCOMP COA issues, etc

severinbeauvais:
*Issue #:* /bcgov/entity#2759

*Description of changes:*
- first commit: fixed latent NR-related unit test warnings
- second commit: fixed latent BCOL-related unit test warnings
- third commit:
  - fixed BCOMP COA pending filings check
  - fixed pending chip style
  - fixed latent lint errors
- fourth commit: updated app version

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -438,8 +438,8 @@ export default {
         // is this a Future Effective Incorp App?
         const isIaFutureEffective = this.isFutureEffective(filing)
 
-        // if it's not a FE IA, is it paid?
-        const isPaid = !isIaFutureEffective && (header.status === FilingStatus.PAID)
+        // is this a paid filing?
+        const isPaid = (header.status === FilingStatus.PAID)
 
         const name = this.filingTypeToName(filingType)
         let docFilename: string
@@ -514,8 +514,8 @@ export default {
           filingType === FilingTypes.CHANGE_OF_ADDRESS &&
           this.isFutureEffective(filing)
 
-        // if it's not a BCOMP COA FE, is it paid?
-        const isPaid = !isBcompCoaFutureEffective && (header.status === FilingStatus.PAID)
+        // is this a paid filing?
+        const isPaid = (header.status === FilingStatus.PAID)
 
         // build filing item
         const item: any = {
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -61,7 +61,7 @@
                       </v-chip>
                     </template>
                     <span>The updated office addresses will be legally effective on {{ coaEffectiveDate }},
-                      12:01 AM(Pacific Time). No other filings are allowed until then.</span>
+                      12:01 AM (Pacific Time). No other filings are allowed until then.</span>
                   </v-tooltip>
                 </v-scale-transition>
                 <v-btn text small color="primary"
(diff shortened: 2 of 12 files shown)


severinbeauvais:
Looks like this now:

BCOMP COA FE

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -438,8 +438,8 @@ export default {
         // is this a Future Effective Incorp App?
         const isIaFutureEffective = this.isFutureEffective(filing)
 
-        // if it's not a FE IA, is it paid?
-        const isPaid = !isIaFutureEffective && (header.status === FilingStatus.PAID)
+        // is this a paid filing?
+        const isPaid = (header.status === FilingStatus.PAID)
```
severinbeauvais, line 442: There was no reason to have isPaid mutually exclusive with isIaFutureEffective. Really, in some cases, they can both be true. The only caveat is that, depending on what needs to display, the template should first check if isIaFutureEffective, else check if isPaid... the template already does that so we're good!

src/views/Dashboard.vue
```diff
@@ -263,24 +263,24 @@ export default {
...
+    checkPendingFilings (filings: Array<any>): void {
+      if (!filings?.length) return // safety check
+
+      const foundPaid = filings.some(filing => {
+        if (filing.isPaid) {
+          if (filing.isBcompCoaFutureEffective) {
+            this.coaPending = true
+            this.coaEffectiveDate = filing.effectiveDate
+          }
           this.hasBlockerFiling = true
```
severinbeauvais, line 279: Now, if any filing is status Paid, we consider that a blocker filing. (Previously checked for only BCOMP and only COA.)

src/views/Dashboard.vue
```diff
@@ -336,12 +334,6 @@ section header {
   }
 }
 
-.pending-chip {
-  max-width: 3.75rem;
-  height: 1rem;
-  overflow: visible;
-  font-size: 8px;
-}
```
severinbeauvais, line 344: It turns out all of these (except overflow) were overridden by other styles. And overflow did nothing in this case.

tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -1466,12 +1464,13 @@ describe('Change of Directors - BCOL error dialog on save', () => {
         ConfirmDialog: true,
         PaymentErrorDialog: true,
         ResumeErrorDialog: true,
+        BcolErrorDialog: true,
```
severinbeauvais, line 1467: As this dialog wasn't stubbed, the tests complained of being unable to find the component's ID tag to attach to.

tests/unit/TodoList.spec.ts
```diff
@@ -1854,7 +1856,30 @@ describe('TodoList - Click Tests - NRs and Incorp Apps', () => {
...
+                stateProvinceCd: 'BC',
+                emailAddress: 'abc@test.com',
+                phoneNumber: '7777777777',
+                firstName: 'Adam',
+                middleName: 'John',
+                lastName: 'Smith'
+              },
+              consentFlag: null,
+              expirationDate: 'Thu, 31 Dec 2099 08:00:00 GMT',
+              state: 'APPROVED'
```
severinbeauvais, line 1882: So this test called several of the NR helper functions, so they pretty much needed a fully populated NR to work with.

PR #80, opened 2020-05-27
refactor to use Temp Reg Number for IAs

severinbeauvais:
*Issue #:* /bcgov/entity#3499

*Description of changes:*
- changed route param from NR Number to Temporary Registration Number
- refactored fetchData for simpler code
- now fetch and parse draft IA filing
- store task if IA is draft status
- store filing if IA is paid status
- changed console.error() to console.log() for handled errors
- updated EntityInfo for new states
- updated FilingHistoryList for new states
- added FilingHistoryList emitter for simpler dashboard code
- updated TodoList for new states
- commented out TodoList NR handling (for now)
- updated IA delete dialog messages, and added redirects
- deleted obsolete CSS
- updated text in NR Auth Error dialog
- updated text in NR Invalid dialog
- changed nrNum -> nrNumber
- removed unneeded await for button.trigger() calls
- deleted some unneeded code from unit tests
- commented out trigger calls that don't work
- cleared session storage in some tests as needed
- changed misc unit tests from callback to async mechanism
- also: on app start error, display alert before trying to redirect

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -39,7 +39,6 @@
         :key="index"
         :class="{
           'disabled': !task.enabled,
-          'draft': isStatusDraft(task) && !isTypeCorrection(task) && !isBcolError(task),
           'bcol-error': isBcolError(task)
         }"
       >
@@ -47,8 +46,7 @@
           <div class="list-item">
             <div class="todo-label">
               <h3 class="list-item__title">{{task.title}}
-                <v-btn
-                  v-if="isTypeCorrection(task) && isStatusDraft(task)"
+                <v-btn v-if="isTypeCorrection(task) && isStatusDraft(task)"
                   class="expand-btn ml-0"
                   outlined
                   color="red"
@@ -59,8 +57,8 @@
                 </v-btn>
               </h3>
 
-              <div class="bcorps-ar-subtitle"
-                v-if="businessId && isBComp() && isTypeAnnualReport(task) && isStatusNew(task)"
+              <div v-if="businessId && isBComp() && isTypeAnnualReport(task) && isStatusNew(task)"
+                class="bcorps-ar-subtitle"
               >
                 <p>Verify your Office Address and Current Directors before filing your Annual Report.</p>
                 <v-checkbox
@@ -75,7 +73,7 @@
               <div class="list-item__subtitle">
                 <div v-if="task.subtitle" class="todo-subtitle">
                   <div>{{ task.subtitle }}</div>
-                  <div class="payment-status" v-if="isTypeNameRequest(task)">
+                  <div v-if="task.nameRequest" class="payment-status">
                     <v-btn
                       class="expand-btn"
                       outlined
@@ -103,7 +101,7 @@
                   </v-btn>
                 </div>
 
-                 <div v-else-if="isBcolError(task)" class="todo-subtitle">
+                <div v-else-if="isBcolError(task)" class="todo-subtitle">
                   <div>FILING PENDING</div>
                   <div class="vert-pipe"></div>
                   <div class="payment-status">
@@ -121,7 +119,7 @@
                   </div>
                 </div>
 
-                <div v-else-if="isStatusDraft(task)" class="todo-subtitle">
+                <div v-else-if="isStatusDraft(task) && !isTypeIncorporationApplication(task)" class="todo-subtitle">
                   <div>DRAFT</div>
                 </div>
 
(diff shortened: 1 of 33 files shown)


tests/unit/Dashboard.spec.ts
```diff
@@ -60,37 +61,36 @@ describe('Dashboard - UI', () => {
...
 
-  it('enables standalone filing buttons when there are no blocker filings in the to-do list', () => {
-    wrapper.find(TodoList).vm.$emit('has-blocker-filing', false)
+  it('enables standalone filing buttons when there are no blockers', () => {
+    expect(vm.disableChanges).toEqual(false)
+    expect(wrapper.find('#standalone-addresses-button').attributes('disabled')).toBeUndefined()
+    expect(wrapper.find('#standalone-directors-button').attributes('disabled')).toBeUndefined()
+  })
+
+  it('disables standalone filing buttons when there is a blocker filing in the todo list', () => {
```
severinbeauvais, line 70: When there is a blocker TASK...

tests/unit/Dashboard.spec.ts
```diff
@@ -60,37 +61,36 @@ describe('Dashboard - UI', () => {
...
-    expect(wrapper.find('#standalone-directors-button')
-      .attributes('disabled')).toBeUndefined()
+    expect(vm.hasBlockerTask).toEqual(true)
+    expect(wrapper.find('#standalone-addresses-button').attributes('disabled')).toBe('true')
+    expect(wrapper.find('#standalone-directors-button').attributes('disabled')).toBe('true')
   })
 
-  it('disables standalone filing buttons when there is a blocker filing in the to-do list', () => {
-    wrapper.find(TodoList).vm.$emit('has-blocker-filing', true)
+  it('disables filing buttons when there is a future effective filing pending', () => {
```
severinbeauvais, line 78: disables standalone filing buttons when there is a blocker filing in the filing history list

src/components/EntityInfo.vue
```diff
@@ -4,9 +4,13 @@
 
       <!-- Entity Name, Entity Status -->
       <div class="title-container">
-        <div class="mb-1" id="entity-legal-name">
-          <span>{{ entityName || 'Not Available' }}</span>
+        <div v-if="businessId" class="mb-1" id="entity-legal-name" aria-label="Business Legal Name">
```
cameron-freshworks, line 7: Would it make sense to use <template> here? After you had recommended it to me, i found it made a lot of sense when you want an invisible wrapper to multiple conditional elements. Though, in this case you have a single element and with styling, so it's probably good as it is :) 
Just thinking out loud here!

src/mixins/namex-request-mixin.ts
```diff
@@ -62,15 +63,16 @@ export default class NamexRequestMixin extends Mixins(DateMixin) {
...
    * @param filingId the filing id
    */
-  generateNameRequestState (nr: any, filingId: number): NameRequestIF {
+  parseNameRequest (nr: any, filingId: number): NameRequestIF {
     const approvedName = nr.names.find(name => name.state === NameRequestStates.APPROVED).name
     return {
-      nrNumber: nr.nrNum,
-      // TODO: Update entityType to use nr.requestTypeCd when namex supports our entity types
+      // workaround for old or new property name
+      nrNumber: nr.nrNum || nr.nrNumber,
```
cameron-freshworks, line 74: Sorry, whats the difference between nrNum & nrNumber?

src/components/EntityInfo.vue
severinbeauvais: I think there would be little to gain here by using a template element to wrap the v-ifs -- I would still need the divs and spans for styling. But I did fix a few places where there were divs just for the same of a wrapper element.

src/mixins/namex-request-mixin.ts
severinbeauvais: "ber"

nrNum is what the old property is called in the namex NR. nrNumber is what we've been calling it everywhere else in auth-web, filings UI, create UI and Legal API. So I added this code to support both the old name and the new name. If/when namex changes to nrNumber, then we can remove our workaround (it's in a few places, always with the same comment to make it easier to find).

src/components/Dashboard/TodoList.vue
```diff
@@ -705,20 +714,33 @@ export default {
...
+          title,
+          subtitle,
           draftTitle: 'Incorporation Application',
           status: filing.header.status,
           enabled: Boolean(task.enabled),
           order: task.order,
           paymentToken: filing.header.paymentToken || null,
-          bcolErrObj: bcolObj
+          bcolErrObj: bcolObj,
+          isEmptyFiling: !haveData,
```
severinbeauvais, line 742: I need to look into this... doesn't appear to be working.

Please don't merge yet. Thx.

severinbeauvais: Fixed in latest commit!

src/components/dialogs/NameRequestAuthErrorDialog.vue
```diff
@@ -1,19 +1,19 @@
...
+        <p class="genErr">Your account is currently unable to access this Incorporation Application.
+          This may be because of the following:</p>
         <ul>
-          <li>Your account is not authorized to access this Name Request &mdash; contact the Name
-            Request owner to get access.</li>
+          <li>Your account is not authorized to access this Incorporation Application &mdash; contact
+            the account owner to get access.</li>
           <li>Your login session has timed out &mdash; please exit and then login again.</li>
-          <li>The specified Name Request Number is not valid.</li>
+          <li>The specified Incorporation Application number is not valid.</li>
```
peter-freshworks, line 13: Terminology question - incorporation application number? Is that a thing?

severinbeauvais: This is what, internally, we've calling the Temporary Registration Number. I termed it "IA number" to make sense to the user. Scott seemed OK with it. Open to ideas though.

PR #84, opened 2020-06-02
3597 - update to use documents list from legal api

ochiu:
*Issue #:* /bcgov/entity#3597

*Description of changes:*
- update report functions to use documents list from legal-api

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -298,8 +298,8 @@ export default {
 
   created (): void {
     // constants
-    this.DOCUMENT_TYPE_REPORT = 1
-    this.DOCUMENT_TYPE_RECEIPT = 2
+    this.DOCUMENT_TYPE_REPORT = 'REPORT'
+    this.DOCUMENT_TYPE_RECEIPT = 'RECEIPT'
 
     // load data into this page
     this.loadData()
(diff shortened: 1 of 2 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -376,6 +376,7 @@ export default {
     loadAnnualReport (filing: any) {
       const header = filing?.header
       const annualReport = filing?.annualReport
+      const documents = filing?.documents
 
       if (header && annualReport) {
```
severinbeauvais, line 381: Should you ensure 'documents' is truthy here?

Also in other methods below.

ochiu: I do ```documents: documents || [] as Array<any>``` when assigning it to the item object, is that okay?

ochiu:
> Well that was easy 😁

Thanks to all your hard work! It made this bit easy :)

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: You do that so the template for loop doesn't fail if there are no documents.

I'm actually wondering if 'documents' are required in all filings in the history list? I suspect they are (except for paper only).

ochiu: Currently the api will return an empty array if the logic determines there are no relevant documents to return.

severinbeauvais: Cool. So I suggest doing the same thing as with 'comments':
- don't bother ensuring it's truthy on line 380
- move the assignment on line 379 to line 398 (the filings obj) and keep your fallback safety check ([])
- do the same for the other filing types

ochiu: Updated. Thanks!

PR #85, opened 2020-06-05
handle consent received and consent waived states

severinbeauvais:
*Issue #:* /bcgov/entity#3945

*Description of changes:*
- allow conditional NR if consent is received or waived
- if approved, get name whose state is "APPROVED"
- if conditional, get name whose state is "CONDITION"
- added case for waived consent flag
- added case for null consent flag
- added unit tests
- updated sbc-common-components dependency to 2.1.32


severinbeauvais:
@cameron-freshworks At the moment we only care about "consumable" NRs, which much be in state APPROVED, or CONDITIONAL with special Consent Flag values:
null = consent not required
R = consent received
N = consent waived

PR #89, opened 2020-06-09
3955 - handling temporary IA complete state

ochiu:
*Issue #:* /bcgov/entity#3955

*Description of changes:*
- update initialization check to allow COMPLETED state
- changed EntityStatus.PAID_INCORP_APP to EntityStatus.FILED_INCORP_APP
- added component for complete filing (used only for completed IA with temp reg num)
- removed has-blocker-filing event from FilingHistoryList as an IA filing will show up when complete and using the real business identifier which will disable changes(we don't want this). This check has been changed to always disable changes if it we are using a temp reg num (check is in dashboard component) and any pending states in the history filings are already checked in the dashboard component as well.
- removed hasBlockerFiling code from Dashboard component
- update pending filing with return to dashboard button
- Complete Filing unit tests
- update dashboard unit test (temp reg num should disable changes)
- update pending filing test to check for return to dashboard button
- update entity info unit test to use updated entity status (FILED_INCORP_APP)

### DIFF
--- a/src/components/Dashboard/CompleteFiling.vue
+++ b/src/components/Dashboard/CompleteFiling.vue
@@ -22,22 +22,17 @@ import { Component, Vue } from 'vue-property-decorator'
 
 @Component({
   computed: {
-    ...mapState(['entityName', 'entityIncNo'])
+    ...mapState(['entityName'])
   }
 })
 export default class CompleteFiling extends Vue {
   readonly entityName!: string
-  readonly entityIncNo!: string
 
   /** Determine the name for the entity.
    * Use the entity name if it is available, otherwise build name for numbered company.
    */
   private get name (): string {
-    // Safety check
-    if (!this.entityName && !this.entityIncNo) {
-      return ''
-    }
-    return this.entityName || `${this.entityIncNo.replace('BC', '')} B.C. LTD.`
+    return this.entityName || 'A numbered benefit company'
   }
 
   private returnToDashboard (): void {
--- a/src/components/Dashboard/CompleteFiling.vue
+++ b/src/components/Dashboard/CompleteFiling.vue
@@ -1,12 +1,10 @@
 <template>
-  <div class="pending-filing body-2">
-    <h4>Filing Pending</h4>
+  <div class="complete-ia-filing body-2">
+    <h4>Incorporation Complete</h4>
 
-    <p>This {{title}} is paid, but the filing has not been completed by the BC
-      Registry yet. Some filings may take longer than expected.</p>
-    <p>If this issue persists, please contact us.</p>
+    <p>{{name}} has been successfully incorporated.</p>
+    <p>Return to your Manage Businesses dashboard to access your business and file changes.</p>
 
-    <ErrorContact />
     <div class="to-dashboard-container">
       <v-btn
         color="primary"
@@ -19,19 +17,25 @@
 </template>
 
 <script lang="ts">
+import { mapState } from 'vuex'
 import { Component, Vue, Prop } from 'vue-property-decorator'
 import { ErrorContact } from '@/components/common'
 
 @Component({
-  components: { ErrorContact }
+  components: { ErrorContact },
+  computed: {
+    ...mapState(['entityName', 'entityIncNo'])
+  }
 })
-export default class PendingFiling extends Vue {
-  /** The subject filing. */
-  @Prop() private filing: any
+export default class CompleteFiling extends Vue {
+  readonly entityName!: string
+  readonly entityIncNo!: string
 
-  /** The tite of the subject filing. */
-  private get title (): string {
-    return this.filing?.title || 'filing'
+  /** Determine the name for the entity.
+   * Use the entity name if it is available, otherwise build name for numbered company.
+   */
+  private get name (): string {
+    return this.entityName || `${this.entityIncNo.replace('BC', '')} B.C. LTD.`
   }
 
   private returnToDashboard (): void {
@@ -62,7 +66,4 @@ p {
   margin-bottom: 0.5rem !important;
 }
 
-.error-contact {
-  padding-top: 0.75rem;
-}
 </style>
--- a/src/components/Dashboard/CompleteFiling.vue
+++ b/src/components/Dashboard/CompleteFiling.vue
@@ -9,7 +9,7 @@
       <v-btn
         color="primary"
         @click.native.stop="returnToDashboard()"
-        >
+      >
         <span>Return to Dashboard</span>
       </v-btn>
     </div>
--- a/src/components/Dashboard/CompleteFiling.vue
+++ b/src/components/Dashboard/CompleteFiling.vue
@@ -18,11 +18,9 @@
 
 <script lang="ts">
 import { mapState } from 'vuex'
-import { Component, Vue, Prop } from 'vue-property-decorator'
-import { ErrorContact } from '@/components/common'
+import { Component, Vue } from 'vue-property-decorator'
 
 @Component({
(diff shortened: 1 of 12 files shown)


src/components/Dashboard/PendingFiling.vue
```diff
@@ -7,6 +7,14 @@
     <p>If this issue persists, please contact us.</p>
 
     <ErrorContact />
+    <div class="to-dashboard-container">
```
ochiu, line 10: Added a return to dashboard button for now, waiting for mockup updates

src/components/Dashboard/CompleteFiling.vue
```diff
@@ -0,0 +1,69 @@
...
+        <span>Return to Dashboard</span>
+      </v-btn>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { mapState } from 'vuex'
+import { Component, Vue, Prop } from 'vue-property-decorator'
+import { ErrorContact } from '@/components/common'
```
severinbeauvais, line 22: Obsolete?

src/components/Dashboard/CompleteFiling.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <div class="complete-ia-filing body-2">
+    <h4>Incorporation Complete</h4>
+
+    <p>{{name}} has been successfully incorporated.</p>
+    <p>Return to your Manage Businesses dashboard to access your business and file changes.</p>
```
severinbeauvais, line 6: Mismatch.
The text says Manage Businesses dashboard but the URL is the Business Registry home page.

If you really meant the Manage Businesses dashboard then look at https://github.com/bcgov/business-filings-ui/blob/01b3c1d1eeecc23e5b40d9fc82478d68dd8aea7b/src/components/Dashboard/TodoList.vue#L881 .

src/components/Dashboard/CompleteFiling.vue
```diff
@@ -0,0 +1,69 @@
...
+})
+export default class CompleteFiling extends Vue {
+  readonly entityName!: string
+  readonly entityIncNo!: string
+
+  /** Determine the name for the entity.
+   * Use the entity name if it is available, otherwise build name for numbered company.
+   */
+  private get name (): string {
+    return this.entityName || `${this.entityIncNo.replace('BC', '')} B.C. LTD.`
```
severinbeauvais, line 38: What's an example of entityIncNo in this scenario?
Also, you should check that entityIncNo isn't null.

ochiu: I checked a numbered company completed filing, and it comes in with entityIncNo = 'BCxxxx' and the entityName is null.

I will add that check.

src/components/Dashboard/CompleteFiling.vue
ochiu: ahh good catch, my brain glossed over the 'business' part

src/components/Dashboard/CompleteFiling.vue
ochiu: Good catch! and I thought I managed to clean this up. Will update.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -156,6 +170,11 @@
             <v-divider class="mt-7 mb-5"></v-divider>
           </template>
 
+          <template v-else-if="filing.isCompleteIA && tempRegNumber">
```
severinbeauvais, line 173: Why do you need both conditions? Isn't isCompleteIA enough?

ochiu: Based on what I see on the api, when we have a real business number the complete IA should return as part of the history, but in that case we don't need to show this piece.

src/components/Dashboard/CompleteFiling.vue
severinbeauvais: Is that really the name we display? I thought this referred to number company names?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Ah, you are correct!

src/components/Dashboard/CompleteFiling.vue
ochiu: Yes, that is for the numbered company name, otherwise the entity name should be used. When I loaded a numbered company the entityName came back null so I added this piece in so it displays properly for a numbered company.

severinbeauvais: For a numbered company, entityName is not set.

I think the BCxxx number is not the "company number" we're talking about. This is the business number.

The name for a numbered company is the incorp # (assigned on completion in COLIN).

We should discuss this further with Scott.

src/components/Dashboard/PendingFiling.vue
ochiu: Mock ups were updated to just add a `Return to Dashboard` button, which matches the implementation.

src/components/Dashboard/CompleteFiling.vue
ochiu: Discussed with Scott, going with "A numbered benefit company"

src/components/Dashboard/CompleteFiling.vue
```diff
@@ -0,0 +1,66 @@
...
+    <h4>Incorporation Complete</h4>
+
+    <p>{{name}} has been successfully incorporated.</p>
+    <p>Return to your Manage Businesses dashboard to access your business and file changes.</p>
+
+    <div class="to-dashboard-container">
+      <v-btn
+        color="primary"
+        @click.native.stop="returnToDashboard()"
+        >
```
cameron-freshworks, line 12: Think the end of this tag has an extra tab space

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -444,6 +457,9 @@ export default {
         // is this a paid filing?
         const isPaid = (header.status === FilingStatus.PAID)
 
+        // is this a completed filing?
+        const isCompleteIA = (header.status === FilingStatus.COMPLETED)
```
cameron-freshworks, line 461: parentheses required? Maybe a preference for readability? Either way, no action required, just curious :)

src/views/Dashboard.vue
```diff
@@ -171,14 +169,24 @@ export default {
...
+      return (this.entityStatus === EntityStatus.FILED_INCORP_APP)
     },
 
-    /** Whether to block a new filing because another item has to be finished first. */
+    /** Whether to block a new filing because another item has to be finished first.
+     * No changes are allowed if
+     * 1) this is a temporary reg number until it switches to a real business number
+     * 2) has a pending filing (is PAID) and waiting for completion
+     * 3) has a blocker task in the todo list (ie. draft, paid, error, correction )
+     */
```
cameron-freshworks, line 182: Great commments!

src/components/Dashboard/FilingHistoryList.vue
ochiu: Just followed isPaid, but I kind of like it? haha

severinbeauvais: (FWIW - I prefer them - easier to see what this statement is doing.)

src/components/Dashboard/CompleteFiling.vue
ochiu: fixed

severinbeauvais: I see this is now fixed.

PR #91, opened 2020-06-11
#3927 Filings UI BreadCrumb

cameron-freshworks:
*Issue #:* /bcgov/entity#3927

*Description of changes:*

* Implemented a breadcrumb in the filings ui to help navigate between filings and also back to the manage business dashboard.

* Updated unit testing.

* Updated Babel: Unit test suites weren't running in the GHA pipeline

*NOTES:*

* Still need to implement a breadcrumb in create-ui to allow navigation back to the filings UI Dash or Manage Business Dash.

* If you're running local, you won't have access to `CURRENT_ACCOUNT` from session storage and won't be able to return directly to the correct manage business dashboard. This data however is available in our deployed environments. FYI.


package.json
```diff
@@ -39,7 +39,7 @@
     "@types/vuelidate": "^0.7.11",
     "@typescript-eslint/eslint-plugin": "^2.23.0",
     "@typescript-eslint/parser": "^2.23.0",
-    "@vue/cli-plugin-babel": "4.2.3",
+    "@vue/cli-plugin-babel": "4.4.1",
```
severinbeauvais, line 42: Please call out this change in the PR notes.

Also please update this project's version (line 3 above).

cameron-freshworks: I noted it in change description

kk, will do 👍 .

src/interfaces/breadcrumb-interface.ts
```diff
@@ -0,0 +1,9 @@
+import Vue from 'vue'
+
+export interface BreadcrumbInterface extends Vue {
```
severinbeauvais, line 3: It is Breadcrumb or BreadCrumb? Code uses both :)

cameron-freshworks: I believe it's one word. Let me search out those camelcase instances.

src/routes.ts
```diff
@@ -25,31 +25,63 @@ export default [
     name: ANNUAL_REPORT,
     component: AnnualReport,
     meta: {
-      requiresAuth: true
+      requiresAuth: true,
+      breadcrumb: [
+        {
+          text: 'File Annual Report',
```
severinbeauvais, line 31: The mockup shows "File 2018 Annual Report" but I think design won't allow that. Check with Scott or a PO for an exemption :)

tests/unit/EntityInfo.spec.ts
```diff
@@ -131,4 +134,84 @@ describe('EntityInfo', () => {
...
+    sessionStorage.clear()
+    sessionStorage.setItem('TEMP_REG_NUMBER', 'T123456789')
+
+    store.state.entityName = 'My Named Company'
+    const wrapper = mount(EntityInfo, { vuetify, store, router })
+
+    await Vue.nextTick()
+
+    expect(wrapper.find('.breadcrumb').exists()).toBe(true)
+    expect(wrapper.find('.breadcrumb').text()).toContain('My Account')
```
severinbeauvais, line 151: I thought we were going with "Manage Businesses" for the first level.

Also, if would be great if you could check each level individually instead of checking that the who thing "contains" the separate crumbs.

src/routes.ts
cameron-freshworks: Will do!

tests/unit/EntityInfo.spec.ts
cameron-freshworks: 'My account' does go to Manage businesses. The dashboard name is the business name.
You will always have the My Account and Dashboard level. It only goes one level deeper, and those are all checked individually.

src/components/EntityInfo.vue
```diff
@@ -180,6 +195,29 @@ export default class EntityInfo extends Mixins(EnumMixin) {
     // assume Business Profile URL is always reachable
     window.location.assign(businessProfileUrl)
   }
+
+  /** Track route for breadcrumb. */
+  @Watch('$route', { immediate: true })
+  private getBreadCrumbs (): Array<BreadcrumbInterface> {
```
severinbeauvais: Interesting. So this is both an event callback and a getter?

Wouldn't a getter suffice?

src/components/EntityInfo.vue
```diff
@@ -180,6 +195,29 @@ export default class EntityInfo extends Mixins(EnumMixin) {
...
+    const breadCrumbs = this.$route?.meta?.breadcrumb
+    // The Current Account data only available in deployed environments
+    const currentAccount = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))
+
+    return [
+      {
+        text: 'My Account',
+        disabled: false,
+        href: `${sessionStorage.getItem('AUTH_URL')}account/${currentAccount?.id}/business`
```
severinbeauvais, line 210: There's actually a short-cut URL to this page, eg.
https://dev.bcregistry.ca/cooperatives/auth/business

^^ auth-web auto-forwards to the correct account id based on whatever's in the session storage (I think)

src/components/EntityInfo.vue
cameron-freshworks: Not sure, will have to look into this!

package.json
severinbeauvais: Done, thanks!

src/components/EntityInfo.vue
cameron-freshworks: Hmm, ok fair enough. I wasn't sure as locally i didn't have anything for Current Account, wasn't able to test that. I'll try this out.

tests/unit/EntityInfo.spec.ts
severinbeauvais: See comment in epic ticket
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/3740

src/components/EntityInfo.vue
severinbeauvais: Yeah but you already assume you have current account in session storage -- isn't that the data auth-web needs for the shortcut?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "2.11.0",
+  "version": "2.12.0",
```
severinbeauvais, line 3: This didn't get updated in the package-lock.json. Dunno if you need to do `npm i` or something to update it.
I think we forgot to update the lock file previously and the world hasn't ended... wait, COVID, riots, umm...

tests/unit/EntityInfo.spec.ts
cameron-freshworks: Very Confusing.. 

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/3763

This ticket says see the wireframe above which shows my account.. ?

package.json
cameron-freshworks: We found the root cause of 2020's problems! 

Will look into this 👍

tests/unit/EntityInfo.spec.ts
cameron-freshworks: https://preview.uxpin.com/b3e60baaf1f0418fec3fac72fcc7bdce6f772e13#/pages/128469428/specification/no-panels

src/components/EntityInfo.vue
cameron-freshworks: Perhaps, i think so. But i couldn't guarantee it goes to the specific manage business dashboard testing from local, so i tried to be explicit. I can remove it and we'll assume it does what we're hoping?

cameron-freshworks: Edit* It works!

src/components/EntityInfo.vue
cameron-freshworks: A getter will suffice 👍

tests/unit/EntityInfo.spec.ts
severinbeauvais: Maybe ask Scott for the latest decision on this?

src/components/EntityInfo.vue
severinbeauvais: Does it work if you start from the Create UI (instead of starting from auth-web)? Wondering if the necessary data is set in this scenario, and if not whether it handles it anyway.

tests/unit/EntityInfo.spec.ts
cameron-freshworks: Way ahead of ya :)

src/components/EntityInfo.vue
cameron-freshworks: Sorry, i should have clarified. You were correct, it handles it already by using the short cut URL.

package.json
cameron-freshworks: Updated in next commit!

src/routes.ts
cameron-freshworks: I'd like to make this work actually, working on it!

src/components/EntityInfo.vue
```diff
@@ -180,6 +197,30 @@ export default class EntityInfo extends Mixins(EnumMixin) {
     // assume Business Profile URL is always reachable
     window.location.assign(businessProfileUrl)
   }
+
+  /** Get route breadcrumbs. */
+  private getBreadcrumbs (): Array<BreadcrumbInterface> {
```
severinbeauvais, line 202: This doesn't have to be a getter? (private get xxx(): yyy)
With the current code, does the breadcrumb update when the route changes?
Theory I'm going on here is that getters are reactive -- their value changes when the properties they depend on change -- and there's an event for whoever is using them (v-breadcrumbs).
But, if it works as-is then it works. I dunno.

src/components/EntityInfo.vue
```diff
@@ -180,6 +197,30 @@ export default class EntityInfo extends Mixins(EnumMixin) {
...
+    const breadcrumbs = this.$route?.meta?.breadcrumb
+
+    // Apply the filing year to the breadcrumb trail for Annual Reports
+    let ArCrumb = breadcrumbs?.find(item => item.to.name === ANNUAL_REPORT)
```
severinbeauvais, line 206: This could be a const since you never change its value.

severinbeauvais: Nice solution BTW.

src/components/EntityInfo.vue
```diff
@@ -192,6 +233,24 @@ export default class EntityInfo extends Mixins(EnumMixin) {
...
+
+  .breadcrumb {
+    padding: 0;
+  }
+
+  .v-breadcrumbs li {
+    font-size: .75rem;
+  }
+
+  ::v-deep {
```
severinbeauvais, line 245: Yessss!

cameron-freshworks: The sonarCube doesn't like this tho.. :( It's a bug as far as it's concerned. Any ideas?

src/routes.ts
severinbeauvais: Fixed!

src/components/EntityInfo.vue
cameron-freshworks: Thanks, will update this.

tests/unit/EntityInfo.spec.ts
severinbeauvais: Updated!

tests/unit/EntityInfo.spec.ts
```diff
@@ -131,4 +134,104 @@ describe('EntityInfo', () => {
...
+    sessionStorage.clear()
+    sessionStorage.setItem('TEMP_REG_NUMBER', 'T123456789')
+
+    store.state.entityName = 'My Named Company'
+    store.state.ARFilingYear = '2020'
+    const wrapper = mount(EntityInfo, { vuetify, store, router })
+
+    await Vue.nextTick()
+
+    const breadcrumbs = wrapper.vm.$el.querySelectorAll('.v-breadcrumbs li')
```
severinbeauvais, line 176: BTW, the other way to do this is with `wrapper.findAll()` and then you use `at()` to get each found item. No need to change atm.

src/components/EntityInfo.vue
severinbeauvais: I've seen this previously. I don't know how to disable this check so just ignore it for now. It won't report this again in future PRs.

src/components/EntityInfo.vue
cameron-freshworks: It works as is yea, it updates when the route changes. Though i think it might not need to be a computed value. 
This is me speculating, but when we change routes it is re-rendering the virtual Dom because of the changes. 
We have many instances of these functions providing dynamic values to the virtual DOM that aren't computed values, which work (i think) because Vue's reactivity system.
However, i can update this, it's no big deal!

tests/unit/EntityInfo.spec.ts
cameron-freshworks: Interesting, thanks!

src/components/EntityInfo.vue
severinbeauvais: SonarCloud should be configurable... can you spend a small amount of time finding out how to configure it? Maybe Patrick knows...

src/components/EntityInfo.vue
severinbeauvais: If it ain't broke...

src/components/EntityInfo.vue
cameron-freshworks: Great idea, when i finish this ticket i will 100% look into that 👍

src/components/EntityInfo.vue
cameron-freshworks: I 100% agree with you tho, i updated to a getter. It should be computed. Like all things, it made me dig a little deeper into the process and think on it!
Great convo, thanks!

src/components/EntityInfo.vue
```diff
@@ -180,6 +197,30 @@ export default class EntityInfo extends Mixins(EnumMixin) {
     // assume Business Profile URL is always reachable
     window.location.assign(businessProfileUrl)
   }
+
+  /** Get route breadcrumbs. */
+  private get breadcrumbs (): Array<BreadcrumbInterface> {
```
severinbeauvais, line 202: Thanks for the change. This "feels" better, even if it worked before.

cameron-freshworks:
Resolved Conflict and has 3 approvals. Will merge 👍

PR #93, opened 2020-06-12
Bug fixes + Code to hide legal obligation for temp reg number 

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4026, /bcgov/entity#4033

*Description of changes:*
Bug fixes 
Hide Legal Obligation for a temp reg no with a completed incorporation filing

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -74,7 +74,7 @@
                 </div>
 
                 <div v-else-if="filing.isCompleteIA && tempRegNumber" class="filing-subtitle">
-                  <span>PENDING FILING (filed by {{filing.filingAuthor}} on {{filing.filingDate}})</span>
+                  <span>FILED AND PAID (filed by {{filing.filingAuthor}} on {{filing.filingDate}})</span>
                   <v-btn
                     class="details-btn"
                     outlined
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -697,10 +698,12 @@ export default {
       }
     },
 
-    async downloadDocument (document: any) {
+    async downloadDocument (document: any, index: number) {
       this.loadingDocument = true
+      this.downloadingDocIndex = index
       await this.downloadOneDocument(document)
```
ochiu, line 704: Should we also reset the index and loadingDocument indicator within the download error block as well?

lekshmimallika-aot: Those are getting reset after await this.downloadOneDocument(document) and there is not much processing in the error block other than setting the error dialog flag to true. Do we need that?

ochiu: I think the only reason would be if someone were to close the dialog to try again it would be the loading indicator instead of the file name.

ochiu: Ahh I get it now. Nvm it resets as it just passes through that code.

src/components/Dashboard/LegalObligation.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-  <div v-if="showLegalObligation && isBusinessWithNoMaintenanceFilings" class="legal-obligation-container">
+  <div v-if="showLegalObligation && isBusinessWithNoMaintenanceFilings && !tempRegNumber"
```
severinbeauvais, line 2: Good fix. Please mention it in one of the tickets for Riyaz to see. (I sometimes add a "Test Notes" section in the ticket to suggest things to verify.)

lekshmimallika-aot: Sure. Updated ticket 3828

PR #96, opened 2020-06-17
4040 - delete draft IA updates

ochiu:
*Issue #:* /bcgov/entity#4040

*Description of changes:*
- updated delete Incorporation Application logic so the doDeleteDraft() function allows for NOT triggering a dashboard reload as this causes an IA to be re-evaluated briefly showing an error message when the redirect is slow.

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -874,7 +874,9 @@ export default {
       ).then(async (confirm) => {
         // if we get here, "Delete" was clicked
         if (confirm) {
-          await this.doDeleteDraft(task)
+          // delete without refreshing the dashboard as it triggers an error loading an IA
+          // a redirect will happen taking the user off this page
+          await this.doDeleteDraft(task, false)
 
           if (this.nameRequest) {
             // redirect to Manage Businesses page
@@ -895,14 +897,16 @@ export default {
       })
     },
 
-    async doDeleteDraft (task) {
+    async doDeleteDraft (task, refreshDashboard: boolean = true) {
       const id = this.entityIncNo || this.tempRegNumber
       let url = `businesses/${id}/filings/${task.id}`
       await axios.delete(url).then(res => {
         if (!res) { throw new Error('Invalid API response') }
 
-        // emit dashboard reload trigger event
-        this.$root.$emit('triggerDashboardReload')
+        if (refreshDashboard) {
+          // emit dashboard reload trigger event
+          this.$root.$emit('triggerDashboardReload')
+        }
       }).catch(error => {
         if (error?.response) {
           if (error.response.data.errors) {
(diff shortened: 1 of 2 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -874,7 +874,9 @@ export default {
       ).then(async (confirm) => {
         // if we get here, "Delete" was clicked
         if (confirm) {
-          await this.doDeleteDraft(task)
+          // delete without refreshing the dashboard as it triggers an error loading an IA
```
severinbeauvais, line 877: Is there some sort of spinner during the delete action? Or does the dialog just stay up?

ochiu: There doesn't seem to be any spinner and the dialog closes after confirmation of delete.

src/components/Dashboard/TodoList.vue
```diff
@@ -895,14 +897,16 @@ export default {
...
       const id = this.entityIncNo || this.tempRegNumber
       let url = `businesses/${id}/filings/${task.id}`
       await axios.delete(url).then(res => {
         if (!res) { throw new Error('Invalid API response') }
 
-        // emit dashboard reload trigger event
-        this.$root.$emit('triggerDashboardReload')
+        if (refreshDashboard) {
+          // emit dashboard reload trigger event
+          this.$root.$emit('triggerDashboardReload')
```
cameron-freshworks, line 908: Interesting! Is this an alternative to @Emit property decorator?

ochiu: I believe the emit decorator just emits to the parent, while the $root emit will send it to the root. In App.vue we listen on this event.

cameron-freshworks: oh thats pretty cool! Right on, thanks :)

src/components/Dashboard/TodoList.vue
severinbeauvais: I have tested this locally, and while the error dialog no longer displays (because the dashboard is not reloaded), for a second or so I see the old dashboard and it still looks enabled. I think you need to do more here so the user does not get confused by seeing an "active-looking" dashboard.

severinbeauvais: What do you think of leaving that dialog up until the redirection happens? So don't "return" from confirmDeleteIncorporation() ? Or perhaps make the dashboard disabled, or look like it's in loading state?

ochiu: I am also thinking of the options you suggested. I will see what is easiest and simplest to do. thanks!

ochiu: - updated confirm dialog component to allow for the dialog to stay open and added a close() function in the event someone wants to use this and manually close the dialog when some condition is met.

- added a loading indicator when someone hits the yes/confirm button and disabled all the other buttons.

severinbeauvais: Nice.

PR #97, opened 2020-06-18
4072 - future effective filing test update

ochiu:
*Issue #:* /bcgov/entity#4072

*Description of changes:*
- update future effective filing component text for when there is no entity name (Numbered Benefit Company)

### DIFF
--- a/src/components/Dashboard/FutureEffectiveFiling.vue
+++ b/src/components/Dashboard/FutureEffectiveFiling.vue
@@ -2,7 +2,7 @@
   <div class="future-effective-filing body-2">
     <h4>Future Effective Incorporation Date</h4>
 
-    <p>The incorporation date and time for {{entityName || 'unknown'}}
+    <p>The incorporation date and time for {{entityName || 'this Numbered Benefit Company'}}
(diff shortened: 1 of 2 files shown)


src/components/Dashboard/FutureEffectiveFiling.vue
```diff
@@ -2,7 +2,7 @@
   <div class="future-effective-filing body-2">
     <h4>Future Effective Incorporation Date</h4>
 
-    <p>The incorporation date and time for {{entityName || 'unknown'}}
+    <p>The incorporation date and time for {{entityName || 'this Numbered Benefit Company'}}
```
severinbeauvais, line 5: I think this makes the 3rd (or 4th) place that we assume it's a numbered benefit company if there is no entity name. We will need to change this in the near future when we handle more than just benefit companies. When we get to that point, it might be easiest to determine in _one place_ what we should display as entity name and just set the entity name globally there.
Sorry for the rant. I'm not asking to change this one. I'm just raising the fact the WE WILL have to fix this soon-ish :-}

ochiu: Agreed we should set this in one place when we get to that point to reduce the need to update all the different places.

PR #100, opened 2020-06-19
fix 3685 and 4093  - filings UI

peter-freshworks:
*Issue #:* /bcgov/entity#3685
*Issue #:* /bcgov/entity#4093
*Description of changes:*


src/views/AnnualReport.vue
```diff
@@ -870,6 +870,7 @@ export default {
           this.haveChanges = false
         }).catch(async error => {
           if (error && error.response && error.response.status === PAYMENT_REQUIRED) {
+            this.haveChanges = false
```
severinbeauvais, line 873: Please add comment explaining why this special case.
Also in other 2 components.
Also in Correction.vue?

PR #101, opened 2020-06-19
conditionally skip app startup alert

severinbeauvais:
*Issue #:* /bcgov/entity#4126

*Description of changes:*
- display alert only if app error isn't due missing ID
- updated app version to 2.13


severinbeauvais:
> Just curious, are we updating the version on every commit or just in alignment with the next release?

For next release! Create UI needs to be updated as well, I see.

PR #106, opened 2020-06-23
handle IA where FE date as passed but is not yet completed

severinbeauvais:
*Issue #:* /bcgov/entity#4043

*Description of changes:*

Note that these changes are broken up into 2 main commits for easier review:

1. misc cleanup:
  - replaced another instance of Cooperatives Online
  - added check for  empty LD id (to prevent errors locally)
  - renamed Common Mixin to Object Mixin
  - fixed duplicate nrNumber error in App.vue
  - replaced corpDisplayName with an Enum Mixin
  - replaced some mixin methods with getters
2. the feature:
  - renamed some sub-components
  - added FE IA Pending sub-component
  - added header for FE IA Pending
  - added content for FE IA Pending
  - misc property cleanup
  - added calculation for past FE dates
  - added/updated unit tests

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -26,6 +26,7 @@
               <h3 class="list-item__title mb-0">{{filing.title}}{{applyCorrectionTag(filing)}}</h3>
 
               <div class="list-item__subtitle d-flex">
+                <!-- is this a BCOMP FE COA? -->
                 <div v-if="filing.isBcompCoaFutureEffective" class="filing-subtitle">
                   <span>FILED AND PENDING (filed by {{filing.filingAuthor}} on {{filing.filingDate}})</span>
                   <v-tooltip top content-class="pending-tooltip">
@@ -39,10 +40,9 @@
                   </v-tooltip>
                 </div>
 
-                <div v-else-if="filing.isIaFutureEffective" class="filing-subtitle">
-                  <span>FUTURE EFFECTIVE INCORPORATION</span>
-                  <span class="vert-pipe"></span>
-                  <span>PAID (filed by {{filing.filingAuthor}} on {{filing.filingDate}})</span>
+                <!-- is this a COMPLETED IA? (incorp app mode only) -->
+                <div v-else-if="tempRegNumber && filing.isCompleted" class="filing-subtitle">
+                  <span>FILED AND PAID (filed by {{filing.filingAuthor}} on {{filing.filingDate}})</span>
                   <v-btn
                     class="details-btn"
                     outlined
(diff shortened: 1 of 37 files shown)


severinbeauvais:
> Nice clean up! Tested this locally as well with a FED numbered company, looks good.

Thanks for testing this... I really appreciate you looking at this closely.

src/components/EntityInfo.vue
```diff
@@ -213,7 +213,7 @@ export default class EntityInfo extends Mixins(EnumMixin) {
         href: `${sessionStorage.getItem('AUTH_URL')}business`
       },
       {
-        text: this.entityName || this.corpDisplayName,
```
cameron-freshworks, line 216: What was the difference between the entityName and the corpDisplayName?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -450,36 +494,36 @@ export default {
...
+
         let receiptFilename: string
-        if (isIaFutureEffective) {
+        if (isFutureEffectiveIa) {
           receiptFilename = `${this.entityIncNo} - Receipt (Future Effective) - ${filingDate}.pdf`
         } else {
           receiptFilename = `${this.entityIncNo} - Receipt - ${filingDate}.pdf`
         }
 
+        const corpName = this.entityName || this.entityTypeToTitle(this.entityType)
```
cameron-freshworks, line 513: Nice clean up, this essentially a fix to that `Numbered Benefit Company` bug as well.

src/components/EntityInfo.vue
severinbeauvais: Previously, corpDisplayName was a method in CommonMixin that returned entityName (if not falsy), otherwise returned a string like "Numbered Benefit Company".

Now, entityTypeToTitle, which is a method in EnumMixin, just returns the string as before.

So this code is a simplification and use of a general method.

src/mixins/enum-mixin.ts
```diff
@@ -46,6 +46,19 @@ export default class EnumMixin extends Vue {
...
+   * Converts the entity type to an entity title.
+   * @param type the entity type to convert
+   */
+  entityTypeToTitle (type: EntityTypes): string {
+    switch (type) {
+      case EntityTypes.COOP: return 'Numbered Cooperative'
+      case EntityTypes.BCOMP: return 'Numbered Benefit Company'
+      case EntityTypes.CORP: return 'Numbered Corporation'
+    }
+    return 'Unknown' // should never happen
```
cameron-freshworks, line 59: Would it make sense to use a `default: return 'Unknown' in the event none of the cases are matched?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Yup, thanks.

I tried to clean this up so it's easier to understand and extend in the future :)

src/mixins/enum-mixin.ts
severinbeauvais: I think both ways are equivalent.

Interestingly, the latest VSCode extensions show that the 'return' is unreachable code. I believe this is because it knows the values for 'type' and that they're all handled. Here's a screenshot of doing it both ways:

image

cameron-freshworks: Fair enough, i'm good with either or :) Resolved!

PR #110, opened 2020-07-02
#2824 Bcomp AR Bug Fixes

cameron-freshworks:
*Issue #:* /bcgov/entity#2824

*Description of changes:*

* Fixed Bcomp annual report Due Date being the same date for all of the pending Annual Reports.

* Fixed UI bug where checkbox selection was selecting multiple checkboxes in pending Annual Reports.

* Fixed UI bug where grey overlay was triggered on box selection.

* Fixed console error from `tempRegNumber` not being defined on the vue instance.

* Style / UI Tweaks.

* Updated Unit tests.

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -465,7 +465,7 @@ export default {
       deleteErrorDialog: false,
       cancelPaymentErrors: [] as Array<any>,
       cancelPaymentErrorDialog: false,
-      enableCheckbox: [],
+      enableCheckbox: [] as Array<any>,
       confirmEnabled: false,
       currentFilingId: null as number,
       panel: null as number, // currently expanded panel
@@ -1045,9 +1045,9 @@ export default {
 
     /** Determine the Annual Report Due date for a given filing. */
     arDueDate (filingYear: number, foundingDate: Date) {
-      let dueDate = new Date(foundingDate)
+      const dueDate = new Date(foundingDate)
       dueDate.setFullYear(filingYear)
-      dueDate.setDate(dueDate.getDate() + 61)
+      dueDate.setDate(dueDate.getDate() + 60) // The due date is 60 days AFTER the anniversary date for a given year
       return this.toReadableDate(dueDate)
     }
   },
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -465,7 +465,7 @@ export default {
       deleteErrorDialog: false,
       cancelPaymentErrors: [] as Array<any>,
       cancelPaymentErrorDialog: false,
-      enableCheckbox: [] as Array<number>,
+      enableCheckbox: [],
       confirmEnabled: false,
       currentFilingId: null as number,
       panel: null as number, // currently expanded panel
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -67,15 +67,16 @@
                 </v-btn>
               </h3>
 
-              <div v-if="businessId && isBComp && isTypeAnnualReport(task) && isStatusNew(task)"
+              <div v-if="businessId && isBComp && task.enabled && isTypeAnnualReport(task) && isStatusNew(task)"
                 class="bcorps-ar-subtitle"
               >
                 <p>Verify your Office Address and Current Directors before filing your Annual Report.</p>
                 <v-checkbox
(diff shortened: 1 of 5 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -1038,6 +1041,14 @@ export default {
     /** Closes current panel or opens new panel. */
     togglePanel (index: number) {
       this.panel = (this.panel === index) ? null : index
+    },
+
+    /** Determine the Annual Report Due date for a given filing. */
+    arDueDate (filingYear: number, foundingDate: Date) {
+      let dueDate = new Date(foundingDate)
+      dueDate.setFullYear(filingYear)
+      dueDate.setDate(dueDate.getDate() + 61)
```
severinbeauvais, line 1050: Please add comment re: why 61.

src/components/Dashboard/TodoList.vue
```diff
@@ -1038,6 +1041,14 @@ export default {
     /** Closes current panel or opens new panel. */
     togglePanel (index: number) {
       this.panel = (this.panel === index) ? null : index
+    },
+
+    /** Determine the Annual Report Due date for a given filing. */
+    arDueDate (filingYear: number, foundingDate: Date) {
+      let dueDate = new Date(foundingDate)
```
severinbeauvais, line 1048: Can be const, since you don't reassign this property.

src/components/Dashboard/TodoList.vue
```diff
@@ -463,7 +465,7 @@ export default {
       deleteErrorDialog: false,
       cancelPaymentErrors: [] as Array<any>,
       cancelPaymentErrorDialog: false,
-      confirmCheckbox: false,
+      enableCheckbox: [],
```
severinbeauvais, line 468: It feels like this array should be sized (allocated) before it's used in the template but I dunno.

src/views/Dashboard.vue
```diff
@@ -178,7 +178,8 @@ export default {
       inProcessFiling: null as any,
       coaPending: false,
       coaEffectiveDate: null as string,
-      coaWarningDialog: false
+      coaWarningDialog: false,
+      tempRegNumber: sessionStorage.getItem('TEMP_REG_NUMBER')
```
severinbeauvais, line 182: It would be better to fix this like this:
https://github.com/severinbeauvais/business-filings-ui/commit/e1c1e32d269b31ffa81b4a4d15032d1fb15f4154#diff-ac372967af9eeeb576848a16c48d6e24

src/components/Dashboard/TodoList.vue
cameron-freshworks: Absolutely, there is a reason!

src/components/Dashboard/TodoList.vue
cameron-freshworks: good catch, thanks!

src/components/Dashboard/TodoList.vue
cameron-freshworks: i'm not sure exactly what you mean. Couldn't this potentially be any size?

src/views/Dashboard.vue
cameron-freshworks: Will do, thanks!

src/components/Dashboard/TodoList.vue
cameron-freshworks: Maybe i was overthinking this 😆  
The business rule was: "The due date is 60 days AFTER the anniversary date for a given year", so i was thinking the day-of anniversary wasn't counted but it never was. Updated to 60 days and added comment in next commit!

src/components/Dashboard/TodoList.vue
severinbeauvais: ~~I'm not 100% sure how arrays are allocated in JS.~~ I just read up on this. JS arrays are dynamically resizable.

In other languages, you'd have to specify the size (length) before using any arbitrary index (as you are doing in the template), otherwise you'd get a runtime array bounds exception or memory corruption.

Looks like we can get away with this magic after all:
https://stackoverflow.com/a/33903611

I can't quickly find the official docs but I found this:
https://www.educba.com/dynamic-array-in-javascript/

cameron-freshworks:
> Looks OK.
> Since this will be merged soon, please update the app version to 3.1.
> Thx!

Sounds good, updated in next commit!

PR #114, opened 2020-07-07
fixed double subtitles in Todo list

severinbeauvais:
*Issue #:* /bcgov/entity#4145

*Description of changes:*
- updated "contact us" text for consistency
- simplified AR subtitle case
- added IA subtitle case (which is mutually exclusive with other cases)
- deleted some obsolete NR code
- added unit tests

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -31,17 +31,19 @@
       @okay="resetCancelPaymentErrors"
       attach="#todo-list"
     />
+
     <v-card class="no-results" v-if="!tempRegNumber && isBComp && !allowBCompMaintenanceFiling" flat>
       <v-card-text>
         <div class="no-results__title">
           Coming Soon - Online Maintenance Filings such as Address and Director Changes
         </div>
-        <div class="no-results__subtitle">If you need to make changes to your Benefit Company, contact us  </div>
+        <div class="no-results__subtitle">If you need to make changes to your Benefit Company, please contact us.</div>
         <div class="error-contact-container">
           <ErrorContact/>
         </div>
       </v-card-text>
     </v-card>
+
     <v-expansion-panels v-else-if="taskItems && taskItems.length > 0" accordion  v-model="panel">
       <v-expansion-panel
         class="align-items-top todo-item"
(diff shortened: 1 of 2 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -81,8 +83,13 @@
                 />
               </div>
 
-              <div class="list-item__subtitle">
-                <div v-if="task.subtitle && !task.bcolErrObj" class="todo-subtitle">
+              <div v-else class="list-item__subtitle">
```
severinbeauvais, line 86: We have to be careful in this large if/else-if code block that the cases are mutually exclusive. (Bit by bit as I work on the Todo list and History list, I am refactoring to ensure cases are mutually exclusive by cleaning up the logic to make it so.)

I think we haven't yet seen a conflict between the previous case (line 82) and this subtitle block but I think this v-else needs to be here. The unit tests pass, and as long as they verify the difference scenarios then we should be OK. Thoughts?

src/components/Dashboard/TodoList.vue
```diff
@@ -81,8 +83,13 @@
...
+                <div v-if="task.subtitle && isTypeAnnualReport(task)" class="todo-subtitle">
```
severinbeauvais, line 87: According to the script (code), this block is only for AR, so I'm explicitly checking for this condition now.

I added a new block for IA only... on line 92 below.

tests/unit/TodoList.spec.ts
```diff
@@ -1385,6 +1253,205 @@ describe('TodoList - UI - BCOMP', () => {
   })
 })
 
+describe('TodoList - UI - Incorp Apps', () => {
```
severinbeauvais, line 1256: Next time someone is in this unit test file, if possible, they should ensure there is a test for all possibly scenarios. I started things off here with several new tests: 2 to test my fix and 2 more to verify that I didn't break another scenario. (These include "negative checking" -- that stuff that isn't supposed to display -- like duplicate subtitles -- doesn't.)

src/components/Dashboard/TodoList.vue
cameron-freshworks: I think if they are meant to be mutually exclusive then this would be the best approach!

severinbeauvais: I think if you're displaying the BCOMP AR "verify" checkbox then you don't want to show any other subtitle or buttons, right?

tests/unit/TodoList.spec.ts
cameron-freshworks: Sounds good, definitely agree! Though it may be challenging for just one person to know all of the scenarios without a good time investment. 
Definitely agree its useful to incorporate those negative checks alongside verifying the other items!

src/components/Dashboard/TodoList.vue
cameron-freshworks: We have a subtile for the BCOMP AR above the checkbox and we also want to show the button but in it's disabled state until the user "confirms" there information is up to date.

Screen Shot 2020-07-07 at 9 24 14 AM

PR #117, opened 2020-07-27
Filings UI Bug Fix - Watcher to handle Emit

cameron-freshworks:
*Issue #:* /bcgov/entity#4354

*Description of changes:*
* Bux fix where filings UI was updating the parent before the tasks/filings were loaded.
* Added a watcher on the taskItems and updates the parent (Dashboard) when changes occur in the taskItems array.

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -519,18 +519,6 @@ export default {
           console.log('ERROR - got unknown task =', task)
         }
       })
-
-      this.$emit('task-count', this.taskItems.length)
-      this.$emit('task-items', this.taskItems)
-
-      // If there are any draft/pending/error/paid/correction tasks, emit this event to the parent component.
-      // This indicates that a new filing cannot be started because this item has to be completed first.
-      this.$emit('has-blocker-task',
-        this.taskItems.filter(item => {
-          return (this.isStatusDraft(item) || this.isStatusPending(item) || this.isStatusError(item) ||
-            this.isStatusPaid(item) || this.isTypeCorrection(item))
-        }).length > 0
-      )
     },
 
     loadTodoItem (task) {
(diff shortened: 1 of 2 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -1043,6 +1031,20 @@ export default {
       // if tasks changes, reload them
       // (does not fire on initial page load)
       this.loadData()
+    },
+    taskItems () {
```
severinbeauvais, line 1035: First, this will now fire every time a task is added to this array, which means it's more doing work than it needs to.
Second, why was emitting these values after the main list is processed (ie, where it was before) a problem? I don't understand the timing issue.

severinbeauvais:
> * Bux fix where filings UI was updating the parent before the tasks/filings were loaded.

Perhaps if you need the events to be emitted AFTER the UI is updated, you could just do that after the next tick?

src/components/Dashboard/TodoList.vue
cameron-freshworks: What i found in testing, was the `loadData` was the only place the task `emit`'s were called which happens when the component was created. At point of that initial emit, the taskItems didn't seem to have the values yet, even tho the `loadFilingItem` was being `awaited`.  I could look a little deeper into the asynchronous behaviour here..

cameron-freshworks:
> > * Bux fix where filings UI was updating the parent before the tasks/filings were loaded.
> 
> Perhaps if you need the events to be emitted AFTER the UI is updated, you could just do that after the next tick?

For sure, i'll give this a try!

src/components/Dashboard/TodoList.vue
severinbeauvais: Maybe there's something missing to make the "await loadFilingItem" work?

cameron-freshworks: Absolutely, i'll look deeper into this!

cameron-freshworks: The best explanation i could find...
 "Using Babel will transform async/await to generator function and using forEach means that each iteration has an individual generator function, which has nothing to do with the others. so they will be executed independently and has no context of next()"...

Long story short, await does not seem to work with the forEach function :P

Doing a little refactoring here and fixing some tests!

severinbeauvais:
Except the unit tests still fail in GHA?

cameron-freshworks:
> Except the unit tests still fail in GHA?

Needed a few more flushPromises(), thought it was passing locally 🤔 All good now!

cameron-freshworks:
> Good solution!

thanks!

PR #128, opened 2020-08-20
Removed Sentry.captureException from axios error and added on specific scenario's (main.start, dashboard unavailable dialog) #3933

vysakh-menon-aot:
Removed Sentry.captureException from axios error and added on specific scenario's (main.start, dashboard unavailable dialog) #3933

*Issue #:* /bcgov/entity#3933

*Description of changes:*

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -60,6 +60,7 @@
 import { mapState, mapActions } from 'vuex'
 import axios from '@/axios-auth'
 import KeycloakService from 'sbc-common-components/src/services/keycloak.services'
+import * as Sentry from '@sentry/browser'
 
 // Components
 import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -282,6 +285,7 @@ export default {
         } catch (err) {
           console.log(err) // eslint-disable-line no-console
           this.dashboardUnavailableDialog = true
+          Sentry.captureException(err)
```
severinbeauvais, line 288: Why did you decide to log the error here but not elsewhere? Perhaps add a code comment.

src/main.ts
```diff
@@ -71,6 +71,7 @@ async function start () {
 
 // execution and error handling
 start().catch(error => {
+  Sentry.captureException(error)
```
severinbeauvais, line 74: I'd love to see a comment here indicating the context for why to log this error (but not others).
PS I love comments that help me understand what/why the code is doing something :)

severinbeauvais:
> As a follow-on exercise, it would be nice to see what these logged errors look like in Sentry, and whether we are capturing the right errors and no more (eg, app error, user error, env error, data error)?

I guess I'm looking for some documentation explaining the decision of which errors to log and which not to... What was the criteria for deciding? Why some errors and not others?

PR #129, opened 2020-08-26
launch Correct UI when correcting an IA

severinbeauvais:
**NB1: the redirection will fail until the Correct UI is deployed.**
**NB2: I am assuming what the Correct UI URL will be.**

*Issue #:* /bcgov/entity#4514

*Description of changes:*
- updated k8s data
- updated local configuration
- redirect to Correct UI for IA correction
- loaded new config value
- added unit test
- changed startup console messages to type info


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -875,8 +875,11 @@ export default {
             params: { correctedFilingId: item.filingId } })
           break
         case FilingTypes.INCORPORATION_APPLICATION:
-          // for now, not allowed
-          alert('At this time, you cannot correct an incorporation application. Please contact Ops if needed.')
```
lekshmimallika-aot, line 879: Why arent we using our standard dialog here?

src/utils/fetch-config.ts
```diff
@@ -43,6 +43,10 @@ export async function fetchConfig (): Promise<void> {
   sessionStorage.setItem('CREATE_URL', createUrl)
   console.log('Set Create URL to: ' + createUrl)
 
+  const correctUrl = response.data['CORRECT_URL']
+  sessionStorage.setItem('CORRECT_URL', correctUrl)
+  console.log('Set Correct URL to: ' + correctUrl)
```
lekshmimallika-aot, line 48: Should this be console.info? I see both console.log and console.info statements in this file.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: This is defensive coding only so it uses the easy way instead of using the confirm dialog or some custom dialog.

This is existing code. Do you think I should update it?

src/utils/fetch-config.ts
severinbeauvais: Good catch. I think these should all be console.info. Thanks, fixed in next commit.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -875,8 +875,11 @@ export default {
...
+          // redirect to Correct web app to correct this Incorporation Application
+          const correctUrl = sessionStorage.getItem('CORRECT_URL')
+          const url = `${correctUrl}correction/?id=${item.filingId}`
```
cameron-freshworks, line 880: This will require a change in the framework to use the filingId opposed to the business identifier. I went that path, as i was thinking down the road to Alteration and Profile view as well. Maybe best for now to leave this as is, and alter the correction to take the filingId?

src/utils/fetch-config.ts
```diff
@@ -33,15 +33,19 @@ export async function fetchConfig (): Promise<void> {
 
   const businessesUrl = response.data['BUSINESSES_URL']
   sessionStorage.setItem('BUSINESSES_URL', businessesUrl)
-  console.log('Set Businesses URL to: ' + businessesUrl)
+  console.info('Set Businesses URL to: ' + businessesUrl)
```
cameron-freshworks, line 36: interesting, is the difference the format in which it displays in the console?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Ah, right, good catch.
I guess it does make sense to open Correct UI with a filing in mind.
I will update the code above to refer to the parameter as "filingId" (so noone else confuses it with business id).

src/utils/fetch-config.ts
severinbeauvais: They are potentially styled differently, it may be possible to filter them by type (log/info/warn/erro), and it indicates to another developer that this is informative text instead of something else.
https://developer.mozilla.org/en-US/docs/Web/API/console

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: You bet, will put in the small change to handle the Id in the correct-ui 👍 . It definitely does make sense for a correction!

src/utils/fetch-config.ts
cameron-freshworks: ahh ok, that is a pretty clear distinction. I like it!

src/components/Dashboard/FilingHistoryList.vue
lekshmimallika-aot: No. This is fine.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: For anyone else reading this:
Cam and I decided that, when correcting an IA, the query param is "filingId".
However, when the Correct UI allows filing an alteration, or viewing/updating the business profile, then the query param will be "id" (or perhaps businessId).

PR #130, opened 2020-08-27
Updated build.groovy

severinbeauvais:
*Issue #:* None

*Description of changes:*
- fixed context directory
- removed Ody from submitters list


severinbeauvais:
> This code was put in place back when every pipeline would trigger off any commit to lear. Since we no longer have that issue we can delete the 'if triggerBuild' clause and the function too. It's just in there because we haven't touched the pipeline in so long

Thanks @kialj876 . I've updating this in the second commit.

PR #140, opened 2020-09-18
IA Corrections Dashboard Functionality

cameron-freshworks:
*Issue #:* /bcgov/entity#4646

*Description of changes:*

* Updates the logic in the Filings UI to handle Creation of Draft Incorporation Corrections before redirecting to Edit UI

* Implements Resume / Delete functionality on 'in progress' staff correction filings

* Verifies client is only 'read only` when it comes to Draft IA Corrections and the Dashboard is in a disabled state

* Updates unit testing.


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -876,11 +876,28 @@ export default {
...
+            // redirect to Edit web app to correct this Incorporation Application
+            const editUrl = sessionStorage.getItem('EDIT_URL')
+            const url = `${editUrl}${this.getIncorporationNumber}/correction?correction-id=${draftCorrectionId}`
+
+            // assume Correct URL is always reachable
+            window.location.assign(url)
+          } catch (error) {
+            // eslint-disable-next-line no-console
+            console.log(`Correction Creation error = ${error}`)
+            this.downloadErrorDialog = true
```
severinbeauvais, line 899: Maybe not the right dialog.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1092,6 +1109,47 @@ export default {
...
         disableThisIaCorrection)
+    },
+
+    /**
+     * Fetches a filing by its id.
+     * @param businessId The business identifier
+     * @param filingId The filing identifier
+     * @returns a promise to return the filing of the specified type
+     */
+    fetchFilingById (businessId: string, filingId: number): Promise<any> {
```
severinbeauvais, line 1120: I suggest moving this to legal-api-mixin.ts (best practice, and so others can use this).

src/interfaces/correctionFiling-interface.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/** Incorporation Application filing loaded from / saved to the Legal API. */
+export interface CorrectionFilingIF {
+  header: {
+    name: string
+    certifiedBy: string
+    date: string
+    routingSlipNumber?: string
+    folioNumber?: string
+    effectiveDate?: string
```
severinbeauvais, line 11: I think you're missing some fields. See Vysakh's PR for all the header fields.

src/interfaces/correctionFiling-interface.ts
```diff
@@ -0,0 +1,25 @@
...
+    legalType: string
+    legalName: string
+    identifier: string
+  }
+  correction: {
+    correctedFilingId: string
+    correctedFilingType: string
+    correctedFilingDate: string
+    comment: string
+    incorporationApplication?: IncorporationApplicationIF
```
severinbeauvais, line 23: Yes, optional today, and may have other IFs in the future 👍

src/components/Dashboard/TodoList.vue
```diff
@@ -458,7 +484,7 @@ export default {
   },
 
   computed: {
-    ...mapGetters(['isBComp', 'isCoop', 'isRoleStaff']),
+    ...mapGetters(['getIncorporationNumber', 'isBComp', 'isCoop', 'isRoleStaff']),
```
severinbeauvais, line 487: Incorporation Number is the same as Entity Inc No. Let's stick with a single property :)

(Add a comment where you use this if it helps clarity.)

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -365,7 +365,7 @@ export default {
   },
 
   computed: {
-    ...mapGetters(['isBComp', 'isRoleStaff', 'nrNumber']),
+    ...mapGetters(['getIncorporationNumber', 'isBComp', 'isRoleStaff', 'nrNumber']),
```
severinbeauvais, line 368: Incorporation Number is the same as Entity Inc No... let's stick with a single property :)

Maybe add a comment where this is used if it helps code clarity.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1092,6 +1109,47 @@ export default {
 
       return (this.disableChanges || item.isCorrection || item.isFutureEffectiveIa || item.isColinFiling ||
         disableThisIaCorrection)
+    },
+
+    /**
+     * Fetches a filing by its id.
+     * @param businessId The business identifier
+     * @param filingId The filing identifier
+     * @returns a promise to return the filing of the specified type
```
severinbeauvais, line 1118: There is no reference to type here. Please update this comment.

src/mixins/filing-mixin.ts
```diff
@@ -88,4 +91,54 @@ export default class FilingMixin extends Vue {
...
+        legalType: iaFiling.business.legalType,
+        identifier: iaFiling.business.identifier,
+        legalName: iaFiling.business.legalName
+      },
+      correction: {
+        correctedFilingId: iaFiling.header.filingId,
+        correctedFilingType: FilingTypes.INCORPORATION_APPLICATION,
+        correctedFilingDate: iaFiling.header.date,
+        comment: '',
+        incorporationApplication: {
```
severinbeauvais, line 116: I'm thinking that we don't need all the interfaces you've ported over, since you can use assign the filing object here.
Explanation: other than here, we don't need to know the structure of the IA. I think we can just handle it as an object of type "any" for the purposes of creating a draft correction filing.

src/mixins/filing-mixin.ts
```diff
@@ -1,14 +1,16 @@
...
+import { IncorporationFilingIF } from '@/interfaces/incorporation-interfaces'
 
 /**
  * Mixin that provides some useful filing utilities.
  */
 @Component({
   computed: {
-    ...mapState(['filingData', 'entityType'])
+    ...mapState(['filingData', 'entityType']),
+    ...mapGetters(['getCurrentDate'])
```
severinbeauvais, line 13: There's already a state to get currentDate.
(But if you wanted to change it to a getter, I'm in favour.)

src/interfaces/incorporation-interfaces/incorporationFiling-interface.ts
```diff
@@ -0,0 +1,21 @@
+import { IncorporationApplicationIF } from '@/interfaces/incorporation-interfaces'
+
+/** Incorporation Application filing loaded from / saved to the Legal API. */
+export interface IncorporationFilingIF {
```
severinbeauvais, line 4: Maybe don't need this.

src/components/Dashboard/TodoList.vue
```diff
@@ -816,7 +851,7 @@ export default {
       // open confirmation dialog and wait for response
       this.$refs.confirm.open(
         'Delete Draft?',
-        'Delete your ' + task.draftTitle + '? Any changes you\'ve made will be lost.',
+        'Delete your ' + task.title + '? Any changes you\'ve made will be lost.',
```
severinbeauvais, line 854: This may break other correction types, as they were using "draftTitle". Check where the correction item is loaded to see what properties are set (draftTitle vs title).

src/components/Dashboard/TodoList.vue
```diff
@@ -227,6 +227,32 @@
                   >
                     <span>Resume</span>
                   </v-btn>
+                  <v-menu offset-y left>
```
severinbeauvais, line 230: This code adds the menu for all correction types, but delete only applies to IA correction right?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1092,6 +1109,47 @@ export default {
...
+        })
+    },
+
+    /**
+     * Create a draft correction filing.
+     * @param businessId The business identifier
+     * @param filing the object body of the request
+     * @returns a promise to return the filing
+     */
+    createCorrection (businessId: string, filing: CorrectionFilingIF): Promise<any> {
```
severinbeauvais, line 1140: I think we should get in the habit of putting these calls in a mixin (or service) module.

I suggest generalizing the name ("create filing?") and the filing type can be "any", so that this can be reused.

severinbeauvais (commented):
A few suggestions but the code looks really good (nice of of asyncs, etc).
I'd like to re-review before approving.

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: ah thanks, i just implemented this for testing. I'm not exactly sure which Dialog to use here to be honest.

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: I'm totally good with that, except this project doesn't have a legal-api mixin :)
There is a filing Mixin but it doesn't handle any requests, more just some functionality. Would you like me to create something similar here to what we use in the Create and edit ui?

src/interfaces/correctionFiling-interface.ts
cameron-freshworks: Copy that, will do!

src/interfaces/correctionFiling-interface.ts
cameron-freshworks: Exactly!

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Hmm, ok good point. This is probably a duplicate then, i will stick with the original Entity Inc no. Will fix!

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Updated in next commit!

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Yes, please create new mixin, but only for your one or 2 methods (no need to refactor all other code atm).
Baby steps towards a better code architecture! :)

src/mixins/filing-mixin.ts
cameron-freshworks: For sure, it isn't 100% required but i don't think having the interfaces here is a bad thing. Going forward into Alterations or Company profile view etc we may find that we are wanting these interfaces. 
In a perfect world, we are importing these interfaces, but since we are not there yet, is it ok to leave these in?

src/mixins/filing-mixin.ts
cameron-freshworks: Yea, you sold me on getters awhile ago! Lol Lets stick with em

src/components/Dashboard/TodoList.vue
cameron-freshworks: Will dig into this, i couldn't see the difference  other than 'Correction -' being present, which is why i opted for this. I had a feeling when you deleted a Draft correction previously, it wasn't specified that it was a Correction.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I was looking at what we do on other Todo List errors... and the answer is not much.
But here we really should let the user know that we can't create a draft correction. Maybe something like the Save Error Dialog or Resume Error Dialog in other views?

src/mixins/filing-mixin.ts
severinbeauvais: I think it's better to wait until we can import them. It will lessen the risk of have multiple copies that possibly diverge (thus creating more work to fix later). Plus we're not _really_ using them atm.

src/components/Dashboard/TodoList.vue
cameron-freshworks: This would allow the deletion of any Correction Draft for Staff.

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: For sure! As mentioned above, we don't have a mixin handling requests in the Filings UI. Everything is pretty contained in the components but can certainly create one!

cameron-freshworks:
> A few suggestions but the code looks really good (nice of of asyncs, etc).
> I'd like to re-review before approving.

Absolutely, updating!

src/components/Dashboard/TodoList.vue
severinbeauvais: These properties, `draftTitle` or `title`, as assigned when the Todo List item is loaded into memory. I believe different types of items have different properties. I think, here, drafts/correction may not have `draftTitle` and/or `title`. You can look at the code or test drafts + corrections to verify.

src/interfaces/incorporation-interfaces/incorporationFiling-interface.ts
cameron-freshworks: I'd prefer to leave the interfaces until a time when we are importing them, if possible :)

src/components/Dashboard/TodoList.vue
severinbeauvais: Yes, but since other corrections are currently a manual process, I don't know if the BE handles this correctly.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Yes please!

src/interfaces/incorporation-interfaces/incorporationFiling-interface.ts
severinbeauvais: What do the other reviewers think?

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: K, sounds good, will do!

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Absolutely :) No worries there lol

src/mixins/filing-mixin.ts
cameron-freshworks: Ok, fair points! No worries at all, will update this 👍

src/interfaces/incorporation-interfaces/incorporationFiling-interface.ts
cameron-freshworks: All good, i don't disagree! Will remove :)

src/components/Dashboard/TodoList.vue
cameron-freshworks: Can staff save Director corrections or Address Corrections as Drafts?

severinbeauvais: Hmm, it looks like they can. Weird that there's only a Resume button and not a Delete option.
Check with Kaine/Linda/Scott on whether they want this.
Also, does it work with the existing BE?

cameron-freshworks: It does! I do remember Linda confirming they want to be able to delete Drafts, but i will double check 👍

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Updated to  `LoadCorrection Dialog` with a small tweak there.

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Updated in next commit

src/interfaces/correctionFiling-interface.ts
cameron-freshworks: Updated in next commit

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: So this was just poor naming convention. I created a getter to fetch entityIncNo.. and called it something different :P 
Fixed!

src/mixins/filing-mixin.ts
cameron-freshworks: Updated in next commit

src/components/Dashboard/TodoList.vue
cameron-freshworks: After looking at both properties the naming is a little inconsistent for the `title` prop, will revert to draftTitle.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Would you mind replacing "entityIncNo" with your new getter in this file?

src/mixins/filing-mixin.ts
severinbeauvais: Could you simply do this here?
```
   incorporationApplication: iaFiling.incorporationApplication,
```

src/components/dialogs/LoadCorrectionDialog.vue
```diff
@@ -46,6 +46,6 @@ export default class LoadCorrectionDialog extends Vue {
   @Prop() private attach: string
 
   // Pass click event to parent.
-  @Emit() private exit () { }
+  @Emit('close') private exit () { }
```
severinbeauvais, line 49: I think will break the "exit" handler in Correction.vue (line 9).

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Absolutely!

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -13,6 +13,12 @@
       attach="#filing-history-list"
     />
 
+    <load-correction-dialog
```
severinbeauvais, line 16: Great reuse of this dialog 👍

src/mixins/filing-mixin.ts
cameron-freshworks: I'm not sure, i will try!

src/components/dialogs/LoadCorrectionDialog.vue
cameron-freshworks: Hmm, yea that is a little strange. Going to look into this

vysakh-menon-aot (approved):

src/components/dialogs/LoadCorrectionDialog.vue
cameron-freshworks: Updated 👍

cameron-freshworks: Fixed

src/mixins/filing-mixin.ts
severinbeauvais: Yay :)

severinbeauvais (approved):
Excellent work. Approved with gratitude!

cameron-freshworks:
> Excellent work. Approved with gratitude!

hahah the gratitude is all mine, appreciated!

PR #142, opened 2020-09-24
cleaned up enum names

severinbeauvais:
*Issue #:* None

*Description of changes:*
- renamed LegalTypes to EntityTypes
- removed unneeded LegalNames
- renamed typeEnum to entityType

The purposes of this change are:
1. to rename the "legal types" enum to be the same as in the Create UI and Edit UI, and which is in line with what the property is named
2. to remove an enum that seemed to overload a schema property

Note that "entity type" is also known by other names, ie:
1. `business.legalType`
2. `filing.business.legalType`
3. `alteration.alterCorpType.corpType`
(and maybe others)


cameron-freshworks (approved):

cameron-freshworks:
Looks good, though should this clean up wait until after Kials changes to the Bcomp identifier? 
Not 100% sure on the changes but BCOMP's would be referred to as `BEN` i think.

severinbeauvais:
> Looks good, though should this clean up wait until after Kials changes to the Bcomp identifier?
> Not 100% sure on the changes but BCOMP's would be referred to as `BEN` i think.

The 'BEN' change is a different ticket.

severinbeauvais:
> So in a perfect world, we would just need to update our Bcomp value in the enum... though that can certainly come later once those changes are commit on the BE.

This is getting us closer to the perfect world where we use the same types, names, etc :)

kialj876 (approved):

severinbeauvais:
This PR has been reviewed and approved, and I just confirmed that it would not "undo" any changes recently made (ie, BC -> BEN), so now I'm merging it...

severinbeauvais:
Oops, merge conflicts... will fix next.

PR #150, opened 2020-10-08
Feature/5091

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5091

*Description of changes:*
 - sbc common component version update

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "provinces": "^1.11.0",
     "regenerator-runtime": "^0.13.3",
     "register-service-worker": "^1.6.2",
-    "sbc-common-components": "2.4.1",
+    "sbc-common-components": "2.3.8",
     "v-mask": "^2.1.0",
     "vue": "^2.6.11",
     "vue-affix": "^0.5.2",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "provinces": "^1.11.0",
     "regenerator-runtime": "^0.13.3",
     "register-service-worker": "^1.6.2",
-    "sbc-common-components": "^2.3.7",
+    "sbc-common-components": "2.4.1",
     "v-mask": "^2.1.0",
     "vue": "^2.6.11",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "provinces": "^1.11.0",
     "regenerator-runtime": "^0.13.3",
     "register-service-worker": "^1.6.2",
-    "sbc-common-components": "^2.3.7",
+    "sbc-common-components": "2.4.1",
```
severinbeauvais, line 31: I think this change is premature. According to /bcgov/entity#5244 this change is targeted for next sprint and will require synchronization between the UI config and some OS changes (done by Relationships team).

severinbeauvais (changes requested):

package.json
severinbeauvais: OK, I see you did all 3 UIs now (that solves part of the problem).

Now I'm wondering what change was made to the common components? I assume you updated some images and/or text, but I think there are Keycloak-related changes to (as per my 5244 comment).

Let's discuss with POs and Sumesh on the timing for this/these changes... Please start the discussion in RocketChat. Thx.

vysakh-menon-aot: So should I hold until then?

vysakh-menon-aot: No update on images. Only text change

severinbeauvais: Well, (your PR)[https://github.com/bcgov/sbc-common-components/pull/191/files] went into version 2.3.8, so we could do just that...

On the other hand, it's the same amount of work to update to 2.4.1, so it might be nice to kill two birds with one PR (as it were)... if we can get everyone in sync over this.

Let's discuss and decide together.

severinbeauvais: Sumesh says we shouldn't go to 2.4.1 unless we also do some Keycloak changes.

Per standup today, please update to 2.3.8.

severinbeauvais (approved):

cameron-freshworks (approved):

PR #152, opened 2020-10-23
Correct time does not appear for Detail after filing a correction

vysakh-menon-aot:
*Issue #:* /bcgov/entity#5355

*Description of changes:*
 - filingDate instead of correctedFilingDate in FilingHistoryList

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -750,7 +750,7 @@ export default {
           filingId: header.filingId,
           filingAuthor: header.certifiedBy,
           filingDateTime: this.convertUTCTimeToLocalTime(header.date), // used for receipt
-          filingDate: this.convertUTCTimeToLocalTime(correction.correctedFilingDate),
+          filingDate: filingDate,
           isCorrection: true,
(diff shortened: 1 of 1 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -750,7 +750,7 @@ export default {
           filingId: header.filingId,
           filingAuthor: header.certifiedBy,
           filingDateTime: this.convertUTCTimeToLocalTime(header.date), // used for receipt
-          filingDate: this.convertUTCTimeToLocalTime(correction.correctedFilingDate),
+          filingDate: filingDate,
```
lekshmimallika-aot, line 753: You can just use the shorthand here (just put filingDate).

lekshmimallika-aot (approved):

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: While you're here, is `filingDateTime` also broken? And same solution?

severinbeauvais (approved):
I've requested a change but if you don't think you should do it, or if you do it, then I'm OK for you to merge after that.

PR #157, opened 2020-11-09
fixed COD issues

severinbeauvais:
*Issues:* /bcgov/entity#5454, /bcgov/entity#5442, /bcgov/entity#4429

*Description of changes:*

Commit 1:
- added new error dialog
- generalized an interface

Commit 2:
- allow COD with no body in Todo List

Commit 3:
- updated an event (and hander) name
- updated a computed method name
- changed fetchData() to async
- await fetchData() in created()
- properly restore COD Draft Date when resuming a filing
- don't double-update filing data when resuming a filing
- don't double-fetch directors when resuming a filing

Commit 4:
- updated element ids
- refactored conditional around addresses block (bugfix)
- refactored conditional around buttons
- added disabled Reset button (for future)
- created interfaces to enforce typings
- added types to properties, etc
- changed director id to number
- cleaned up comments, etc
- changed ".filter" to ".some"
- removed useless mount()
- replaced JSON stringify/parse with cloneDeep
- made getDirectors() async (bugfix)
- emit event on fetch error
- added conditional around mailing address (bugfix)
- replaced JSON comparison with isEqual
- refactored asOfDate watchers
- await getDirectors() (bugfix)
- added Fetch Error Dialog
- added fetch error event handling
- created/updated unit tests


src/components/common/Directors.vue
```diff
@@ -359,39 +360,41 @@
                     />
                   </div>
 
-                  <label class="address-sub-header">Delivery Address</label>
```
severinbeauvais, line 362: This was a bug where "Delivery Address" would display when changing a director's name.

src/components/common/Directors.vue
```diff
@@ -452,27 +455,44 @@
...
+                        @click="deleteDirector(director.id)"
+                      >
+                        <span>Remove</span>
+                      </v-btn>
+                    </div>
+
+                    <!-- v-show doesn't support <template> so use a <div> instead -->
+                    <div v-show="!isNew(director)">
+                      <!-- FUTURE: implement this -->
+                      <v-btn color="error" outlined
```
severinbeauvais, line 471: I added this button to be consistent with change of director name (but I left it disabled as implementing the functionality is beyond the scope of these tickets).

src/components/common/Directors.vue
```diff
@@ -524,7 +545,28 @@ import { CommonMixin, DateMixin, DirectorMixin, ResourceLookupMixin } from '@/mi
 import { Actions, FilingStatus } from '@/enums'
 
 // Interfaces
-import { FormIF, BaseAddressIF, AlertMessageIF } from '@/interfaces'
+import { FormIF, AddressIF, ComponentIF, AlertMessageIF } from '@/interfaces'
+
+// Interfaces used locally only (since they don't match global interfaces)
+interface OfficerIF {
```
severinbeauvais, line 551: This interface and the one below are not used outside of this file so I just created them locally. Having them really helped with dev and debugging!

src/components/common/Directors.vue
```diff
@@ -545,44 +587,55 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
 
-  /**
-   * Indicates whether this component should be enabled or not.
-   */
+  /** Indicates whether this component should be enabled or not. */
   @Prop({ default: true })
   private componentEnabled: boolean
 
-  private directorEditInProgress: boolean = false;
+  // Local definitions of computed properties for static type checking.
```
severinbeauvais, line 607: I simply moved these up from lower down in the file. (I like having class variables declared at the top.)

src/components/common/Directors.vue
```diff
@@ -676,61 +733,34 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
   }
 
-  /**
-   * Computed value.
-   * One or more actions taken on directors (add, cease) require a single fee, so check if at least one
-   * director in the list is marked as requiring a fee.
-   * @returns Whether at least one director has a fee applied.
-   */
-  private get directorsChange (): boolean {
-    return this.directors.filter(director => director.isFeeApplied).length > 0
```
severinbeauvais, line 714: "filter" goes through the whole list, while "some" stops when it finds a match.

src/components/common/Directors.vue
```diff
@@ -795,36 +821,19 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-    rules.push(rule2)
+    rules.push(v => this.dateIsNotFuture(v) || 'Cessation Date cannot be in the future')
 
     return rules
   }
 
-  /**
-   * Called when component is mounted.
-   */
-  private mounted (): void {
```
severinbeauvais, line 815: This never did anything because the As Of Date wasn't set at this point in time, so the directors never got fetched.

src/components/common/Directors.vue
```diff
@@ -845,32 +854,26 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-  private getOriginalDirectors () {
-    this.getDirectors(true)
-  }
-
   /**
    * Function called internally and externally to fetch the list of directors.
-   * FUTURE: change this to a prop?
    */
-  public getDirectors (getOrigOnly: Boolean = false): void {
+  public async getDirectors (getOrigOnly: boolean = false): Promise<void> {
```
severinbeauvais, line 860: This was pretty important to the bugs here. Since this was not async (and not awaited by its callers), some things would sometimes happen in different orders (ie, timing bug).

src/components/common/Directors.vue
```diff
@@ -1124,45 +1128,53 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-      if (this.inheritDeliveryAddress) {
-        director.mailingAddress = director.deliveryAddress
+        if (this.inheritDeliveryAddress) {
+          director.mailingAddress = director.deliveryAddress
+        }
       }
 
       /* COMPARE changes to original director data, for existing directors */
       if (director.actions.indexOf(Actions.APPOINTED) < 0) {
-        const origDirector = this.directorsOriginal.filter(el => el.id === id)[0]
```
severinbeauvais, line 1137: "filter" goes through the whole list, while "find" stops when it finds a match.

Also the new code checks that director was found before dereferencing ("[0]").

src/components/common/Directors.vue
```diff
@@ -1124,45 +1128,53 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
+        const origDirector = this.directorsOriginal.find(director => director.id === id)
 
-        // check whether address has changed
-        if ((JSON.stringify(origDirector.deliveryAddress) !== JSON.stringify(director.deliveryAddress)) ||
```
severinbeauvais, line 1140: This doesn't always work because order of properties is not guaranteed.

Anyway, "isEqual" is cleaner.

src/components/common/Directors.vue
```diff
@@ -1325,128 +1337,105 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-  /**
-   * When as-of date is initialized, set initial director appointment date.
-   */
-  @Watch('asOfDate', { immediate: true })
-  private onAsOfDateInitialized (date: string): void {
+  /** Called when As Of Date is set or changes (from parent component). */
+  @Watch('asOfDate')
+  private async onAsOfDateChange (date: string): Promise<void> {
+    // set director appoinment date for new directors
     this.director.appointmentDate = date
```
severinbeauvais, line 1369: This did nothing in the normal code because As Of Date is initialized to null.

This was used in some unit tests where Prop Data was set when mounting this component. But it's a bad idea for the tests to run differently from normal code.

src/interfaces/address-interfaces.ts
```diff
@@ -1,9 +1,3 @@
-import Vue from 'vue'
```
severinbeauvais, line 1: I moved this to its own file.

src/interfaces/address-interfaces.ts
```diff
@@ -1,9 +1,3 @@
...
-
-export interface BaseAddressIF extends Vue {
```
severinbeauvais, line 3: I moved this to its own file.

src/views/AnnualReport.vue
```diff
@@ -110,7 +116,7 @@
                       directors that were ceased at a later date.</p>
                   </header>
                   <directors ref="directorsList"
-                    @directorsChange="directorsChange"
+                    @directorsPaidChange="directorsPaidChange"
```
severinbeauvais, line 119: Renamed to be consistent with "directorsFreeChange" below.

src/views/AnnualReport.vue
```diff
@@ -526,40 +541,44 @@ export default {
...
       const url = `businesses/${this.entityIncNo}/filings/${this.filingId}`
-      axios.get(url).then(response => {
+      await axios.get(url).then(async response => {
         if (response && response.data) {
           const filing = response.data.filing
           try {
             // verify data
             if (!filing) throw new Error('Missing filing')
             if (!filing.header) throw new Error('Missing header')
+            const header = filing.header
```
severinbeauvais, line 553: The diff below looks worse than it is. I simply created header/business/annualReport to simplify the code below.

lekshmimallika-aot (approved):

src/views/AnnualReport.vue
```diff
@@ -569,71 +588,56 @@ export default {
...
             if (changeOfDirectors) {
-              if (changeOfDirectors.directors && changeOfDirectors.directors.length > 0) {
-                if (this.$refs.directorsList && this.$refs.directorsList.setAllDirectors) {
+              if (changeOfDirectors?.directors.length > 0) {
+                if (this.$refs.directorsList?.setAllDirectors) {
                   this.$refs.directorsList.setAllDirectors(changeOfDirectors.directors)
                 }
-
-                // add filing code for paid changes
-                if (changeOfDirectors.directors.filter(
```
severinbeauvais, line 599: This is unnecessary code and caused duplicate filing data updates.
The Directors component emits events back to THIS file to set the filing data accordingly.

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -767,60 +794,37 @@ export default {
             const changeOfDirectors = filing.changeOfDirectors
             if (changeOfDirectors) {
               if (changeOfDirectors.directors && changeOfDirectors.directors.length > 0) {
-                if (this.$refs.directorsList && this.$refs.directorsList.setAllDirectors) {
+                if (this.$refs.directorsList.setAllDirectors) {
                   this.$refs.directorsList.setAllDirectors(changeOfDirectors.directors)
                 }
-
-                // add filing code for paid changes
-                if (changeOfDirectors.directors.filter(
```
severinbeauvais, line 775: Unneeded code. It causes duplicate updates. The Directos component already emits events to do this.

tests/unit/BcolErrorDialog.spec.ts
```diff
@@ -15,7 +13,7 @@ describe('BcolErrorDialog - Verify parameters passed in are displayed correctly'
     const wrapper = shallowMount(BcolErrorDialog,
       {
         propsData: {
-          filingType: FilingTypes.ANNUAL_REPORT,
+          filingType: 'annualReport',
```
severinbeauvais, line 16: Tests should use hard-coded types, so that they break if you change the type and forget to update the tests.

tests/unit/Directors.spec.ts
```diff
@@ -105,13 +106,8 @@ describe('Directors as a COOP', () => {
     const instance = new Constructor({ store, vuetify })
     vm = instance.$mount()
 
-    // set as-of date
+    // set As Of Date to trigger watcher
     vm.asOfDate = '2019-04-01'
-
-    // set vm2 for draft test (setup is different and does not fire the 'asOfDate' watcher)
```
severinbeauvais, line 111: This appears to be some work-around unit test. The As Of Date watcher is always fired (for new or draft COD) so this unit test is obsolete.

tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -388,9 +388,28 @@ describe('Standalone Directors Filing - Part 2A - Resuming with FAS staff paymen
     // verify that we stored the Filing ID
     expect(+vm.filingId).toBe(123)
 
-    // verify that we loaded the director data correctly
-    expect(vm.filingData.filter(el => el.filingTypeCode === 'OTCDR').length).toEqual(1)
```
severinbeauvais, line 392: As the component under test was shallow-mounted, the Directors sub-component does not emit events to update the filing data.
So I moved these checks out of here and into their own test below (where I explicitly call the event handlers).

vue.config.js
```diff
@@ -42,6 +42,13 @@ module.exports = {
         pathRewrite: {
           '/local-keycloak-config-url': ''
         }
+      },
+      // this is needed to avoid a PAYBC Not Found error when running locally
+      '/status/PAYBC': {
```
severinbeauvais, line 47: This will go away when I rebase.

cameron-freshworks (approved):
Looks good!

PR #159, opened 2020-11-11
fixes for AR/COA/COD + misc cleanup

severinbeauvais:
*Issues:*
/bcgov/entity#2461 - new filings should query latest data
/bcgov/entity#5550 - proper spinner timing
/bcgov/entity#5442 - reworked COD to make it similar to COA
/bcgov/entity#5622 - middle initial fix (was previously only in bugfix branch)

*Description of changes:*
- updated project version
- cleaned up property and method names for clarity and consistency between filings
- moved local interfaces to interfaces folder
- refactored props and events (ie, general interface between parents and children)
- created "empty" objects for cleaner init
- replaced JSON comparison with deep watcher
- cleaned up fetch and load methods
- generally improved comments to help current and future developers
- maintain all address fields (incl actions and addressType) from start to finish
- fixed created/mounted flow
- removed current date conversion (not needed since its refactor to yyyy-mm-dd)
- updated isSame() same as other projects
- updated filing data update to fix its reactivity sometimes not working
  - fixes Fee Summary sometimes not updating (BUGFIX)
- added comments about future improvements (or things that will break)
- updated main spinner to display until data is loaded (5550)
- added alternate spinner for reloads (ie, due to changed date) (5550)
- made date change reloads async
- added fetch error dialog to COA (same as COD and AR)
- don't set Have Changes flag until initial data is loaded
  - fixes "you have unsaved data" when there were no changes (BUGFIX)
- fixed office addresses unit tests
- **NB: some other unit tests are current disabled; I am updating them and will submit them in a future PR**


src/components/common/Directors.vue
```diff
@@ -598,11 +598,11 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
 
   /** Effective date for querying current directors and appointing/ceasing them. */
   @Prop({ default: '' })
-  private asOfDate: string
+  readonly asOfDate: string
```
severinbeauvais, line 601: All props should be `readonly` because you can't set them.

src/components/common/Directors.vue
```diff
@@ -83,7 +83,7 @@
                     <base-address ref="baseAddressNew"
                       :editing="true"
                       :schema="addressSchema"
-                      @update:address="updateBaseAddress"
+                      @update:address="updateDeliveryAddress"
```
severinbeauvais, line 86: This Directors component works fine. This is just a bit of cleanup / commonality with OfficeAddresses component.

src/components/common/OfficeAddresses.vue
```diff
@@ -240,35 +253,14 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
   @Prop({ default: null })
-  readonly addresses: BcorpAddressIf | null
+  readonly addresses: RegRecAddressesIF
 
-  /**
-   * Registered Office address object passed in from the parent which is pulled from store.
-   * This address is used as a baseline address in the event the user wants to reset the changes before submitting.
-   */
-  @Prop({ default: null })
-  private registeredAddress: BaseAddressObjIF
```
severinbeauvais, line 250: The fundamental change here is that this component doesn't care what's in the store -- it will (re)load the office addresses for the specified As Of Date. (The COA filing is always "as of" today, but the AR filing can set an "as of" date.) (The API does not yet support returning office addresses "as of" a specified date but now the UI is ready for it.)

src/components/common/OfficeAddresses.vue
```diff
@@ -278,185 +270,135 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
+        })
     }
+  }
 
-    if (this.recordsAddress) {
-      recDeliveryAddressFinal = { ...this.recDeliveryAddress }
-      recMailingAddressFinal = { ...this.recMailingAddress }
+  /** Called when As Of Date prop is set or changes. */
+  @Watch('asOfDate', { immediate: true })
+  private async onAsOfDateChanged (date: string): Promise<void> {
```
severinbeauvais, line 351: This watcher is the trigger for fetching the original addresses from the API. It all depends on this.

src/components/common/OfficeAddresses.vue
```diff
@@ -278,185 +270,135 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
-    if (this.recordsAddress) {
-      this.recDeliveryAddress = isEmpty(this.addresses)
-        ? {} as AddressIF
-        : this.addresses.recordsOffice['deliveryAddress']
-
-      this.recMailingAddress = isEmpty(this.addresses)
-        ? {} as AddressIF
-        : this.addresses.recordsOffice['mailingAddress']
-    }
+  private onAddressesChanged (): void {
```
severinbeauvais, line 369: FYI, since "addresses" is a prop, when we emit an update event for it (ie, sync modifier in parent) then we receive it back here.

To me it feels weird to "hear back from the parent that we updated its addresses", but then we (component) can't update the prop directly anyway, so I guess this is how it goes.

src/components/common/OfficeAddresses.vue
```diff
@@ -466,188 +408,130 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
-   * Assign the Records address objects actions and omit the addressType
-   *
-   * @param addressBase The base address object to be parsed and assigned to its Office Address class property
-   */
-  private assignRecordAddresses (addressBase: BaseAddressObjIF): void {
-  // Assign Delivery Address
-    const deliveryAddress = addressBase.deliveryAddress
-    if (deliveryAddress) {
-      deliveryAddress.actions = []
-      this.recDeliveryAddressOriginal = { ...deliveryAddress }
```
severinbeauvais, line 630: This code was originally part of the addresses initialization.

You'll notice here that any "actions" are cleared (for easier comparison in this component) but I thought this was worth keeping and I simply updated the comparison code to account for it.

You'll also notice that the original addresses are set here, which is one of the bugs I fixed -- when the user resumes a draft filing, we should NOT assign the original addresses from the draft data -- we should instead load the current addresses from the API.

src/interfaces/address-interfaces.ts
```diff
@@ -1,23 +1,24 @@
...
-  streetAddressAdditional?: string;
+  actions?: string[]
+  addressCity: string
+  addressCountry: string
+  addressRegion: string
+  deliveryInstructions?: string
+  postalCode: string
+  streetAddress: string
+  streetAddressAdditional?: string
+  addressType?: 'mailing' | 'delivery'
```
severinbeauvais, line 11: I added "addressType" so that I could maintain through the data and not have type errors.

src/interfaces/address-interfaces.ts
```diff
@@ -1,23 +1,24 @@
...
+  deliveryAddress: AddressIF
+  mailingAddress: AddressIF
 }
 
-// Interface to define the Bcorps address response
-export interface BcorpAddressIf {
-  registeredOffice: BaseAddressObjIF;
-  recordsOffice: BaseAddressObjIF;
+/** Combined Registered + Records office addresses interface. */
+export interface RegRecAddressesIF {
```
severinbeauvais, line 21: May not be Bcorp-specific in near future.

src/views/AnnualReport.vue
```diff
@@ -43,12 +43,15 @@
     />
 
     <!-- Initial Page Load Transition -->
-    <div class="loading-container auto-fade-out">
-      <div class="loading__content">
-        <v-progress-circular color="primary" :size="50" indeterminate></v-progress-circular>
-        <div class="loading-msg">{{loadingMessage}}</div>
+    <v-fade-transition>
+      <!-- <div class="loading-container" v-show="showLoadingContainer"> -->
```
severinbeauvais, line 47: This is preparation for a future bug that I almost fixed in this PR...

The problem is that it's not just the parent component that awaits certain things (like fetching a draft) -- it's also the child component(s) (like OfficeAddresses and Directors) that await the fetching of their data from the API. ATM the parent doesn't know the child(ren) is(are) busy so I can't properly implement the "showLoadingContainer" variable.

It's the same problem in COA and COD filing pages.

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -738,76 +750,77 @@ export default {
     async fetchChangeOfDirectors (): Promise<void> {
       const url = `businesses/${this.entityIncNo}/filings/${this.filingId}`
       await axios.get(url).then(async response => {
-        if (response && response.data) {
-          const filing = response.data.filing
-          try {
```
severinbeauvais, line 743: I got rid of this redundant try-catch becayse the next line verifies that `filing` exists (and throws if not).

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -738,76 +750,77 @@ export default {
...
+        if (filing.header.effectiveDate) {
+          const effectiveDate = this.convertUTCTimeToLocalTime(filing.header.effectiveDate).slice(0, 10)
+          // restore date in COD Date component
+          this.initialCODDate = effectiveDate
+          // restore Draft Date in Directors List component
+          // FUTURE: use props instead of $refs (which cause an error in the unit tests)
+          if (this.$refs.directorsList?.setDraftDate) {
+            this.$refs.directorsList.setDraftDate(effectiveDate)
+          }
+        } else if (filing.header.date) {
```
severinbeauvais, line 800: I added this fallback logic -- if "effective date" isn't present then try to use "filing date" (which was set to current date when this draft was saved).

There's a similar sort of thing in the AR filing: if there is no effective data then fall back to AGM date.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -365,95 +393,93 @@ export default {
...
-              this.staffPaymentData = {
-                option: StaffPaymentOptions.NO_FEE
-              }
-            } else {
-              this.staffPaymentData = {
-                option: StaffPaymentOptions.NONE
-              }
-            }
+      await axios.get(url).then(response => {
+        const filing = response?.data?.filing
```
severinbeauvais, line 399: Hide whitespace changes to see this better.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -365,95 +393,93 @@ export default {
...
+          this.coaDate = this.convertUTCTimeToLocalTime(filing.header.effectiveDate).slice(0, 10)
+        } else if (filing.header.date) {
+          this.coaDate = this.convertUTCTimeToLocalTime(filing.header.date).slice(0, 10)
+        } else {
+          throw new Error('Missing effective date')
+        }
+
+        const changeOfAddress = filing.changeOfAddress?.offices
+        if (changeOfAddress) {
+          if (changeOfAddress.recordsOffice && changeOfAddress.registeredOffice) {
```
severinbeauvais, line 448: FYI, I'm a little uneasy about checking the data for what it contains -- instead of requiring specific data to exist (ie, for BCOMP). But I didn't change this here. I think we will have to look at this again when we have other entity types with 1 or 2 sets of addresses.

src/interfaces/address-interfaces.ts
```diff
@@ -1,23 +1,34 @@
...
+  addressRegion: string
+  deliveryInstructions?: string
+  postalCode: string
+  streetAddress: string
+  streetAddressAdditional?: string
+  addressType?: 'mailing' | 'delivery'
 }
 
-// Interface to define the joint base address response
-export interface BaseAddressObjIF {
```
severinbeauvais, line 14: This name suggests it's the address used for "base address component", which is incorrect. So I renamed it per its current usage. (It's not used for director addresses atm.)

src/components/StandaloneDirectorChange/CODDate.vue
```diff
@@ -98,7 +98,7 @@ export default class CodDate extends Mixins(DateMixin) {
    * @returns The maximum date that can be entered.
    */
   private get maxDate (): string {
-    return this.currentDate ? this.currentDate.split('/').join('-') : null
```
severinbeauvais, line 101: At some point, current date was changed to format YYYY-MM-DD (instead of YYYY/MM/DD) so this split/join is unnecessary.

You will see that I cleaned this up globally (and in unit tests where it's set).

src/components/common/Directors.vue
```diff
@@ -61,19 +61,19 @@
                     <v-text-field filled class="item"
                       label="First Name"
                       id="new-director__first-name"
-                      v-model="director.officer.firstName"
```
severinbeauvais, line 64: There were too many things named "director" or "directors" !!!

The new name identifies this variable's specific purpose.

src/components/common/Directors.vue
```diff
@@ -179,55 +179,56 @@
...
-        <v-subheader v-if="this.directors.length && !directorEditInProgress" class="director-header">
+        <v-subheader v-if="allDirectors.length && !directorEditInProgress" class="director-header">
           <span>Names</span>
           <span>Delivery Address</span>
           <span v-if="isBComp">Mailing Address</span>
           <span>Appointed/Elected</span>
         </v-subheader>
+        <!-- START FOR LOOP -->
         <li class="director-list-item"
-          v-for="(director, index) in directors"
```
severinbeauvais, line 189: Again, too many things named "director".

Since this one is a loop variable, I thought of just using "d" (and "i" for index) but in the end I went with "dir" for the new name. But if we write something like this again, I would go with "d" and "i".

src/components/common/Directors.vue
```diff
@@ -542,31 +543,10 @@ import { WarningPopover } from '@/components/dialogs'
...
 
 // Enums
-import { Actions, FilingStatus } from '@/enums'
+import { Actions } from '@/enums'
 
 // Interfaces
-import { FormIF, AddressIF, ComponentIF, AlertMessageIF } from '@/interfaces'
-
-// Interfaces used locally only (since they don't match global interfaces)
-interface OfficerIF {
```
severinbeauvais, line 551: I moved these to main interfaces folder so we can reuse them.

src/components/common/Directors.vue
```diff
@@ -591,75 +570,56 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-    editDirectorForm: Array<FormIF>,
-    baseAddressEdit: Array<ComponentIF>,
-    mailAddressEdit: Array<ComponentIF>
+    editDirectorForm: FormIF[],
+    baseAddressEdit: ComponentIF[],
+    mailAddressEdit: ComponentIF[]
   }
 
-  /** Effective date for querying current directors and appointing/ceasing them. */
-  @Prop({ default: '' })
```
severinbeauvais, line 600: I moved these down in this file for consistency between vue files.

This is not a rule but I like to see files laid out this way:
- external getters (from store)
- props
- local variables
- internal getters
- created/mounted
- other main methods
- helper methods
- watchers
- emitters

src/components/common/Directors.vue
```diff
@@ -591,75 +570,56 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-    },
-    appointmentDate: null,
-    cessationDate: null,
-    cessationDateTemp: null,
-    actions: []
-  }
-  private inProgressAddress: AddressIF = null
+
+  /** V-model for new director form. */
+  private newDirector: DirectorIF = { ...EmptyDirector }
```
severinbeauvais, line 620: This is close to the "copy constructor" concept. Maybe some day we can go to that.

src/components/common/Directors.vue
```diff
@@ -736,30 +696,25 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
    */
-  private readonly directorLastNameRules: Array<Function> = [
+  private readonly directorLastNameRules: Function[] = [
     v => !!v || 'A last name is required',
     v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
     v => !/\s$/g.test(v) || 'Invalid spaces' // trailing spaces
   ]
 
-  /** JSON representation of directors list for detecting array size and property changes. */
-  private get directorsJson (): string {
```
severinbeauvais, line 746: This was to be able to identify changes to arrays/sub-objects. This is not needed if you do a deep watch, so I changed it.

src/components/common/Directors.vue
```diff
@@ -831,98 +786,105 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
+      streetAddressAdditional: address.streetAddressAdditional || ''
     }
   }
 
-  /**
-   * Function called externally to set the draft date.
-   * FUTURE: change this to a prop
-   * @param date The draft date to set.
-   */
-  public setDraftDate (date): void {
```
severinbeauvais, line 853: This was originally needed for some logic in this file. But that logic belongs in the parent, so I moved it.

src/components/common/Directors.vue
```diff
@@ -831,98 +786,105 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-              if (directors[i].mailingAddress) {
-                directors[i].mailingAddress = this.formatAddress(directors[i].mailingAddress)
-              }
+      await axios.get(url).then(response => {
+        if (response?.data?.directors) {
+          const directors = response.data.directors as DirectorIF[]
+
+          this.original = directors.sort(this.fieldSorter(['lastName', 'firstName', 'middleName']))
+
+          this.original.forEach((director, i) => {
```
severinbeauvais, line 814: Simple change from for loop to forEach loop. (Ignore whitespace changes to see this better.)

src/components/common/Directors.vue
```diff
@@ -831,98 +786,105 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-        .catch(error => {
-          // eslint-disable-next-line no-console
-          console.log('getDirectors() error =', error)
-
-          // emit event to parent to display Fetch Error Dialog
-          this.$root.$emit('fetch-error-event')
-        })
+  /** Called externally to _asynchronously_ get the original directors on As Of Date. */
+  // FUTURE: this logic should be in the parent component
+  public async getOrigDirectors (asOfDate: string, updateWorkingDirectors: boolean): Promise<void> {
```
severinbeauvais, line 848: This is very similar to a method in OfficeAddresses (ie, consistency of architecture).

src/components/common/Directors.vue
```diff
@@ -1214,32 +1172,23 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
   private updateMailingAddress (val: AddressIF): void {
     this.inProgressMailAddress = val
   }
 
-  /**
-   * Function called externally to set all directors from existing filing.
-   * FUTURE: change this to a prop
-   * @param directors The list of directors to set.
-   */
-  public setAllDirectors (directors: DirectorIF[]): void {
```
severinbeauvais, line 1239: Again, I refactore the interface between parent and child (ie, API) and this method was not needed.

src/components/common/Directors.vue
```diff
@@ -1341,70 +1288,32 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
     return (director.isDirectorActionable !== undefined) ? director.isDirectorActionable : true
   }
 
-  /** Called when there is a change (ie, some <-> none) in directors with a paid change. */
-  @Watch('directorsPaidChange')
```
severinbeauvais, line 1345: I simplified this by sending these events every time we emit an update (see new lines 1303-1304).

src/components/common/Directors.vue
```diff
@@ -1341,70 +1288,32 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
   @Watch('directorFormValid')
-  private onDirectorFormValid (val: boolean): void {
+  private onDirectorFormValidChanged (): void {
     // emit event back up to parent
-    this.emitDirectorFormValid(val)
+    this.emitDirectorFormValid()
   }
 
-  /** Called when As Of Date is set or changes (from parent component). */
-  @Watch('asOfDate')
```
severinbeauvais, line 1366: Again, architecture.

It should not be THIS component that monitors whether to re-fetch the original directors and whether to update the working copy; it should be the parent.

severinbeauvais: To be clear, THIS component should concern itself with displaying/editing its specific data, not high-level logic. (I think even the fetch should not be in this component but in the parent or some common mixin/service.)

severinbeauvais: PS I refactored AR/COA/COA (parents) and Directors/OfficeAddresses (children) to implement the correct architecture.

src/components/common/Directors.vue
```diff
@@ -1413,31 +1322,45 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
   /** Emits an event that indicates a director compliance warning has been triggered. */
   @Emit('complianceDialogMsg')
-  private emitcomplianceDialogMsg (val: AlertMessageIF): void { }
+  private emitcomplianceDialogMsg (): AlertMessageIF {
+    return this.complianceMsg
+  }
+
+  /** Emits original addresses object to the parent page. */
+  @Emit('original')
+  private emitOriginalDirectors (): DirectorIF[] {
```
severinbeauvais, line 1355: I'd argue THIS component should not be telling the parent about the original directors list, but rather that the parent should be telling this component (if it even needs to know).
But that's for another day.

src/components/common/OfficeAddresses.vue
```diff
@@ -75,16 +77,16 @@
                   v-model="inheritDeliveryAddress"
                 />
               </div>
-              <div class="address-wrapper"
-                v-if="!isSame(deliveryAddress, mailingAddress, 'actions') || showAddressForm"
+              <div class="address-wrapper" v-if="showAddressForm ||
```
severinbeauvais, line 80: In an "or" clause, it's more efficient to check the easy/fast things first :)

src/components/common/OfficeAddresses.vue
```diff
@@ -75,16 +77,16 @@
...
+                !isSame(deliveryAddress, mailingAddress, ['actions','addressType'])"
```
severinbeauvais, line 81: All addresses now always have actions and addressType properties, so need to omit both in this comparison.

src/components/common/OfficeAddresses.vue
```diff
@@ -130,7 +132,9 @@
               <div class="meta-container">
                 <label>{{ showAddressForm ? "Mailing Address" : "" }}</label>
                 <div class="meta-container__inner">
-                  <label v-if="!isSame(recDeliveryAddress, recMailingAddress, 'actions') && !showAddressForm">
+                  <label v-if="!showAddressForm &&
```
severinbeauvais, line 135: In this "and" clause... same comment as for "or" clause.

(Because the execution may not have to check the remaining conditions.

src/components/common/OfficeAddresses.vue
```diff
@@ -221,242 +228,169 @@ import { Actions } from '@/enums'
...
-  /**
-   * Records Office address object passed in from the parent which is pulled from store.
-   * This address is used as a baseline address in the event the user wants to reset the changes before submitting.
+   * Addresses object from parent page, used to set our working addresses (4 properties below).
+   * As needed, we emit this back to parent (ie, update/sync).
    */
-  @Prop({ default: null })
-  private recordsAddress: BaseAddressObjIF
+  @Prop({ default: () => {} })
+  readonly addresses: RegRecAddressesIF
```
severinbeauvais, line 250: This prop object is now correctly initialized as an object (ie, so we can test "if addresses.something", which you can't do if it's null).
This needed a change to check "ifEmpty" in a few places.

src/components/common/OfficeAddresses.vue
```diff
@@ -221,242 +228,169 @@ import { Actions } from '@/enums'
...
-      return {
-        registeredOffice: {
-          deliveryAddress: deliveryAddressFinal,
-          mailingAddress: mailingAddressFinal
+
+        // records office is required only for BCOMP
+        const recordsOffice = response?.data?.recordsOffice
+        if (recordsOffice) {
+          this.original.recordsOffice = {
+            mailingAddress: { ...recordsOffice.mailingAddress, actions: [] },
```
severinbeauvais, line 322: Yay, spreading and merging into a new object :)

src/interfaces/address-interfaces.ts
severinbeauvais: (ie, same as actions)

(Previously, sometimes addresses had actions or addressType and sometimes they didn't. Now they always do.)

src/interfaces/director-interface.ts
```diff
@@ -1,4 +1,26 @@
...
+  cessationDateTemp?: string
+  isFeeApplied?: boolean
+  isDirectorActionable?: boolean
+  actions: Array<Actions>
+}
+
+export const EmptyDirector: DirectorIF = {
+  id: null,
+  officer: { ...EmptyOfficer },
+  deliveryAddress: { ...EmptyAddress },
```
severinbeauvais, line 20: I'm using spread/merge here to handle "deep copy".
(Ie, In cause anyone who uses "EmptyDirector" does a shallow copy, and here I handle the sub-properties being copied. So the result is an all-new object, no object reference copies.)

src/mixins/filing-mixin.ts
```diff
@@ -58,7 +58,7 @@ export default class FilingMixin extends Vue {
     priority: boolean | undefined,
     waiveFees: boolean | undefined
   ): void {
-    let myFilingData: Array<FilingDataIF> = this.filingData
+    let myFilingData: Array<FilingDataIF> = [...this.filingData]
```
severinbeauvais, line 61: I noticed that, sometimes, the Fee Summary did not update itself even though the filing data was changed. I think this because we restructured filing data in the template in App.vue and it wasn't always reactive (I don't know why). THIS code change makes the new filing data a new object (and I removed the restructuring in App.vue) and now the Fee Summary always updates.

src/store/getters.ts
```diff
@@ -68,7 +68,9 @@ export default {
     return state.currentFilingStatus
   },
 
-  // get last Change of Directors filing from list of past filings
+  /** Returns date of last Change of Directors filing from list of past filings. */
+  // FUTURE: this will break if we only retrieve a page of filings
```
severinbeauvais, line 72: Re: Thor has been talking about changing the filings API (that we use for the history list) to be more efficient, including supporting paging.

src/store/getters.ts
```diff
@@ -68,7 +68,9 @@ export default {
...
+  // FUTURE: the API should be giving us this date
```
severinbeauvais, line 73: Yes, I think the API should be giving the UI its "parameters". Two reasons:
1. there should be minimal business logic in the UI,
2. the API needs to calculate those parameters anyway to verify any data the UI posts to it.

src/views/Correction.vue
```diff
@@ -350,20 +350,26 @@ export default {
...
-      if (this.filingId > 0) {
-        await this.fetchDraftFiling()
-      }
-      this.dataLoaded = true
     }
   },
 
   /** Called when component is mounted. */
-  mounted (): void {
+  async mounted (): Promise<void> {
```
severinbeauvais, line 357: I split up created/mounted here same as everywhere else.

The reason is that accessing any children by `$ref` (which is done in the other filings) doesn't work in "created" because the children are not created yet. But it works fine in "mounted".

src/views/Correction.vue
```diff
@@ -408,60 +414,49 @@ export default {
     async fetchDraftFiling (): Promise<void> {
       const url = `businesses/${this.entityIncNo}/filings/${this.filingId}`
       await axios.get(url).then(res => {
-        if (res && res.data) {
-          const filing: any = res.data.filing
-          try {
```
severinbeauvais, line 413: The "if" is redundant as checking the filing below will throw anyway.

The "try" is redundant because THIS code is already in the "then function" for the axios call, so any throws in here will be caught by the "catch function" below.

src/views/AnnualReport.vue
```diff
@@ -43,12 +43,24 @@
...
+          <v-progress-circular color="primary" :size="50" indeterminate></v-progress-circular>
+          <div class="loading-msg">{{loadingMessage}}</div>
+        </div>
       </div>
-    </div>
+    </v-fade-transition>
+
+    <!-- Alternate Loading Spinner -->
+    <v-fade-transition>
+      <div class="loading-container grayed-out" v-show="isFetching">
```
severinbeauvais, line 57: This is the same grayed out spinner I used when we create a draft IA correction. I'm using it here when the user changes the As Of Date as that requires a new fetch.
(Scott/Tracey may come up with a better spinner style, and if they do then we can fix all the places it's used. But as of now, they're all the same -- consistent -- and not terrible :) )

src/views/AnnualReport.vue
```diff
@@ -96,13 +108,12 @@
...
-                    :registeredAddress.sync="registeredAddress"
-                    :recordsAddress.sync="recordsAddress"
-                    :asOfDate="asOfDate"
+                    ref="officeAddressesComponent"
+                    :addresses.sync="updatedAddresses"
                     :componentEnabled="allowChange('coa')"
+                    @original="originalAddresses=$event"
                     @modified="officeModifiedEventHandler($event)"
-                    @valid="addressFormValid = $event"
+                    @valid="addressesFormValid=$event"
```
severinbeauvais, line 116: Note the new interface (API) between the parent and child.
In addition, the parent makes a call to one of the child's methods to load data. I don't like this but it's the only way to make the fetch async.
A better architecture would be for the parent to do the fetch (or put it in a common mixin/service). That will be another day.

src/views/AnnualReport.vue
```diff
@@ -115,14 +126,15 @@
                     <p v-else>This is your list of directors active as of {{asOfDate}}, including
                       directors that were ceased at a later date.</p>
                   </header>
-                  <directors ref="directorsList"
+                  <directors
+                    ref="directorsComponent"
+                    :directors.sync="updatedDirectors"
+                    :componentEnabled="allowChange('cod')"
+                    @original="originalDirectors=$event"
```
severinbeauvais, line 133: Note the new interface here as well.

And, this interface is now used consistently between all filings.

src/views/AnnualReport.vue
```diff
@@ -141,6 +153,19 @@
...
+              <!-- but don't show them -->
+              <div class="d-none">
+                <office-addresses
+                  ref="officeAddressesComponent"
+                  @original="originalAddresses=$event"
+                />
+                <directors
+                  ref="directorsComponent"
+                  @original="originalDirectors=$event"
+                />
```
severinbeauvais, line 166: This is now needed because previously the data was loaded by other events.

src/views/AnnualReport.vue
```diff
@@ -541,177 +589,160 @@ export default {
...
      * Callback method for the "modified" event from OfficeAddress.
-     *
      * @param modified a boolean indicating whether or not the office address(es) have been modified from their
      * original values.
      */
     officeModifiedEventHandler (modified: boolean): void {
-      this.haveChanges = true
-      // when addresses change, update filing data
+      // only record changes once the initial data is loaded
+      this.haveChanges = this.dataLoaded
```
severinbeauvais, line 703: This is to prevent "you have unsaved data" popup when exiting and having made no change.
In other words, this ignores data changes made by loading/resuming a filing.

src/views/AnnualReport.vue
```diff
@@ -541,177 +589,160 @@ export default {
...
     directorsFreeChange (modified: boolean) {
-      this.haveChanges = true
-      // when directors change (free filing), update filing data
+      // only record changes once the initial data is loaded
+      this.haveChanges = this.dataLoaded
+      // when there is a directors free change, update filing data
       // always set Priority flag to false
       // use existing Waive Fees flag
-      this.updateFilingData(modified ? 'add' : 'remove', this.freeFeeCode, false,
+      this.updateFilingData(modified ? 'add' : 'remove', FilingCodes.FREE_DIRECTOR_CHANGE_OT, false,
```
severinbeauvais, line 727: This can never be BCOMP because user cannot change directors in a BCOMP AR.

src/views/AnnualReport.vue
```diff
@@ -840,53 +871,49 @@ export default {
...
+      ) {
         changeOfDirectors = {
           changeOfDirectors: {
-            directors: this.allDirectors
+            directors: this.updatedDirectors
           }
         }
       }
 
-      if (this.hasFilingCode(FilingCodes.ADDRESS_CHANGE_OT) && this.addresses) {
```
severinbeauvais, line 881: "this.addresses" is always true (it could be an empty object)

Anyway, if the user has not made any changes then there is no filing code, so it was redundant (or maybe a safety check no longer needed).

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -82,15 +94,15 @@
 
                 <!-- Director Information -->
                 <section>
-                  <directors ref="directorsList"
+                  <directors
+                    ref="directorsComponent"
+                    :directors.sync="updatedDirectors"
```
severinbeauvais, line 99: Everything in here is similar to AR architecture/functionality/flow.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -68,11 +86,10 @@
...
             <section>
               <office-addresses
-                :addresses.sync="addresses"
-                :registeredAddress.sync="registeredAddress"
-                :recordsAddress.sync="recordsAddress"
+                ref="officeAddressesComponent"
+                :addresses.sync="updatedAddresses"
                 @modified="officeModifiedEventHandler($event)"
-                @valid="officeAddressFormValid = $event"
+                @valid="addressesFormValid=$event"
```
severinbeauvais, line 92: Everything in here is similar to AR architecture/functionality/flow.

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -510,16 +544,88 @@ export default {
   methods: {
     ...mapActions(['setFilingData']),
 
+    async fetchDraftFiling (): Promise<void> {
```
severinbeauvais, line 547: I moved this up from below, both to be consistent with the other filings and because it's used first.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -351,121 +402,88 @@ export default {
...
-                  (this.staffPaymentData.option === StaffPaymentOptions.NO_FEE))
-              }
-            }
-          } catch (err) {
-            // eslint-disable-next-line no-console
-            console.log(`fetchData() error - ${err.message}, filing = ${filing}`)
-            this.resumeErrorDialog = true
-            throw new Error('Invalid change of address')
+        // do not restore COA Date
+        // since user cannot change it from the UI, it will always be "today"
```
severinbeauvais, line 447: OK, this is a bit different from AR and COD.
But anyway, I'm calling it out for a future developer to understand, and in case we eventually allow an "as of" date for COA.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -552,7 +570,8 @@ export default {
           name: FilingTypes.CHANGE_OF_ADDRESS,
           certifiedBy: this.certifiedBy || '',
           email: 'no_one@never.get',
-          date: this.currentDate
+          date: this.currentDate, // NB: API will reassign this date according to its clock
+          effectiveDate: this.convertLocalDateToUTCDateTime(this.coaDate)
```
severinbeauvais, line 574: Effective Date is not used atm, but now this filing is consistent with AR and COD.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -584,33 +603,25 @@ export default {
           legalName: this.entityName
         }
       }
-      if (this.hasFilingCode(this.feeCode) && this.addresses) {
-        if (this.addresses.recordsOffice) {
```
severinbeauvais, line 588: Again, checking "this.addresses" and "this.addresses.recordsOffice" may have be safety checks before de-referencing them below on line 594-595 etc. But now that "this.addresses" is always an object (never null), this is no longer needed.

Anyway, if we have a filing code then addresses must exist.

src/views/AnnualReport.vue
severinbeauvais: Fixed.

src/views/StandaloneOfficeAddressFiling.vue
severinbeauvais: In other words, this code checks to see what data is available and then loads it. Instead, I think this code should know what data it expects and then verify that it gets it (else error).

src/components/common/Directors.vue
severinbeauvais: NB: this is a shallow copy but it's handled in EmptyDirector (see my other comment on this below).

src/components/common/OfficeAddresses.vue
severinbeauvais: NB: for like-to-like comparison we don't need to omit address type, but here it's comparing a delivery and mailing address.

lekshmimallika-aot (approved):
Nice work!

src/components/common/OfficeAddresses.vue
cameron-freshworks: very nice!

src/interfaces/address-interfaces.ts
cameron-freshworks: very good point.

src/views/Correction.vue
```diff
@@ -408,60 +414,49 @@ export default {
...
-          console.log('fetchDraftFiling() error - invalid response =', res)
-          this.resumeErrorDialog = true
+          this.staffPaymentData = {
+            option: StaffPaymentOptions.NONE
+          }
         }
+
+        // load Detail Comment, removing the first line (default comment)
+        const comment: string = filing.correction.comment || ''
+        this.detailComment = comment.split('\n').slice(1).join('\n')
```
cameron-freshworks, line 459: i recognize this!

cameron-freshworks (approved):
Looks good, a lot good work here!

severinbeauvais:
I'm merging in what is there so far. Only some unit tests remain to be updated (which I am doing) and committed.

PR #161, opened 2020-11-12
improved app retry handler

severinbeauvais:
*Issue #:* None

*Description of changes:* When retrying to load app after error (eg, due to KC timeout), force-restart token service (which may force an app reload and subsequent re-login). This will succeed in more situations than just retrying the fetches without a new token.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -676,12 +676,14 @@ export default {
     },
 
     /** Handles Retry click event from dialogs. */
-    onClickRetry (): void {
+    async onClickRetry (): Promise<void> {
       this.dashboardUnavailableDialog = false
       this.businessAuthErrorDialog = false
       this.nameRequestAuthErrorDialog = false
       this.nameRequestInvalidDialog = false
-      this.fetchData()
+      this.tokenService = false
+      await this.startTokenService()
+      await this.fetchData()
     }
   },
 


cameron-freshworks (approved):
Looks good! is this something we should incorporate into the other UI's as well?

severinbeauvais:
> Looks good! is this something we should incorporate into the other UI's as well?

Yes but it's not that easy! Those other UIs don't reload if they fail to start; they redirect to the dashboard instead. But I think they don't handle KC expiry very well and it _should_ be fixed at some point.

severinbeauvais:
> > Looks good! is this something we should incorporate into the other UI's as well?
> 
> Yes but it's not that easy! Those other UIs don't reload if they fail to start; they redirect to the dashboard instead. But I think they don't handle KC expiry very well and it _should_ be fixed at some point.

... such as /bcgov/entity#2494.

severinbeauvais:
I have confirmed that this code change works (whereas Retry didn't work at all before when the KC expired)... but it doesn't work ideally. It turns out the user gets 2 dialogs and app reloads:
- the first one due to a failed API call (eg, due to expired KC)
- the second one due to an invalid KC
... at that point the user is prompted to login again and will resume their dashboard correctly.

I am testing an alternate solution and will PR that ASAP.

cameron-freshworks:
> I have confirmed that this code change works (whereas Retry didn't work at all before when the KC expired)... but it doesn't work ideally. It turns out the user gets 2 dialogs and app reloads:
> 
> * the first one due to a failed API call (eg, due to expired KC)
> * the second one due to an invalid KC
>   ... at that point the user is prompted to login again and will resume their dashboard correctly.
> 
> I am testing an alternate solution and will PR that ASAP.

Sounds good!

PR #163, opened 2020-11-12
fix for empty officer.middleInitial

severinbeauvais:
*Issue #:* /bcgov/entity#5622

*Description of changes:*
- reset empty officer after pushing new director data

### DIFF
--- a/src/components/common/Directors.vue
+++ b/src/components/common/Directors.vue
@@ -526,6 +526,22 @@ import { Actions, FilingStatus } from '@/enums'
 // Interfaces
 import { FormIF, BaseAddressIF, AlertMessageIF } from '@/interfaces'
 
+const emptyOfficer = {
+  firstName: '',
+  lastName: '',
+  middleInitial: ''
+}
+
+const emptyAddress = {
+  streetAddress: '',
+  streetAddressAdditional: '',
+  addressCity: '',
+  addressRegion: '',
+  postalCode: '',
+  addressCountry: '',
+  deliveryInstructions: ''
+}
+
 @Component({
   components: {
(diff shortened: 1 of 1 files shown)


src/components/common/Directors.vue
```diff
@@ -526,6 +526,22 @@ import { Actions, FilingStatus } from '@/enums'
...
+
+const emptyAddress = {
+  streetAddress: '',
+  streetAddressAdditional: '',
+  addressCity: '',
+  addressRegion: '',
+  postalCode: '',
+  addressCountry: '',
+  deliveryInstructions: ''
+}
```
severinbeauvais, line 543: As this PR is against the most recent release branch, I only copied the minimum code here. Another commit in master fully implements this.

src/components/common/Directors.vue
```diff
@@ -926,6 +926,7 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
     }
 
     // set form to initial director data again
+    this.director.officer = { ...emptyOfficer }
```
severinbeauvais, line 929: This is called after adding a new director or other actions. Either way, the working data (director) needs to be re-initialized after the form was reset otherwise it's undefined.

cameron-freshworks (approved):

cameron-freshworks:
Your g2g!

kialj876 (approved):

src/components/common/Directors.vue
severinbeauvais: (Apparently Vuetify does not update its v-model if you don't touch the element.)

PR #164, opened 2020-11-13
improved app retry handler, take 2

severinbeauvais:
*Issue #:* None

*Description of changes:*
When retrying to load app after an auth error, clear the current KC session and hard-reload the page to try again; otherwise try to fetch the data again. The previous solution (#161) typically caused 2 error dialogs and app reloads. This solution is more heavy-handed and causes just 1.


cameron-freshworks (approved):
G2G!
Edit* Just noticed it was draft :) Sorry i was too hasty

PR #166, opened 2020-11-17
updated unit tests

severinbeauvais:
*Issue #:* /bcgov/entity#2461

*Description of changes:*
- fixed Directors unit tests
- fixed AR unit tests
- fixed COD unit tests
- fixed COA unit tests
- fixed Corrections unit tests
- updated some element names/classes/ids
- initialized some properties
- removed some obsolete test config overrides
- updated some test workaround comments
- moved isJestRunning to getters
- ignored some $refs when Jest is running to avoid undefined errors due to shallow mount

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -67,7 +67,7 @@
 
 <script lang="ts">
 // Libraries
-import { mapState, mapActions } from 'vuex'
+import { mapState, mapActions, mapGetters } from 'vuex'
 import axios from '@/axios-auth'
 import KeycloakService from 'sbc-common-components/src/services/keycloak.services'
 import * as Sentry from '@sentry/browser'
@@ -135,6 +135,8 @@ export default {
   },
 
   computed: {
+    ...mapGetters(['isJestRunning']),
+
     ...mapState(['tasks', 'filings', 'entityType', 'entityStatus']),
 
     /** The Auth API string. */
@@ -175,11 +177,6 @@ export default {
     isAuthenticated (): boolean {
       // FUTURE: also check that token isn't expired!
       return Boolean(sessionStorage.getItem(SessionStorageKeys.KeyCloakToken))
-    },
-
-    /** True if Jest is running the code. */
-    isJestRunning (): boolean {
-      return (process.env.JEST_WORKER_ID !== undefined)
     },
 
     /** The About text. */
--- a/src/App.vue
+++ b/src/App.vue
@@ -67,7 +67,7 @@
 
 <script lang="ts">
 // Libraries
-import { mapState, mapActions, mapGetters } from 'vuex'
+import { mapState, mapActions } from 'vuex'
 import axios from '@/axios-auth'
 import KeycloakService from 'sbc-common-components/src/services/keycloak.services'
 import * as Sentry from '@sentry/browser'
@@ -135,8 +135,6 @@ export default {
   },
 
   computed: {
-    ...mapGetters(['isJestRunning']),
-
     ...mapState(['tasks', 'filings', 'entityType', 'entityStatus']),
 
     /** The Auth API string. */
--- a/src/components/common/Directors.vue
+++ b/src/components/common/Directors.vue
@@ -52,7 +52,7 @@
               <label class="appoint-header">Appoint New Director</label>
               <div class="meta-container__inner">
                 <v-form ref="newDirectorForm"
-                  class="appoint-form"
+                  name="new-director-form"
                   v-model="directorFormValid"
                   v-on:submit.prevent="addNewDirector"
                   lazy-validation
@@ -185,6 +185,7 @@
           <span v-if="isBComp">Mailing Address</span>
           <span>Appointed/Elected</span>
         </v-subheader>
+
         <!-- START FOR LOOP -->
         <li class="director-list-item"
           v-for="(dir, index) in allDirectors"
@@ -194,7 +195,7 @@
         >
           <div class="meta-container">
             <label>
-              <span>{{dir.officer.firstName}} </span>
+              <span class="director-list__first-name">{{dir.officer.firstName}} </span>
               <span>{{dir.officer.middleInitial}} </span>
               <span>{{dir.officer.lastName}}</span>
               <div class="director-status">
@@ -242,17 +243,20 @@
                   <div class="address">
                     <base-address :address="dir.deliveryAddress" />
                   </div>
+
                   <div class="address same-address" v-if="isBComp">
                     <span v-if="isSame(dir.deliveryAddress, dir.mailingAddress)">
                       Same as Delivery Address
                     </span>
                     <base-address v-else :address="dir.mailingAddress" />
                   </div>
+
                   <div class="director_dates">
                     <div class="director_dates__date">{{ dir.appointmentDate }}</div>
                     <div v-if="dir.cessationDate">Ceased</div>
                     <div class="director_dates__date">{{ dir.cessationDate }}</div>
                   </div>
+
                   <div class="actions" v-show="isActionable(dir)">
                     <!-- Edit menu -->
                     <span v-show="isNew(dir)">
@@ -339,6 +343,7 @@
               <v-expand-transition>
                 <!-- only render the form for the active director -->
                 <v-form ref="editDirectorForm"
+                  name="edit-director-form"
                   v-if="activeIndex === index"
                   v-model="directorFormValid"
                   lazy-validation
@@ -512,10 +517,10 @@
             icon="mdi-information-outline"
             class="white-background icon-blue"
             :id="`director-${dir.id}-alert`"
-            v-if="complianceMsg && index == messageIndex"
+            v-if="complianceMsg && index === messageIndex"
             v-once
           >
-            <div class="complianceSection">
+            <div class="compliance-section">
               <h3>{{ complianceMsg.title }}</h3>
               <p>{{ complianceMsg.msg }}</p>
             </div>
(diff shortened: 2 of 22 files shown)


src/App.vue
```diff
@@ -177,11 +179,6 @@ export default {
       return Boolean(sessionStorage.getItem(SessionStorageKeys.KeyCloakToken))
     },
 
-    /** True if Jest is running the code. */
-    isJestRunning (): boolean {
```
severinbeauvais, line 181: Moved to getters file.
(We really should create a non-store getters file. Hmm.)

src/components/common/Directors.vue
```diff
@@ -737,6 +742,29 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
     return rules
   }
 
+  /** The array of validation rules for director cessation date. */
```
severinbeauvais, line 745: I simply moved this up from below (to keep the validation rules together).

src/views/AnnualReport.vue
```diff
@@ -174,7 +174,6 @@
                 <ar-date />
                 <br>
                 <summary-office-addresses
-                  v-if="originalAddresses"
```
severinbeauvais, line 177: No longer needed as "originalAddress" is now initialized in a way that the next 2 lines can de-reference it (which is what failed previously).

src/views/AnnualReport.vue
```diff
@@ -530,14 +529,18 @@ export default {
       await this.fetchDraftFiling()
       // fetch original office addresses and directors
       // update working data only if it wasn't in the draft
-      await this.$refs.officeAddressesComponent.getOrigAddresses(this.asOfDate, isEmpty(this.updatedAddresses))
-      await this.$refs.directorsComponent.getOrigDirectors(this.asOfDate, isEmpty(this.updatedDirectors))
+      if (!this.isJestRunning) {
+        await this.$refs.officeAddressesComponent.getOrigAddresses(this.asOfDate, isEmpty(this.updatedAddresses))
```
severinbeauvais, line 533: Can't run this in Jest as we shallow-mount this component.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -239,7 +239,7 @@ export default {
 
   data () {
     return {
-      updatedAddresses: {},
+      updatedAddresses: { registeredOffice: {}, recordsOffice: {} },
```
severinbeauvais, line 242: More thorough init.
Probably not needed for prod code since this property gets assigned early on, but makes unit testing a bit nicer.

tests/unit/AGMDate.spec.ts
```diff
@@ -5,19 +5,10 @@ import { mount, Wrapper } from '@vue/test-utils'
 import { getVuexStore } from '@/store'
 import AgmDate from '@/components/AnnualReport/AGMDate.vue'
 
-// NB: test util async issue
-// in some cases, the elements are not updated during the test
-// the work-around is to first initialize the property we are changing
-// suppress update watchers warnings
+// suppress "Avoid mutating a prop directly" warnings
```
severinbeauvais, line 8: More concise message specific to this file.
(And ditto for all other similar changes below.)

tests/unit/AnnualReport.spec.ts
```diff
@@ -1251,6 +1255,7 @@ xdescribe('AnnualReport - Part 3B - Submitting (BCOMP)', () => {
     router.push({ name: 'annual-report', params: { filingId: '0' } }) // new filing id
 
     const wrapper = mount(AnnualReport, {
+      sync: false,
```
severinbeauvais, line 1258: So this little gem helps the unit test flow better and avoid misc issues with loading the DOM or data.
But using this requires some `await flushPromises()` to let things stabilize. `await Vue.nextTick()` doesn't cut it.

tests/unit/Correction.spec.ts
```diff
@@ -93,9 +95,9 @@ xdescribe('Correction - UI', () => {
     wrapper.destroy()
   })
 
-  it('sets filing data properly', () => {
+  it('sets filing data properly', async () => {
     // mock $route
-    const $route = { params: {} }
+    const $route = { params: { filingId: '0', correctedFilingId: '123' } }
```
severinbeauvais, line 100: I have no idea how this worked before (without an id to fetch) !?!?

But anyway, it DIDN'T work after my changes, so I had to fix it :)

tests/unit/Directors.spec.ts
```diff
@@ -34,258 +23,456 @@ async function click (vm: any, id: string) {
   await Vue.nextTick()
 }
 
-function setValue (vm: any, id: string, value: string) {
+async function setValue (vm: any, id: string, value: string) {
```
severinbeauvais, line 26: Yeah, sorry, this file was pretty much a rewrite -- but you know I looked at it closely! 😉

tests/unit/Directors.spec.ts
```diff
@@ -34,258 +23,456 @@ async function click (vm: any, id: string) {
...
+    // check that director is not marked as ceased
+    expect(vm.$el.querySelector('#director-1 .director-status .v-chip')
+      .getAttribute('style')).toContain('display: none;')
+
+    // check that director object does not have the 'CEASED' action
+    expect(vm.allDirectors.filter(el => el.id === 1)[0].actions).not.toContain('ceased')
+  })
+})
+
+describe('Directors as a COOP (no sync)', () => {
```
severinbeauvais, line 339: So this is the weirest thing. Some of the tests required `sync: false` and some didn't, so I broke up the describe block into 2.

tests/unit/Directors.spec.ts
```diff
@@ -365,291 +552,243 @@ xdescribe('Directors as a COOP', () => {
...
+    // verify directors are active (initial state)
+    // ("Cease" button indicates this is an active director)
+    expect(vm.isActive(vm.allDirectors[0])).toBe(true)
+    expect(vm.isActive(vm.allDirectors[1])).toBe(true)
+    expect(vm.isActive(vm.allDirectors[2])).toBe(true)
   })
 
   it('disables buttons/actions when instructed by parent component', async () => {
     // clear enabled prop
-    vm.componentEnabled = false
```
severinbeauvais, line 652: This is another one of those things that I don't know how it worked before. Theoretically you're not able to assign a prop, but ...?
Anyway, I fixed it.

tests/unit/Directors.spec.ts
```diff
@@ -731,166 +873,180 @@ xdescribe('Directors as a BCOMP', () => {
...
+    expect(vm.newDirector.officer.firstName).toBe('  First')
 
     // verify that validation error is displayed
     expect(vm.$el.querySelectorAll('.v-messages')[0].textContent).toContain('Invalid spaces')
   })
 
   it('displays error for invalid First Name - trailing spaces', async () => {
     wrapper.find('#new-director__first-name').setValue('First  ')
-    await flushPromises()
+    await flushPromises() // needed due to sync:false
```
severinbeauvais, line 957: As mentioned in another comment.

tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -1318,10 +1333,10 @@ xdescribe('Standalone Directors Filing - Part 5 - Data', () => {
     expect(payload.filing).toBeDefined()
     expect(payload.filing.changeOfDirectors).toBeDefined()
 
-    let names = payload.filing.changeOfDirectors.directors.map(el => el.officer.firstName)
-    expect(names).toContain('Unchanged')
-    expect(names).toContain('Appointed')
-    expect(names).toContain('Ceased')
+    const firstNames = payload.filing.changeOfDirectors.directors.map(el => el.officer.firstName)
```
severinbeauvais, line 1336: This may seems like a frivolous change, but I stared at the old code for a while before figuring it out. I hope the new name makes more sense and will save someone else a couple minutes.

tests/unit/StandaloneOfficeAddressFiling.spec.ts
```diff
@@ -300,12 +307,13 @@ xdescribe('Standalone Office Address Filing - Part 1 - UI', () => {
...
 })
 
-xdescribe('Standalone Office Address Filing - Part 2A - Resuming with FAS staff payment', () => {
+describe('Standalone Office Address Filing - Part 2A - Resuming with FAS staff payment', () => {
   beforeEach(() => {
     // init store
     store.state.businessId = 'CP0001191'
     store.state.entityIncNo = 'CP0001191'
     store.state.entityName = 'Legal Name - CP0001191'
+    store.state.entityType = 'CP'
```
severinbeauvais, line 316: This omission really messed with me!!!
Sometimes, entity type was set from a previous test, and sometimes it was actually set to 'BEN' (which really screws with the code execution). Basically you can't guarantee what order tests run in, so if you need something set, set it!

src/App.vue
cameron-freshworks: Would a mixin work?

tests/unit/StaffPayment.spec.ts
```diff
@@ -3,11 +3,12 @@ import Vuetify from 'vuetify'
 import { mount } from '@vue/test-utils'
 import { StaffPayment } from '@/components/common'
 
-Vue.use(Vuetify)
-// suppress "avoid mutating a prop directly" warnings
-// https://vue-test-utils.vuejs.org/api/config.html#silent
+// suppress "Avoid mutating a prop directly" warnings
+// ref: https://github.com/vuejs/vue-test-utils/issues/532
 Vue.config.silent = true
```
cameron-freshworks, line 8: Good comments here!

cameron-freshworks (approved):
Easy review, mostly tests! g2g 👍

src/App.vue
severinbeauvais: Yes, good idea.

This has me wondering if we should create "services" modules too (ie, classes that don't depend on Vue).

Next time?

severinbeauvais: Thanks for the feedback. I moved the function to common mixin :)

lekshmimallika-aot (approved):

PR #173, opened 2021-01-06
redirect to Pay URL depending on payment flag

severinbeauvais:
*Issue #:* /bcgov/entity#5591

*Description of changes:*
- updated version to 3.5.2
- added code to fetch user info and update LD
- updated FF code to latest
- changed AR filing to use isPaymentActionRequired flag
- changed COD to use isPaymentActionRequired flag
- changed COA to use isPaymentActionRequired flag
- updated unit tests

### DIFF
--- a/src/components/EntityInfo.vue
+++ b/src/components/EntityInfo.vue
@@ -64,7 +64,7 @@
 
         <div class="right-column text-right align-self-end">
           <template v-if="businessId">
-            <!-- if LTD or ULC, show _new_ menu -->
+            <!-- if LTD or ULC, show new menu -->
             <div v-if="isLtdOrUlc">
               <v-btn small text color="primary" id="company-information-button" @click="viewChangeCompanyInfo()">
                 <v-icon small>mdi-pencil</v-icon>
@@ -98,7 +98,7 @@
               </template>
             </dl>
 
-            <!-- if not LTD or ULC, show _old_ menu -->
+            <!-- if not LTD or ULC, show old menu -->
             <v-menu v-if="!isLtdOrUlc" bottom left offset-y>
               <template v-slot:activator="{ on }">
                 <v-btn id="entity-settings-button" small icon color="primary" v-on="on" class="mt-n4 ml-1">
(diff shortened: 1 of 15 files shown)


src/components/EntityInfo.vue
```diff
@@ -155,6 +156,7 @@ export default class EntityInfo extends Mixins(CommonMixin, EnumMixin) {
   /** True if current company is LTD or ULC. */
   private get isLtdOrUlc (): boolean {
     // return (this.isLtd || this.isUlc)
+    // return getFeatureFlag('alteration-ui-enabled')
```
severinbeauvais, line 159: For future use (probably).

src/mixins/legal-api-mixin.ts
```diff
@@ -36,7 +36,7 @@ export default class LegalApiMixin extends Vue {
    * @param filing the object body of the request
    * @returns a promise to return the filing
    */
-  createCorrection (businessId: string, filing: CorrectionFilingIF): Promise<any> {
+  async createCorrection (businessId: string, filing: CorrectionFilingIF): Promise<any> {
```
severinbeauvais, line 39: I checked, and the places where these 2 methods are called already "awaited" them.
I'm not sure why no code error was displayed to say that were were awaiting something that wasn't async.

src/utils/feature-flags.ts
```diff
@@ -1,67 +1,80 @@
-import { initialize, LDFlagSet } from 'launchdarkly-js-client-sdk'
+import { initialize, LDClient, LDFlagSet, LDOptions, LDUser } from 'launchdarkly-js-client-sdk'
```
severinbeauvais, line 1: This is a rewrite of this file that I did a while back in Edit UI.
The main changes from before are:
- better init (with configured options)
- use shared user key 'anonymous' for unknown users (no more random keys)
- updated way to get FF, which supports live flag updates
- a way to set (identify) user, to support per-user flags

(Yes, this file should be updated in the other projects - see /bcgov/entity#6093.)

src/utils/feature-flags.ts
```diff
@@ -1,67 +1,80 @@
...
+// get rid of "element implicitly has an 'any' type..."
+declare const window: any
+
+/**
+ * Default flag values when LD is not available.
+ * Uses "business-edit" project (per LD client id in config).
+ */
+const defaultFlagSet: LDFlagSet = {
+  'correction-ui-enabled': false,
+  'alteration-ui-enabled': false
```
severinbeauvais, line 12: I have changed this so that the feature is DISABLED if the user cannot connect to LD. This will prevent users from accessing features accidentally.
(The only impact is that you may need to override these when running locally.)

lekshmimallika-aot (approved):

vysakh-menon-aot (approved):

PR #176, opened 2021-01-13
todo list tweaks and prep for further unit tests

severinbeauvais:
*Issue #:* /bcgov/entity#5802

*Description of changes:*
- added check for pay-error class
- combined pending status blocks
- replaced array filter with find
- very minor cleanup
- created new test file for template-first tests (very incomplete)

### DIFF
--- a/tests/unit/TodoList2.spec.ts
+++ b/tests/unit/TodoList2.spec.ts
@@ -1,0 +1,230 @@
+//
+// This file differs from TodoList.spec.ts in that it tests "backwards" from the HTML,
+// instead of working "forwards" from the data.
+//
+
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+// import Vue2Filters from 'vue2-filters'
+import Vuelidate from 'vuelidate'
+import { mount } from '@vue/test-utils'
+import { getVuexStore } from '@/store'
+import TodoList from '@/components/Dashboard/TodoList.vue'
+
+// suppress "Avoid mutating a prop directly" warnings
+// ref: https://github.com/vuejs/vue-test-utils/issues/532
+Vue.config.silent = true
+
+Vue.use(Vuetify)
+// Vue.use(Vue2Filters) // not sure if this is needed?
+Vue.use(Vuelidate)
+
+const vuetify = new Vuetify({})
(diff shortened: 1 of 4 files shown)


tests/unit/TodoList2.spec.ts
```diff
@@ -0,0 +1,230 @@
+//
```
severinbeauvais, line 1: As you can see, I got almost nothing implemented here (design only)... and I'm not even happy with the design.

The Todo List component is brutally complicated due to all possible scenarios.

I would like to reorganize the Todo List but I am not sure how we should reorganize it: big code blocks for each possible state? But then there will be a lot of code/HTML duplication.

Clearly, the (partial) day I allocated to implement tests for all scenarios was not enough. I'm open to ideas on how to proceed.

Unfortunately there is no tangible benefit to our users to refactor this. There is only benefit to us in having easier/less fragile code to work on and potentially fewer bugs. But it's going to be a lot of work.

Thoughts?

severinbeauvais: I could create a new refactor/cleanup ticket but I'm afraid we would never get around to implementing it.

severinbeauvais: ^^^ I'm going to create a new ticket for this anyways. I think it's our best option atm. (Fixing this now is not in scope.)

... As for this PR, do you think we should merge it?

vysakh-menon-aot (approved):

tests/unit/TodoList2.spec.ts
cameron-freshworks: Yes to all of the above ( As in i 100% agree :) ). 

Like you said, best to stick with your scope for your sanity sake and we can tackle this when we're able.

cameron-freshworks (approved):

cameron-freshworks:
Merge whenever your ready!

tests/unit/TodoList2.spec.ts
severinbeauvais: https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/6168

PR #183, opened 2021-02-11
added defensive code to prevent future errors

severinbeauvais:
*Issue #:* /bcgov/entity#6475

*Description of changes:*

Although the subject bug was due to invalid data in the database, I did the following to prevent future like errors:
- added console error if corp type doesn't match legal type
- added defensive checks to Resource Lookup code
- updated unit tests
- updated app patch version

### DIFF
--- a/src/mixins/resource-lookup-mixin.ts
+++ b/src/mixins/resource-lookup-mixin.ts
@@ -22,11 +22,9 @@ export default class ResourceLookupMixin extends Vue {
      * @returns the appropriate message for the certify component for the current filing flow
      */
     certifyText (feeCode: string): string {
-      if (this.configObject && this.configObject.flows) {
-        const flow = this.configObject.flows.find(x => x.feeCode === feeCode)
-        if (flow && flow.certifyText) {
-          return flow.certifyText
-        }
+      const flow = this.configObject?.flows?.find(x => x.feeCode === feeCode)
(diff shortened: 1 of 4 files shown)


src/mixins/resource-lookup-mixin.ts
```diff
@@ -46,37 +44,45 @@ export default class ResourceLookupMixin extends Vue {
...
       // FUTURE: Too much code for this. Can be condensed and made more reusable.
-      if (directors && directors.length) {
-        const configSection = this.configObject.flows.find(x => x.feeCode === filingCode).warnings
+      if (directors?.length) {
+        const warnings = this.configObject?.flows?.find(x => x.feeCode === filingCode)?.warnings
         let errors = []
         // If this entity has a BC Residency requirement for directors, one of the
         // directors specified needs to have both their mailing and delivery address within British Columbia
-        if (configSection.bcResident) {
+        if (warnings?.bcResident) {
```
severinbeauvais, line 52: This line (and the 2 below) was the real culprit: the filing code did not have any warnings, so de-referencing the property caused an exception, causing the COD page to fail to load.
This code no longer fails, even if the db has invalid legal type.

vysakh-menon-aot (approved):

cameron-freshworks (approved):
Cool, this answered my question about the `corpType`.

PR #185, opened 2021-02-25
updated alterations in Filing History List (+ interface cleanup)

severinbeauvais:
*Issue #:* /bcgov/entity#5110

*Description of changes:*
- imported latest shared components
- renamed "notice of alteration" -> "alteration filing"
- renamed isNoa -> isAlterationFiling
- renamed AlterationFiling -> CompletedAlteration
- made common Future Effective Pending component
- made common Future Effective component
- updated some detail block classnames
- updated some Filing History List method names
- updated alteration filing loading
- renamed type NOTICE_OF_ALTERATION -> ALTERATION
- un-renamed TodoList and FilingHistoryList methods
- updated alteration data
- renamed type TRANSITION_APPLICATION -> TRANSITION
- added header for future effective alteration pending
- added header for future effective alteration
- changed "arrangement" text per new requirements
- changed text in Future Effective component per new requirements
- changed text in Future Effective Pending  per new requirementscomponent per new requirements
- changed text in Pending Filing component per new requirements
- updated/added a bunch of unit tests
- updated app version
- updated local status proxy URL


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.8.0",
+  "version": "3.8.1",
```
cameron-freshworks, line 3: I'll leave the version in my PR, since you're already bumping it up here.

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
```diff
@@ -0,0 +1,88 @@
...
+  components: { ContactInfo }
+})
+export default class FutureEffective extends Vue {
+  readonly entityName!: string
+
+  /** The subject filing. */
+  @Prop({ required: true }) private filing: HistoryItemIF
+
+  /** Data for the subject filing. */
+  private get _ (): any {
```
cameron-freshworks, line 41: This is interesting. I'm sure there is a good reason, but i'm just curious why name a getter `_` of type `any`.

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
```diff
@@ -0,0 +1,88 @@
...
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop } from 'vue-property-decorator'
+import { mapState } from 'vuex'
+import { ContactInfo } from '@/components/common'
+import { HistoryItemIF } from '@/interfaces'
+
+@Component({
+  computed: { ...mapState(['entityName']) },
```
cameron-freshworks, line 31: I've seen a few instances where we've mapped this state to the component, then declared on the component as `readonly`. 
Would this be better as a getter? 
I believe if it is a Getter and it's ONLY used in the template, you don't need to declare it on the component, which is just a bonus if you don't need to use it within the `<script>`

cameron-freshworks (commented):
Looks good, just had a few questions.

package.json
severinbeauvais: Or, the first one of use can go to .1 and the second one to .2
I will adjust my PR to suit whatever you decide in yours.

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
severinbeauvais: 1. I couldn't come up with a good name for this localized data block (`data` works for a while but crashes the tests).
2. I didn't think it was worth an interface to define this data block. It's only used here and it's a pretty small component.

What do you think? Rename? Is there a better data type? Is it OK as is?

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
severinbeauvais: 1. In this component, it is used in the script.
2. In other components it is used in the template only, but still think it's good practice to declare its reference (the readonly bit).
3. TBH, it would be even better to use a store getter for this property, but that's not the pattern in this UI.

Can I leave this?

package.json
cameron-freshworks: Ok, i'll bump as well 👍

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
cameron-freshworks: O ya, fine by me. I was just curious as to why.

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
cameron-freshworks: Of course!

cameron-freshworks (approved):

PR #186, opened 2021-02-25
Filings UI: Alteration Handling

cameron-freshworks:
*Issue #:* /bcgov/entity#6352

*Description of changes:*
- Allow regular users to resume draft alteration filings.
- Disable launching the alteration UI when there is a draft filing.
- Resume alteration filings using the filing ID `/alteration?alteration-id=filingId`.

- Payment can resumed
- Payment can be cancelled
- Alteration filing can be cancelled


src/components/EntityInfo.vue
```diff
@@ -286,9 +287,10 @@ export default class EntityInfo extends Mixins(CommonMixin, EnumMixin) {
 #entity-info {
   background: $BCgovInputBG;
 
-  &.hover {
-    background: $app-bg-lt-blue;
-  }
+  // TODO: Curious about this one. Is this a design requirement?
```
severinbeauvais, line 290: Yes it is.

src/enums/filingNames.ts
```diff
@@ -6,7 +6,7 @@ export enum FilingNames {
   CORRECTION = 'Correction',
   INCORPORATION_APPLICATION = 'Incorporation Application',
   NAME_REQUEST = 'Name Request',
-  NOTICE_OF_ALTERATION = 'Alteration Notice',
+  NOTICE_OF_ALTERATION = 'Changes to Company Information',
```
severinbeauvais, line 9: This will affect the filing name in the History List. Did you see a design for this? (I didn't.)

src/store/actions.ts
```diff
@@ -79,5 +79,8 @@ export default {
   },
   setFilingData ({ commit }, filingData: Array<FilingDataIF>) {
     commit('filingData', filingData)
+  },
+  setHasBlockertask ({ commit }, hasBlockerTask: boolean) {
+    commit('hasBlockerTask', hasBlockerTask)
```
severinbeauvais, line 84: Search for `has-blocker-task`. This global may conflict with that.

src/store/getters.ts
```diff
@@ -23,7 +28,7 @@ export default {
 
   /** Is True if entity is a Limited Company. */
   isLtd (state): boolean {
-    return (state.entityType === CorpTypeCd.LIMITED_CO)
+    return (state.entityType === CorpTypeCd.BC_COMPANY)
```
severinbeauvais, line 31: This is a change from LLC to BC. Intended?

severinbeauvais: How about renaming the getter?

src/components/EntityInfo.vue
severinbeauvais: It's to show users what company information would change. (Tracey's idea.)

src/enums/filingNames.ts
severinbeauvais: According to my story, "Changes to Company Information" is the subtitle.
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/4575

src/components/EntityInfo.vue
cameron-freshworks: kk no prob, it does highlight the entire entity info area when you hover over the btn, so it seemed odd. I will put back in tho!

src/enums/filingNames.ts
cameron-freshworks: hmm, this is al i have to work off! It does show the same title in the history list as well 🤔 

https://preview.uxpin.com/62ac91b2c4745163586b730ec154e8ea024bd8ed#/pages/133498721/simulate/no-panels

src/store/getters.ts
cameron-freshworks: An LTD is 'BC' in Lear, so i think this is correct. 
In our corpTypes 'BC' was labeled as BC_COMPANY though, is that still accurate?

src/store/actions.ts
cameron-freshworks: indeed! That value is emit from the TODO list up to the dashboard, where it was used locally. I set it to store opposed to a local property, so we can access it in the Entity Info as well, which was mounted at the `app` level.

src/enums/filingNames.ts
severinbeauvais: 🤷‍♂️ 
I have asked for clarity on this but still waiting for an answer.

I like your uxpin but it directly conflicts with my story.

cameron-freshworks: oh no! hahah.. classic :) 

No rush on this, this PR can hang out until we get some answers.

src/store/getters.ts
severinbeauvais: Maybe check corp types with a BA.

Please rename getter (eg, isBcCompany).

src/store/actions.ts
severinbeauvais: So they conflict or they play well together? Can the event be replaced by the store property instead?

src/enums/filingNames.ts
severinbeauvais: Mine too.

Should we link in Thor's vid? lol

severinbeauvais: My ticket has a [new uxpin](https://preview.uxpin.com/fac96ab3f3769b5940b0444784eb73a2b2606b0e#/pages/136199952/simulate/no-panels).

The alteration filing title in the Filings History List is different than the alteration filing in the Todo List, so I'd argue you should not change the filing name as above.

I'd also argue that the Todo List should display simply "Alteration Notice" (or "Alteration:") since you don't know whether the filed alteration will be for (1) a legal type change, or (2) some other change to company information (*). But if not then I still think you need to set your title somewhere and not change the filing name.

(*) The Filing History List identifies only these 2 types of changes in an alteration, so I think it makes sense that the Todo List should be similar.

cameron-freshworks: I'm not opposed to leaving it. I will revert it back to `Alteration Notice` and we can push it through to UX. 
When they see it all together we can update if they prefer :)

src/store/actions.ts
cameron-freshworks: No conflicts, the emit value is from child to parent, where it is then set to store. 
I'll have a look if there is any refactoring, but i think it's preferable keeping the child components isolated of the store (as best we can :)).

src/store/getters.ts
cameron-freshworks: Kk, will update!

src/enums/filingNames.ts
severinbeauvais: 1. I think you should not change the filing name just to suit one display item. It impacts other things.
2. Check with BA/UX on what to call the Todo List item. I do not have authority on this.

src/store/actions.ts
severinbeauvais: Well, ok, but it feels like a code smell. Why do it 2 different ways -- can't the child just set the store value? Also, if it's only between a child and parent, it may be cleaner to use an `@event` to pass up the event instead of a global emitter. I think global emitters should be reserved to "whole app" events, whereas the parent-child interface (props/events) should be used otherwise.

But if you've already thought about this and think it's the correct way forward, go for it and ignore this comment :)

src/components/EntityInfo.vue
severinbeauvais: I thought you were going to leave this in.

src/store/actions.ts
cameron-freshworks: `this.$emit` is child/parent interface no?
 Isn't `this.$root.$emit` global? If so, that is not what is used here.

src/components/EntityInfo.vue
cameron-freshworks: Added back in

src/store/actions.ts
cameron-freshworks: There is nothing wrong with a child setting the store value, but the way we've been evolving is trying to isolate the children to make them more reusable (thinking shared components here). Obviously this component is far beyond that point, but it doesn't hurt to try and follow that practice.

severinbeauvais: 1. Oops, I thought this was a global emitter. The old syntax confused me. You're right.
2. Agreed, keep the parent-child API as clean as possible.
3. I'm resolving this conversation. Thanks for humouring me :)

cameron-freshworks: No prob! It's always worth the conversation :)

severinbeauvais (approved):
GTG

PR #191, opened 2021-04-01
6650 update alterations ledger to include court of order fields

argush3:
Issue #: /bcgov/entity#6650

Description of changes:

- Updated package number to 3.8.6
- Updated package.json to use v1.0.17 of bcrs-shared-components/interfaces.  Contains court order related interface updates
- Updated FilingHistoryList.vue component to use court order interface and to properly populate data from BE into history item object
- Created EffectOfOrderTypes enum and added helper methods to enum mixin
- Updated proxy config in vue.config.js as was having issues grabbing keycloak.json locally

Note: existing test already tested the updated fields so no updates were made to tests.

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -375,7 +375,7 @@ import { AddCommentDialog, DownloadErrorDialog, LoadCorrectionDialog } from '@/c
 
 // Enums and Interfaces
 import { FilingStatus, FilingTypes, Routes } from '@/enums'
-import { FilingIF, HistoryItemIF } from '@/interfaces'
+import { CourtOrderIF, FilingIF, HistoryItemIF } from '@/interfaces'
 
 // Mixins
 import { DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
@@ -647,8 +647,11 @@ export default {
       if (header && alteration && business) {
         const newLegalType = this.getCorpTypeDescription(alteration.business?.legalType)
         const oldLegalType = this.getCorpTypeDescription(business.legalType)
-        const courtOrderNumber = '' // FUTURE
-        const isArrangement = false // FUTURE
+        const courtOrder = alteration.courtOrder
+        const courtOrderNumber = courtOrder?.fileNumber ? courtOrder.fileNumber : ''
+
+        // isArrangement is true when effect of order value returned by BE is 'planOfArrangement'
+        const isArrangement = this.isPlanOfArrangement(courtOrder)
 
         let title = 'Alteration'
         if (newLegalType !== oldLegalType) {
@@ -794,6 +797,15 @@ export default {
         }
       }
       return false
+    },
+
+    /** Whether this filing(alteration) is pursuant to a plan of arrangement  */
+    isPlanOfArrangement (courtOrder: CourtOrderIF): boolean {
+      if (!courtOrder || !courtOrder.effectOfOrder) {
+        return false
+      }
+
+      return this.isEffectOfOrderPlanOfArrangement(courtOrder)
     },
 
     loadCorrection (filing: FilingIF) {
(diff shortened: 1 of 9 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -647,8 +647,11 @@ export default {
       if (header && alteration && business) {
         const newLegalType = this.getCorpTypeDescription(alteration.business?.legalType)
         const oldLegalType = this.getCorpTypeDescription(business.legalType)
-        const courtOrderNumber = '' // FUTURE
-        const isArrangement = false // FUTURE
+        const courtOrder = alteration?.courtOrder
```
severinbeauvais, line 650: Conditional chaining is always a red flag for me when there is no fallback handling. I ask myself, what does it mean when the object is null? What happens next?

In this case, I see that you check that `alteration` is truthy on line 647, so actually this conditional chaining is not needed and this code is safe.

Line 651 below is a great example of conditional chaining with a fallback 👍

src/enums/effectOfOrderTypes.ts
```diff
@@ -0,0 +1,3 @@
+export enum EffectOfOrderTypes {
+  PLAN_OF_ARRANGEMENT = 'planOfArrangement'
```
severinbeauvais, line 2: Yup, good use of an enum (or could be a constant, but this is fine), in order to abstract away whatever the BE sends as a "type" 👍

vue.config.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
     proxy: {
       // this is needed to prevent a CORS error when running locally
       '/local-keycloak-config-url/*': {
-        target: 'https://business-filings-dev.pathfinder.gov.bc.ca/businesses/edit/config/kc/',
+        target: 'https://business-filings-dev.apps.silver.devops.gov.bc.ca/business/config/kc/',
```
severinbeauvais, line 41: This is the "internal" URL for Filings UI. If possible, we should specify the external URL here.
Can you please check if this works instead?
`https://dev.bcregistry.ca/business/config/kc/keycloak.json`

severinbeauvais (approved):
Nothing major.
Maybe a small fix to keycloak URL, otherwise GTG.

vue.config.js
argush3: sure, will test this out and update if it works

src/mixins/enum-mixin.ts
```diff
@@ -110,6 +111,15 @@ export default class EnumMixin extends Vue {
     return (item.paymentMethod === PaymentMethod.ONLINE_BANKING)
   }
 
+  //
+  // Effect of Order helpers
+  //
+
+  /** Returns True if effect of order is Plan of Arrangement. */
+  isEffectOfOrderPlanOfArrangement (item: any): boolean {
```
cameron-freshworks, line 119: Would `CourtOrderIF` work here opposed to `any`?

cameron-freshworks (approved):
Looks good, just one comment about a type!

src/mixins/enum-mixin.ts
severinbeauvais: (Actually I don't mind `any` here as it means you can pass any object type to check if it's a match. After all this is in a general mixin. Thoughts?)

cameron-freshworks: But it references a property specific to the `CourtOrder` object, `effectOfOrder`? 
I agree with the mixin being general but this feels like a pretty specific check vs a specific variable. Perhaps this should be a getter since it is specific?

severinbeauvais: Good points.

I don't have a preference whether this is a "enum mixin" or a "getter" (a la "isBcomp"). If we leave it leave it here then it's a bit overkill (but extensible), whereas being a getter seems over-optimized. 🤷

cameron-freshworks: haha totally, not a hill i'd like to die on 😆 
This is g2g!

vue.config.js
argush3: Updated and confirmed https://dev.bcregistry.ca/business/config/kc/ works.

src/components/Dashboard/FilingHistoryList.vue
argush3: updated line 650 to stop using conditional chaining

PR #192, opened 2021-04-12
Cache fixes for issues encountered during prod release

saravanpa-aot:
*Issue #:* /bcgov/entity###

*Description of changes:*

1) vue.config.js changes is for service worker to ignore json files from pre caching. Without this , changing the configuration.json wont have any effect unless we give a new build along with it.
2) fetch-config.js is to avoid the disk cache. Ideally this can be done at nginx level as well. 

This code deployed to prod yesterday since it was blocking the release and testing. So could you please tag it/do a release ;not sure of the conventions for the project . @severinbeauvais 

### DIFF
--- a/src/utils/fetch-config.ts
+++ b/src/utils/fetch-config.ts
@@ -20,7 +20,8 @@ export async function fetchConfig (): Promise<void> {
   // fetch config from API
   // eg, http://localhost:8080/business/config/configuration.json
   // eg, https://dev.bcregistry.ca/business/config/configuration.json
-  const url = `${windowLocationOrigin}${processEnvBaseUrl}config/configuration.json`
+  let random = new Date().toISOString().substring(0, 10)
+  const url = `${windowLocationOrigin}${processEnvBaseUrl}config/configuration.json?${random}`
   const headers = {
     'Accept': 'application/json',
     'ResponseType': 'application/json',
--- a/vue.config.js
+++ b/vue.config.js
@@ -31,7 +31,9 @@ module.exports = {
   pwa: {
     workboxPluginMode: 'InjectManifest',
     workboxOptions: {
-      swSrc: 'src/service-worker.js'
+      swSrc: 'src/service-worker.js',
+      // skip precaching json files such as configs
+      exclude: [/\.json$/]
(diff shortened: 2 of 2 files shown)


src/utils/fetch-config.ts
```diff
@@ -20,7 +20,8 @@ export async function fetchConfig (): Promise<void> {
   // fetch config from API
   // eg, http://localhost:8080/business/config/configuration.json
   // eg, https://dev.bcregistry.ca/business/config/configuration.json
-  const url = `${windowLocationOrigin}${processEnvBaseUrl}config/configuration.json`
+  let random = new Date().toISOString().substring(0, 10)
+  const url = `${windowLocationOrigin}${processEnvBaseUrl}config/configuration.json?${random}`
```
severinbeauvais, line 24: Why do you need both this change (which should bust the cache, I think), and the change in `vue.config.js` (which should skip pre-caching) ? Why isn't one of these changes enough ?

saravanpa-aot: the pre-caching fixes the service worker cache.But the file gets cached in the disk. So it was getting served still from the cache.
We thought closing the browser and logout will invalidate the cache , but it didnt. Hence have to go for the fix.

saravanpa-aot: The ideal fix would be to serve this file from nginx with no-cache header .

severinbeauvais: ^^^ Yes, I agree.

I'm going to put this on my back-burner until it is decided how to proceed with this (re: Thor's planning).

saravanpa-aot: I am good with that..do you want to  tag/add it to the release ? Also incase if the planning/implementation doesn't happen before next release , it will be good to have this code in the next release as well.. But all leaving all this up to you.  :-)

severinbeauvais: @thorwolpert Please let us know your plan for this moving forward. Thanks!

severinbeauvais:
I am closing this PR and will create a new one (against branch "main") with these changes ASAP.

PR #195, opened 2021-04-20
changes for showing in good standing

jordiwes:
*Issue #:* /bcgov/entity#6923

*Description of changes:*

Show text & disable resuming alteration if not in good standing

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -484,7 +484,13 @@ export default {
 
       this.setEntityName(business.legalName)
       this.setEntityType(business.legalType)
-      this.setEntityStatus(business.status)
+      if ('goodStanding' in business) {
+        if (business.goodStanding === true) {
+          this.setEntityStatus(EntityStatus.GOOD_STANDING)
+        } else {
+          this.setEntityStatus(EntityStatus.NOT_IN_COMPLIANCE)
+        }
+      }
       this.setEntityBusinessNo(business.taxId)
       this.setEntityIncNo(business.identifier)
       this.setEntityFoundingDate(business.foundingDate) // datetime
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -543,7 +543,7 @@ export default {
   },
 
   computed: {
-    ...mapGetters(['isBComp', 'isCoop', 'isRoleStaff', 'currentYear']),
+    ...mapGetters(['isBComp', 'isCoop', 'isRoleStaff', 'currentYear', 'isInGoodStanding']),
 
     ...mapState(['tasks', 'entityIncNo', 'entityName', 'nameRequest', 'currentDate',
       'lastAnnualReportDate', 'entityStatus']),
@@ -1182,10 +1182,6 @@ export default {
 
     isPayError (task: TaskItemIF): boolean {
       return !!task.payErrorObj
-    },
-
-    isInGoodStanding (): boolean {
-      return this.entityStatus === EntityStatus.GOOD_STANDING
     },
 
     priorityAlterationTitle (priority: boolean): string {
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -543,7 +543,7 @@ export default {
   },
 
   computed: {
-    ...mapGetters(['isBComp', 'isCoop', 'isRoleStaff', 'currentYear', 'isInGoodStanding']),
+    ...mapGetters(['isBComp', 'isCoop', 'isRoleStaff', 'currentYear']),
 
     ...mapState(['tasks', 'entityIncNo', 'entityName', 'nameRequest', 'currentDate',
       'lastAnnualReportDate', 'entityStatus']),
@@ -601,6 +601,7 @@ export default {
 
   methods: {
     ...mapActions(['setARFilingYear', 'setArMinDate', 'setArMaxDate', 'setNextARDate', 'setCurrentFilingStatus']),
+    ...mapGetters(['isInGoodStanding']),
 
     async loadData () {
       this.taskItems = []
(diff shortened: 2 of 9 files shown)


src/App.vue
```diff
@@ -484,7 +484,13 @@ export default {
 
       this.setEntityName(business.legalName)
       this.setEntityType(business.legalType)
-      this.setEntityStatus(business.status)
+      if ('goodStanding' in business) {
```
severinbeauvais, line 487: Out of curiousity, what is the value when they are not in compliance?

src/components/EntityInfo.vue
```diff
@@ -24,17 +24,6 @@
               <span>{{ entityName || getCorpTypeNumberedDescription(entityType)}}</span>
             </div>
 
-            <!-- Entity Status -->
-            <v-chip v-if="isGoodStanding" class="blue" id="entity-status" small label text-color="white">
```
severinbeauvais, line 28: Can you please delete the now-obsolete getters?

severinbeauvais:
Please update the app version.

src/App.vue
jordiwes: goodStanding coming from api can either be true or false

src/components/EntityInfo.vue
jordiwes: Done

jordiwes:
changed the app version

src/App.vue
severinbeauvais: Why not simplify this check to "if true then ... else if false then ..." ?

src/components/Dashboard/TodoList.vue
```diff
@@ -79,7 +79,23 @@
...
+                >
+                  <p class="red--text">
+                    <v-icon color="red" left>mdi-information-outline</v-icon>
+                    This business is not in good standing.
+                  </p>
+
+                  Before you can alter your business it must be in good standing with the Business Registry.
+                  There may be several reasons a business is not be in good standing, but the most
+                  common reason is an overdue annual report.<br>
+                  To resolve this issue, you MUST contact registries staff:
```
severinbeauvais, line 93: Check capitalization of "registries staff" elsewhere in this app. I think it needs to be title cased.

src/components/Dashboard/TodoList.vue
```diff
@@ -711,6 +728,10 @@ export default {
 
     loadAlteration (task) {
       const filing: FilingIF = task.task.filing
+      if (!this.isInGoodStanding() && this.isStatusDraft(filing.header)) {
+        task.enabled = false
```
severinbeauvais, line 732: This will leave Address Change and Directors Change enabled. Should they be disabled too? If so, search for "has-blocker-task" to see if you reuse that.

jordiwes: Well, the task is "Prevent user from starting alteration if "Not In Good Standing"" ...

src/components/Dashboard/TodoList.vue
jordiwes: changed to standard in app

src/components/Dashboard/TodoList.vue
severinbeauvais: Can you please confirm that COA and COD are allowed?

severinbeauvais (approved):
Looks good.
If COA/COD also need to be blocked then it can be another ticket.

severinbeauvais:
This does not disable the "change company info" button in the Entity Info header, which is how existing companies start an alteration (other than changing business type). Should it?

src/App.vue
jordiwes: Trying to mitigate scenario where goodStanding not passed?

severinbeauvais: If `goodStanding` is not passed then the value of `business.goodStanding` is undefined (not True or False), so you could simplify this.

I only picked on this because I did a double-take on `'goodstanding'`, thinking it needed a const or enum. Then I realized it was a property.

There's nothing wrong with your code. I'm closing this conversation.

src/components/EntityInfo.vue
```diff
@@ -223,19 +212,9 @@ export default class EntityInfo extends Mixins(CommonMixin, EnumMixin) {
...
-  /** True if the entity has the subject status. */
-  private get isPendingDissolution (): boolean {
-    return (this.entityStatus === EntityStatus.PENDING_DISSOLUTION)
-  }
-
-  /** True if the entity has the subject status. */
-  private get isNotInCompliance (): boolean {
-    return (this.entityStatus === EntityStatus.NOT_IN_COMPLIANCE)
+  /** Returns whether business is in good standing */
+  private get isInGoodStanding (): boolean {
```
severinbeauvais, line 216: This getter is now duplicated (see also App.vue), so I recommend this be moved to the store getters file.

severinbeauvais:
1. Not all questions have been answered/actioned.
2. There is a conflicting file.
3. As Steph has left the immediate team, Cam or I (or another developer) probably needs to pick this up.

src/components/Dashboard/TodoList.vue
jordiwes: Sienna says "If the client has an alteration draft, and are not in good standing, not only can they not file any other filings because of the draft, but also can't file the alteration draft.
So to answer your question, correct, they cannot file a director or address change. But not because of the "not in good standing" status, it's because they have a draft open."

severinbeauvais: Have you verified that a draft alteration task blocks COA and COD? If it's OK then this conversation can be resolved.

jordiwes: verified

severinbeauvais (approved):
Looks good. Just need to fix that failing unit test and then we can merge this.

cameron-freshworks (approved):

PR #196, opened 2021-04-21
Updated sbc-common-components dependency + updated caching config

severinbeauvais:
*Issue #:* /bcgov/entity#7263

*Description of changes:*
- added nginx config to disable index.html caching
- updated app version
- updated sbc-common-components dependency
- added workbox config to exclude precaching JSON files

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -26,13 +26,19 @@ http {
 
    location /business {
      alias /app;
-     try_files $uri /index.html;
+     try_files $uri @index;
      gzip on;
      gzip_vary on;
      gzip_min_length 10240;
      gzip_proxied any;
      gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/xml;
      gzip_disable "MSIE [1-6]\.";
+   }
+
+   location @index {
+     add_header Cache-Control no-cache;
+     expires 0;
+     try_files /index.html =404;
    }
 
    # For status of ngnix service, OpenShift is configured to call this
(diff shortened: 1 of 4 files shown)


nginx.conf
```diff
@@ -35,6 +35,12 @@ http {
      gzip_disable "MSIE [1-6]\.";
    }
 
+   location @index {
+     add_header Cache-Control no-cache;
+     expires 0;
+     try_files /index.html =404;
+   }
```
saravanpa-aot, line 42: hm..interesting..why not cache index.html?

cameron-freshworks (approved):

nginx.conf
severinbeauvais: Because it contains references to specific build artifacts, eg:
```
<link href=/business/js/app.54fd7d80.js rel=preload as=script>
<link href=/business/js/chunk-vendors.626d7e3c.js rel=preload as=script>
<link href=/business/css/chunk-vendors.a6a3f87c.css rel=stylesheet>
<link href=/business/css/app.11f39ad1.css rel=stylesheet>
```

By not caching index.html, we force the client to always get the latest index.html, which might refer to newer artifacts,

saravanpa-aot: hm.. ideally service worker should fetch the new index.html since the new build is in place..  service worker will know its new application .. 
But if you ever faced this issue even with service worker working properly , its safe to have this change , I agree

severinbeauvais: Good point. We may have to review the SW eventually (I think it's currently not working correctly as it is displaying the "update available" alert AFTER having loaded the new code, haha).

Anyway, can you please resolve this conversation?

PR #197, opened 2021-04-28
Scrollable sections and UI updates

cameron-freshworks:
*Issue #:* /bcgov/entity#7196

*Description of changes:*

* Allows scrolling on Filings History List and Directors List.

* Added `Effective as of Date` to filings history list when present.

* Minor UI/style adjustments.

* Updates unit testing.

*Note:*

* Hide white space changes to easier review. Adding a wrapper creates a large diff where it actually pretty small :) 

* Does NOT contain any data restructure as noted in the ticket, as the API work is still pending.

### DIFF
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -17,14 +17,14 @@
 @font-face {
   font-family: 'BCSans';
   font-style: normal;
-  font-weight: 700; 
+  font-weight: 700;
   src: url('../fonts/BCSans/BCSans-Bold.woff2') format('woff2'), /* Modern Browsers */
        url('../fonts/BCSans/BCSans-Bold.woff') format('woff'); /* Optimized for very modern browsers */
 }
 @font-face {
   font-family: 'BCSans';
   font-style: italic;
-  font-weight: 700; 
+  font-weight: 700;
   src: url('../fonts/BCSans/BCSans-BoldItalic.woff2') format('woff2'), /* Modern Browsers */
        url('../fonts/BCSans/BCSans-BoldItalic.woff') format('woff'); /* Optimized for very modern browsers */
 }
@@ -66,6 +66,10 @@ h1 {
 
 p {
   font-weight: 400;
+}
+
+.app-action {
+  color: $app-blue !important;
 }
 
 .v-application {
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -19,328 +19,330 @@
       attach="#filing-history-list"
     />
 
-    <v-expansion-panels v-if="historyItems.length > 0" v-model="panel">
-      <v-expansion-panel
-        class="align-items-top filing-history-item"
-        v-for="(filing, index) in historyItems"
-        :key="index"
-      >
-        <!-- NB: bottom padding for when panel is collapsed -->
-        <v-expansion-panel-header class="no-dropdown-icon">
-          <div class="list-item">
-            <div class="filing-label">
-              <h3 class="list-item__title">{{filing.title}}{{correctionTag(filing)}}</h3>
-              <h4 v-if="filing.subtitle" class="list-item__title mt-1">{{filing.subtitle}}</h4>
+    <div class="scrollable-container" style="max-height: 1000px">
+      <v-expansion-panels v-if="historyItems.length > 0" v-model="panel">
+        <v-expansion-panel
+          class="align-items-top filing-history-item"
+          v-for="(filing, index) in historyItems"
+          :key="index"
+        >
+          <!-- NB: bottom padding for when panel is collapsed -->
+          <v-expansion-panel-header class="no-dropdown-icon">
+            <div class="list-item">
+              <div class="filing-label">
+                <h3 class="list-item__title">{{filing.title}}{{correctionTag(filing)}}</h3>
+                <h4 v-if="filing.subtitle" class="list-item__title mt-1">{{filing.subtitle}}</h4>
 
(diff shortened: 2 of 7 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -19,328 +19,330 @@
...
-                        class="add-detail-comment-item"
-                        @click="showCommentDialog(filing.filingId)"
-                      >
-                        Add Detail
-                      </v-list-item-title>
-                    </v-list-item>
-                  </v-list-item-group>
-                </v-list>
-              </v-menu>
+    <div class="scrollable-container" style="max-height: 1000px">
```
severinbeauvais, line 22: Actually, I don't mind the style property here, since it's the whole point for this div.

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -586,7 +586,8 @@ describe('Filing History List - incorporation applications', () => {
     expect(wrapper.find('.filing-label').text()).toContain('Incorporation Application')
     expect(wrapper.find('.filing-label').text()).toContain('Numbered Benefit Company')
     const spans = wrapper.findAll('.list-item__subtitle span')
-    expect(spans.at(0).text()).toBe('FILED AND PENDING (filed by Full Name on 2020-05-06)')
+    expect(spans.at(0).text())
+      .toBe('FILED AND PENDING (filed by Full Name on 2020-05-06) | EFFECTIVE as of 2020-05-06')
```
severinbeauvais, line 590: Are these tests in here for all the other variations (eg, without EFFECTIVE phrase)?

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Thanks!

tests/unit/FilingHistoryList.spec.ts
cameron-freshworks: There is yes :) Not all tests were impacted, ie the instances where there was no effective date it is confirmed to be NOT appended.

src/assets/styles/base.scss
```diff
@@ -68,6 +68,10 @@ p {
   font-weight: 400;
 }
 
+.app-action {
+  color: $app-blue !important;
```
andrepestana-aot, line 72: Do we really need !important? If not, we'd better not use it.

src/views/Dashboard.vue
```diff
@@ -95,12 +95,14 @@
                   <span>Change</span>
                 </v-btn>
               </header>
-              <v-card flat>
-                <director-list-sm
-                  :showCompleteYourFilingMessage="isIncorpAppTask"
-                  :showGrayedOut="isIncorpAppFiling"
-                />
-              </v-card>
+              <div class="scrollable-container" style="max-height: 890px">
```
andrepestana-aot, line 98: Should we use px unit here?

andrepestana-aot (commented):
LGTM, just some questions

src/assets/styles/base.scss
severinbeauvais: IMHO, if we use a base class, we hope it overrides any styles set in the component itself, so I don't mind the "important".

severinbeauvais (approved):

src/views/Dashboard.vue
severinbeauvais: FWIW, I have no opinion either way here. If this was a horizontal size then maybe I'd lean towards rem so that the browser can scale everything the same, but this is a height, and technically pages are infinitely long, so 🤷

andrepestana-aot: I don't think it's critical now as we are not thinking on mobile nor other resolutions but I think it may be a problem if the user increases their font size.

src/assets/styles/base.scss
andrepestana-aot: ok, yeah, no problem in this case. I just prefer not seeing any of them unless in rare cases but as we have already in many places in the project it's ok. These !importants may cause tricky issues if we start using them without discernment.

andrepestana-aot (approved):

src/assets/styles/base.scss
cameron-freshworks: We might be able to get away with NOT using it, but you run the risk of Vuetify overriding depending on where it is used.

src/views/Dashboard.vue
cameron-freshworks: What are you thinking as an alternative? rem?

src/assets/styles/base.scss
severinbeauvais: If we don't include "important", and we do want the style to override something else, what's the alternative? Another class with important, eg, "app-action-i" ? Yuck.
(Yes, CSS uses the most specific selector, but that's not something we should have in the base style file.)

severinbeauvais: > These !importants may cause tricky issues if we start using them without discernment.

Absolutely. My approach is styles in the component when they are specific, and styles in global assets otherwise. This seems to work OK mostly.

src/views/Dashboard.vue
andrepestana-aot: Yes, maybe rem. This needs to be tested changing the font size and check if it fits everything.

severinbeauvais: FWIW, most of our dialog widths are in rem.

cameron-freshworks: Totally, i'll give it a try!

cameron-freshworks: Works good!

PR #199, opened 2021-05-18
4388 - Create StaffNotation component

andrepestana-aot:
*Issue #:* /bcgov/entity#4388
*Description of changes:*

- Added StaffNotation component
- Added tests
- Added feature flag for staff notation feature


andrepestana-aot:
The bug reported by Sonar is related to v-deep and I think we should change its configuration to not report it. Should I ignore it?

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
```diff
@@ -0,0 +1,177 @@
...
+    this.isAddingCourtOrder = true
+  }
+  hideCourtOrderDialog (needReload: boolean): void {
+    this.isAddingCourtOrder = false
+    this.close(needReload)
+  }
+
+  @Emit('close')
+  private close (needReload: boolean): void {
+    // Intentionally empty
```
severinbeauvais, line 116: Who says this is a code smell without a body? We use `{}` in many places. I'm OK with that.

severinbeauvais:
> The bug reported by Sonar is related to v-deep and I think we should change its configuration to not report it. Should I ignore it?

I have done that already. I went into SC and told it this was a false positive.

andrepestana-aot:
@tlebedovich Please, for your preliminary review:

image

image

image

image

package.json
```diff
@@ -1,6 +1,6 @@
 {
```
severinbeauvais, line 1: I'm guessing that you upgraded your version of node/npm, and now it's creating "package-lock.json" with `lockfileVersion: 2`...

This is causing a huge diff for that file, which means we will all also have to update node/npm so that every other PR won't have this huge diff (version 1 to 2 differences).

We could all update, but last time we tried that, the GHA didn't work with the later version. (I can't remember if that was 1 project only or all of them.) Anyway, we had a good reason to not update last time.

It's possible the GHA will work with the new node/npm version now, but we can't tell until GHA is fixed.

So, let's keep an eye on this. If it's OK with this project then we will need to test all other projects. If that's OK then we can update all projects.

(I'd prefer to not require a npm version manager since I jump between project a lot.)

Or, we could defer this node/npm update until a future ticket. Thoughts?

src/assets/styles/base.scss
```diff
@@ -286,3 +286,9 @@ ul.list {
 .invalid-section {
   border-left: 3px solid $app-red;
 }
+
+.default-text {
+  font-weight: normal;
+  color: $gray7;
+  font-size: 1rem;
+}
```
severinbeauvais, line 294: I think you should consider overriding the default Vuetify text instead.

See what we did for NR here:
https://github.com/bcgov/namerequest/blob/9b3ff47e5f8b0a040b3764e6358603d06bed87fb/src/assets/scss/overrides.scss#L8

Rationale: you have defined a new class for something that already exists.

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
```diff
@@ -0,0 +1,177 @@
...
+            attach="#staff-notation"
+            itemName="Court Order"
+            filingType="courtOrder"
+            courtOrderNumberRequired="true"
+        />
+        <div class="filing-item__actions">
+            <v-menu offset-y left transition="slide-y-transition">
+                <template v-slot:activator="{ on }">
+                    <span>
+                        <span
```
severinbeauvais, line 29: Code smell: why do you have 2 spans? Maybe the outer one can be merged with your template element. Or maybe you can assign a style to change the block-level elements to inline ones. Or both.

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
```diff
@@ -0,0 +1,177 @@
...
+<style lang="scss" scoped>
+@import "@/assets/styles/theme.scss";
+#add-staff-filing-label {
+    padding-right: 0.725rem;
+    font-size: 0.875rem;
+    color: $app-blue;
+}
+
+#staff-notation {
+    padding-right: 1.5rem;
```
severinbeauvais, line 130: If possible, move layout/whitespace styles into the template and use the Vuetify helper classes. In my experience it's much easier to understand and update the layout if such styles are right in the template (instead of trying to find which CSS is affecting it).

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
```diff
@@ -0,0 +1,177 @@
...
+    scrollbar-color: transparent transparent; // FireFox uses this property
+
+    // Webkit browsers use ::-webkit-scrollbar pseudo element
+    &::-webkit-scrollbar {
+        background: transparent;
+    }
+}
+#staff-notation > div.filing-item__actions > button {
+    font-size: 0.875rem;
+    color: $app-blue;
```
severinbeauvais, line 144: We should have a global class for this type of button.

package.json
andrepestana-aot: Yes, thanks. My nvm wasn't set to have node 14.15.1 as default. I fixed that and rerun `npm i`. It'll be fixed in the next commit.

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
andrepestana-aot: Yes, I forgot to remove that. 👍

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
```diff
@@ -0,0 +1,177 @@
...
+    padding: 0.25rem !important;
+    color: $app-blue
+  }
+}
+#app > div.v-menu__content {
+    margin: 0.625rem 0 0 0;
+}
+.v-btn.active .v-icon {
+    transform: rotate(-180deg);
+  }
```
severinbeauvais, line 176: Fix indent.

That's a slick way to flip the down arrow :)

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,236 @@
...
+import { Component, Vue, Prop, Watch, Emit, Mixins } from 'vue-property-decorator'
+import { DateMixin } from '@/mixins'
+import { mapState } from 'vuex'
+import axios from '@/axios-auth'
+import { CourtOrderPoa } from '@bcrs-shared-components/court-order-poa'
+import { FormIF } from '@/interfaces'
+
+@Component({
+  computed: {
+    ...mapState(['entityIncNo'])
```
severinbeauvais, line 67: 1. Use a getter instead of reading state.
2. As this is a new TS component, can we use the `@Getter` style instead of this old style?

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,236 @@
...
+  @Prop() readonly filingType: string
+
+  /** Prop to display the dialog. */
+  @Prop() readonly dialog: boolean
+
+  /** Prop to provide attachment selector. */
+  @Prop() readonly attach: string
+
+  /** Prop to require court order number regardless the plan of arrangement. */
+  @Prop() readonly courtOrderNumberRequired: boolean
```
severinbeauvais, line 92: Do any of these props need a default value?

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
```diff
@@ -0,0 +1,177 @@
...
+    padding-right: 0.725rem;
+    font-size: 0.875rem;
+    color: $app-blue;
+}
+
+#staff-notation {
+    padding-right: 1.5rem;
+}
+
+.add-scrollbar-offset {
```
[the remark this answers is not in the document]
andrepestana-aot: 👍

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Emits event to close this dialog.
+   * @param needReload Whether the dashboard needs to be reloaded.
+   */
+  @Emit('close')
+  private emitClose (needReload: boolean): void {
+    this.enableValidation = false
+  }
+
+  /** WIP */
```
severinbeauvais, line 161: Still WIP?

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,236 @@
...
+
+    this.enableValidation = true
+    await this.$nextTick()
+    const isNotationFormRefValid = this.$refs.notationFormRef.validate()
+    const isCourtOrderPoaFormRefValid = this.$refs.courtOrderPoaRef.validate()
+    if (!isNotationFormRefValid || !isCourtOrderPoaFormRefValid) {
+      return
+    }
+
+    this.saving = true
```
severinbeauvais, line 175: I think you should move this as close to the `if saving` (line 165) as possible to minimize the user's opportunity to double-click the button. Especially since you await (ie, async) on line 168, and then do work on lines 169 and 170.

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
andrepestana-aot: That would keep the arrow animation like in a select: https://vuetifyjs.com/en/components/selects/

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,236 @@
...
+
+    const data = {
+      'filing': {
```
severinbeauvais, line 178: Please change format from JSON to object, ie,
```
filing: {
   header: ...
}
```

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: I just followed the other components. Let me check if that works.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,236 @@
...
+        'header': {
+          'name': this.filingType,
+          'date': this.currentDate
+          // 'certifiedBy': 'full name',
+          // 'email': 'no_one@never.get',
+          // 'filingId': 1
+        },
```
severinbeauvais, line 185: This looks incomplete.

Also, if my PR is merged before yours, please add `isStaff: this.isStaffRole` to the header.

Or if your PR is merged before mine, please remind me to add that ;)

src/views/Dashboard.vue
```diff
@@ -186,6 +193,10 @@ export default {
     /** The Incorporation Application's Temporary Registration Number string. */
     tempRegNumber (): string {
       return sessionStorage.getItem('TEMP_REG_NUMBER')
+    },
+
+    visibleForStaff (): boolean {
+      return this.isRoleStaff
```
severinbeauvais, line 199: Do you think we need a feature flag to limit how this feature is deployed? Or do you think this feature will be self-complete and can be deployed through Dev-Test-Prod normally (ie, in sync with API changes)?

src/views/Dashboard.vue
```diff
@@ -37,6 +37,11 @@
                 <h2 class="mb-3" data-test-id="dashboard-filing-history-subtitle">
                   <span>Recent Filing History</span>&nbsp;<span class="gray6">({{historyCount}})</span>
                 </h2>
+                <staff-notation
+                  v-if="visibleForStaff"
```
severinbeauvais, line 41: I'd be OK with you using `isRoleStaff` here directly.

Also see my comment in "visibleForStaff".

tests/unit/AddStaffNotationDialog.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+  it('renders the page contents correctly', () => {
+    const wrapper = shallowMount(AddStaffNotationDialog,
+      {
+        propsData: {
+          dialog: true,
+          itemName: 'Test',
+          attach: '#parent-page'
+        },
+        store,
+        vuetify
```
severinbeauvais, line 25: You may not need "vuetify" here since you're shallow-mounting the component. I'm not sure. Just an FYI - no need to change.

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: Yes, it is incomplete. I put a comment on top of the method. I'm still depending on #4390

I'm gonna add isStaff to the header. Thanks.

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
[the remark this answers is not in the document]
severinbeauvais: Something about having the same offset as the buttons in the filings list, blah blah blah...

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
severinbeauvais: Yes, very cool. I think in other places we toggle the v-icon between up-arrow and down-arrow. This is sexier :)

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Well there's some history here: this app originally didn't use the new class style, so the decoration was how you "imported" state, getters and setters. Now this app has a mixture of component styles. Since this is a net-new component, it would be great to use the new style exclusively instead of a hybrid style. If it works :)  (If not then it probably means we have to update a library, which is not in scope, so you could then ignore this comment.)

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: yes. Waiting for #4390

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: Yes. It won't be a problem not having a default but I added a default value to be explicit.

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: Yes, I tried to import it but can't find `vuex-class` :)

src/components/Dashboard/FilingHistoryList/StaffNotation.vue
andrepestana-aot: Agreed. 😄

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: What if you add it as a dependency?

severinbeauvais:
Would it be possible to reply to a comment without including the entire previous comment? Maybe delete the images at least? This PR is now very tall, requiring lots of scrolling to browse it.

src/views/Dashboard.vue
andrepestana-aot: I've been thinking of it... maybe it could be a good because if it's added and the other ticket to display those filings is not complete, it may be a problem, right?

src/views/Dashboard.vue
andrepestana-aot: yes, I had some issues with the tests. They showed messages that suggested me I should have a computed property then added it as a test. I removed now. Thanks!

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: I added the dependency and create a getter for `entityIncNo`. It works! 👍

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: I see we don't need the isStaff as we should trust the JWT token, right? Not adding then.

src/views/Dashboard.vue
severinbeauvais: OK, use this new FF. Default it to False.
image

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Yes, sorry about that. Finally the design has being discussed. Do not add `isStaff`.

src/views/Dashboard.vue
severinbeauvais: Actually, you'll want to restore `visibleForStaff()` in order to check the FF as well.

Example: https://github.com/bcgov/namerequest/blob/9b3ff47e5f8b0a040b3764e6358603d06bed87fb/src/components/existing-request/existing-request-search.vue#L165

andrepestana-aot:
> Thanks @andrepestana-aot - looking good so far. A couple minor items: Missing the vertical gray line in the "Add Staff Filing" segmented button. And on the modals, please make "Save" bold and "Cancel" should stay regular weight, but should be blue. Also, We don't want the user to be able to drag open the comment box in the modal, we will just show a scroll bar in the text box should they type in more than fits in the viewable area (I just noticed a little draggable icon in the lower right corner of the text box). Thanks.

@tlebedovich which shade of gray should I use for the vertical line? The background is `#f1f3f5` and the other vertical lines I see use the same shade of gray. I didn't see it in the template either. I'll do the other changes you mentioned. Thanks!

src/views/Dashboard.vue
andrepestana-aot: ok. Added.

src/assets/styles/overrides.scss
```diff
@@ -81,3 +81,22 @@
   color: $app-red !important;
   caret-color: $app-red !important;
 }
+
+.filing-item__actions {
```
severinbeauvais, line 85: This file should be global overrides, so should not contain component-specific classes.

Maybe I misunderstand (or didn't explain well)... I am in favour of `expand-btn` and `menu-btn` classes in here -- or reuse some existing (or new) classes in this file for your component, but don't make anything in here specific to "filing-item" for example.

Overrides in this file should directly apply to v- components.

If this can't be done in your scenario then please revert this code -- sorry.

severinbeauvais: Either way, I think your latest css/styles is cleaner than before - great!

andrepestana-aot: Yes, I know. I see that we always do it for that button, not only for this case. So I think it should be a global override.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,243 @@
...
+  }
+  #court-num-form div.v-input__slot {
+    margin-bottom: 0 !important;
+  }
+  #court-order .v-input--checkbox .v-input__slot {
+    margin-bottom: 0 !important;
+  }
+}
+.save-btn {
+  font-weight: 700;
```
severinbeauvais, line 236: Let's use `font-weight: bold` instead. This app has only 2 font weights: 400 and 700. Let's just use "normal" and "bold" instead of 400/500/600/700/etc.

severinbeauvais: (I know font-weight is used inconsistently in this app... Some day we should update that.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: 4390 is now in RFQA. Are  you able to finish this ticket now? If not, what do you think about merging what you have so far and having a "part 2" ticket to complete it later? What do you think?

severinbeauvais:
I re-ran the CI job to see if the linting and testing-coverage errors go away with your new package-lock file. If that doesn't work then you may need to ask Patrick for help since I don't see any obvious in the action logs.

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: Removed the comment and fixed the request. Now I think the most important is ok. I just need to fix css in another comment.

package-lock.json
```diff
@@ -17611,8 +17639,8 @@
       "dev": true
     },
     "vue-test-utils-helpers": {
-      "version": "git+https://github.com/bcgov/vue-test-utils-helpers.git#aeffe77a87b9a37dacf4b38d77bc1833db02a8a2",
-      "from": "git+https://github.com/bcgov/vue-test-utils-helpers.git",
+      "version": "git+ssh://git@github.com/bcgov/vue-test-utils-helpers.git#aeffe77a87b9a37dacf4b38d77bc1833db02a8a2",
+      "from": "vue-test-utils-helpers@git+https://github.com/bcgov/vue-test-utils-helpers.git",
```
andrepestana-aot: I noticed this. Reverted to https. I think that's the issue with GHA

andrepestana-aot: it worked 👍

src/components/dialogs/AddStaffNotationDialog.vue
andrepestana-aot: 👍 It was changed to bold

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -0,0 +1,237 @@
...
+      this.saving = false
+      return
+    }
+
+    const data = {
+      filing: {
+        header: {
+          name: this.filingType,
+          date: this.currentDate,
+          certifiedBy: ''
```
andrepestana-aot, line 180: cetifiedBy is required by the API but it's not used for this type of filing.

severinbeauvais (approved):
LGTM!

PR #201, opened 2021-05-19
Added and used flag to display filings filed by Registry Staff

severinbeauvais:
*Issue #:* /bcgov/entity#6814

*Description of changes:*
- displayed Filed By according to staff flag or certified by value, else fall back to Unknown
- saved isStaff flag in AR, Correction, COD and COA
- updated app version

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -541,7 +541,7 @@ export default {
           filingType,
           title: this.filingTypeToName(filingType, agmYear),
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy,
+          filingAuthor: this.filingAuthor(filing),
           filingDate,
           isPaid: (header.status === FilingStatus.PAID),
           documents: filing?.documents || [] as Array<any>,
@@ -607,7 +607,7 @@ export default {
           filingType,
           title: `${this.getCorpTypeDescription(this.entityType)} ${name} - ${corpName}`,
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy,
+          filingAuthor: this.filingAuthor(filing),
           filingDate,
           effectiveDateTime, // used in Future Effective component
           isFutureEffectiveIa,
@@ -681,7 +681,7 @@ export default {
           filingType: FilingTypes.ALTERATION,
           title,
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy || 'Registry Staff',
+          filingAuthor: this.filingAuthor(filing),
           filingDate,
           effectiveDateTime, // used in Future Effective component
           isFutureEffectiveAlteration,
@@ -744,7 +744,7 @@ export default {
           filingType,
           title: this.filingTypeToName(filingType),
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy,
+          filingAuthor: this.filingAuthor(filing),
           filingDate,
           effectiveDate, // used for BCOMP COA Future Effective tooltip
           isBcompCoaFutureEffective,
@@ -822,7 +822,7 @@ export default {
           filingType: FilingTypes.CORRECTION,
           title: `Correction - ${this.filingTypeToName(correction.correctedFilingType)}`,
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy || 'Registry Staff',
+          filingAuthor: this.filingAuthor(filing),
           filingDateTime: this.apiToSimpleDateTime(header.date), // used for receipt
           filingDate,
           isPaid: (header.status === FilingStatus.PAID),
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -922,7 +922,7 @@ export default {
           title: this.filingTypeToName(filingType),
           subtitle,
           filingId: header.filingId,
-          filingAuthor: 'Registry Staff', // TBD
+          filingAuthor: 'Registry Staff',
```
severinbeauvais, line 925: As far as I know, alteration filings are filed by Registry Staff (Ops) only.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -541,7 +541,7 @@ export default {
           filingType,
           title: this.filingTypeToName(filingType, agmYear),
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy,
+          filingAuthor: header.isStaff ? 'Registry Staff' : (header.certifiedBy || 'Unknown'),
```
severinbeauvais, line 544: I wasn't sure if this was worth a new function to call from several places... so I didn't. Thoughts?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -541,7 +541,7 @@ export default {
           filingType,
           title: this.filingTypeToName(filingType, agmYear),
           filingId: header.filingId,
-          filingAuthor: header.certifiedBy,
```
severinbeauvais, line 544: Note that we were missing the Registry Staff fallback for some of our filings. (Oops.) Fixed now.

Note also that the new behaviour is the same as the previous behaviour (except for the fixed filing types), and is backward compatible with current filings (that don't have the `isStaff` header property).

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Oh ya know i'm fine with ternary!

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Wait, is this true? 
Doesn't the client themselves file the Alterations? The staff only load the business and create the draft right?

severinbeauvais: I'll review this tomorrow!

src/components/Dashboard/FilingHistoryList.vue
andrepestana-aot: I'm ok with that too.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: You're right: alterations can be filed by users. And that code is correct (on line 684)...

Line 925 (above) is actually for the Transition Filings. I typed in the wrong filing type in my comment :P

severinbeauvais:
I can't see to tell SonarCloud to ignore the code duplication, but I'm not going to fix it as that's a refactoring exercise not in this ticket's scope. So please ignore that specific check.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1003,6 +1003,12 @@ export default {
       }
     },
 
+    filingAuthor (header: HeaderIF): string {
+      if (header.isStaff) return 'Registry Staff'
+      if (header.certifiedBy) return header.certifiedBy
+      return 'Unknown'
+    },
```
severinbeauvais, line 1010: I know y'all loved the ternary (with a || thrown in) but I think THIS code is easier to understand 😁

src/interfaces/header-interface.ts
```diff
@@ -2,11 +2,12 @@ import { FilingTypes } from '@/enums'
...
-  status: string;
-  availableInColinOnly?: boolean;
-  availableOnPaperOnly?: boolean;
+  date: string
+  filingId: number
+  name: FilingTypes
+  status: string
+  availableInColinOnly?: boolean
+  availableOnPaperOnly?: boolean
+  isStaff?: boolean
```
severinbeauvais, line 11: Optional since older filings don't have this field.

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -1315,7 +1316,8 @@ describe('Filing History List - alterations', () => {
             name: 'alteration',
             date: '2020-03-24T19:20:05.670859+00:00',
             status: 'COMPLETED',
-            filingId: 85114
+            filingId: 85116,
```
severinbeauvais, line 1319: This is the dumbest thing: SonarCloud was complaining about the duplicated 2 lines, so I changed the first line to work around that error 😛

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Sounds good! You can resolve when ready 👍

severinbeauvais:
There's a new design in the works. I am closing this PR and will open another (cleaner) one.

PR #203, opened 2021-05-21
History List Handling for Staff Filings

cameron-freshworks:
*Issue #:* /bcgov/entity#4389

*Description of changes:*
- Updated the filing history to include the new "registrar's notation" ledger item
- Updated the filing history to include the new "registrar's order" ledger item
- Updated the filing history to include the new "Court order" ledger item

- These ledger items cannot be corrected
- Detail comments enabled.

- Updated unit tests

*NOTES:*

* This was done off these UXpins, currently all we have: https://preview.uxpin.com/1f3b95dd1c4122ddb79c854e9498f5e8073cbfa1#/pages/130733104/simulate/no-panels

* Also done off the current schema, since these filings don't exist and needed to be mocked out. Schema is subject to change but this won't impact the Filings History List unless the title and/or FilingType changes. I'm assuming that won't.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.9.0",
+  "version": "3.9.1",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -72,6 +72,12 @@ p {
   color: $app-blue !important;
 }
 
+.info-text {
+  font-size: .875rem;
+  color: $gray7;
+  font-weight: normal;
+}
+
 .error-text {
   color: $app-red !important;
 }
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -787,9 +787,9 @@ export default {
               break
           }
 
-          item.notationOrOrder = baseFiling.orderDetails
-          item.fileNumber = baseFiling.fileNumber
-          item.planOfArrangement = baseFiling.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
+          item.notationOrOrder = baseFiling?.orderDetails
+          item.fileNumber = baseFiling?.fileNumber
+          item.planOfArrangement = baseFiling?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
         }
 
         // add receipt
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -182,7 +182,9 @@
                 >
                   <span v-if="filing.isColinFiling || filing.isPaperFiling" class="app-action">
                     {{ (panel === index) ? "Close" : "Request a Copy" }}</span>
-                  <span v-else class="app-action">{{ (panel === index) ? "Hide Documents" : "View Documents" }}</span>
+                  <span v-else class="app-action">
+                    {{ (panel === index) ? hideLabel(filing) : viewLabel(filing) }}
+                  </span>
                 </v-btn>
 
                 <!-- the drop-down menu -->
@@ -194,7 +196,7 @@
                   </template>
                   <v-list dense>
                     <v-list-item-group color="primary">
-                      <v-list-item :disabled="disableCorrection(filing)">
+                      <v-list-item v-if="!isStaffFiling(filing.filingType)" :disabled="disableCorrection(filing)">
                         <v-list-item-icon>
                           <v-icon class="app-action">mdi-file-document-edit-outline</v-icon>
                         </v-list-item-icon>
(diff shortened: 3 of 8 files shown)


vysakh-menon-aot (approved):

severinbeauvais (approved):
Great job!

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -501,6 +503,15 @@ export default {
...
                 this.loadTransitionFiling(filing)
                 break
+              case FilingTypes.REGISTRARS_NOTATION:
+                this.loadOtherFiling(filing, filing.registrarsNotation)
+                break
+              case FilingTypes.REGISTRARS_ORDER:
+                this.loadOtherFiling(filing, filing.registrarsOrder)
+                break
+              case FilingTypes.COURT_ORDER:
+                this.loadOtherFiling(filing, filing.courtOrder)
```
severinbeauvais, line 513: So the default item bodies display according to the comps?

cameron-freshworks: It does a great job yea! Worked by default without much effort :) 

I do however having another commit which will apply some staff only properties conditional, so stay tuned, will require another review!

cameron-freshworks:
> Great job!

Thanks! Little more to come 😁

severinbeauvais:
I clicked the "request review" button for me and Vysakh to remind us to re-review :)

src/interfaces/filing-interface.ts
```diff
@@ -11,5 +11,8 @@ export interface FilingIF {
   header: HeaderIF,
   incorporationApplication?: any;
   alteration?: AlterationIF;
+  registrarsNotation?: any;
+  registrarsOrder?: any;
+  courtOrder?: any;
```
cameron-freshworks, line 16: Future state: Add an interface to our shared components for these filings as we have other filings. 
Concerned we will still have some changes to come due to these filings still in development, so i'm certain some changes will happen along the way 👍

src/assets/styles/base.scss
```diff
@@ -72,6 +72,12 @@ p {
   color: $app-blue !important;
 }
 
+.info-text {
+  font-size: .875rem;
+  color: $gray7;
+  font-weight: normal;
+}
```
severinbeauvais, line 79: 👍

severinbeauvais, line 79: 👍

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -283,6 +285,14 @@
               <!-- NB: no documents so no divider needed -->
             </template>
 
+            <!-- is this a staff only filing? -->
+            <template v-else-if="isStaffFiling(filing.filingType)" id="staff-filings-info">
+              <p class="info-text">{{filing.notationOrOrder}}</p>
+              <p class="info-text my-0">Court Order Number: {{filing.fileNumber}}</p>
+              <p class="info-text my-0">{{filing.planOfArrangement}}</p>
```
severinbeauvais, line 292: Worth creating a sub-component?

cameron-freshworks: errr i don't think so. Tbh there is no visual design for this either, just a UX pin. 
If there is some standard format for this in the future perhaps. Right now it's as simple as this 🤷

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -756,6 +775,23 @@ export default {
...
+          switch (filingType) {
+            case FilingTypes.REGISTRARS_NOTATION: baseFiling = filing.registrarsNotation
+              break
+            case FilingTypes.REGISTRARS_ORDER: baseFiling = filing.registrarsOrder
+              break
+            case FilingTypes.COURT_ORDER: baseFiling = filing.courtOrder
+              break
+          }
+
+          item.notationOrOrder = baseFiling.orderDetails
```
severinbeauvais, line 790: If this code ever gets out of sync with `isStaffFiling()` then it's possible none of these cases apply and baseFiling will remain undefined, causing an error here.
You could init baseFiling to {}, or use conditional chaining.

cameron-freshworks: ah interesting! good catch. I will add conditional chaining 👍

src/interfaces/filing-interface.ts
severinbeauvais: Agreed. Do you think we need to keep track of this (new ticket or code comment?), or is it something you'll likely to do on the side, lol ?

severinbeauvais (approved):

src/interfaces/filing-interface.ts
cameron-freshworks: 😆 Good question. Can we leave this un resolved for now?
I will most likely just do this on the side. These filings are very small but have been subject to change.
Will be very small to add.

andrepestana-aot (approved):
Sorry for the late review. Really good!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.9.0",
+  "version": "3.9.1",
```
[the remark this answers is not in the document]
cameron-freshworks: Will do!

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -756,6 +775,23 @@ export default {
...
+              break
+            case FilingTypes.REGISTRARS_ORDER: baseFiling = filing.registrarsOrder
+              break
+            case FilingTypes.COURT_ORDER: baseFiling = filing.courtOrder
+              break
+          }
+
+          item.notationOrOrder = baseFiling?.orderDetails
+          item.fileNumber = baseFiling?.fileNumber
+          item.planOfArrangement = baseFiling?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
```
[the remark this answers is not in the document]
cameron-freshworks: Yea, display calls for another copy of the verbiage. So if the value exists display something else 😆

PR #208, opened 2021-05-25
4388 - UX Assurance - Postioned label aligned to the header

andrepestana-aot:
*Issue #:* /bcgov/entity#4388

*Description of changes:*

- Postioned label aligned to the header

### DIFF
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -22,7 +22,7 @@
             filingType="courtOrder"
             courtOrderNumberRequired="true"
         />
-        <div class="filing-item__actions">
+        <div class="filing-item__actions mb-2">
             <v-menu offset-y left transition="slide-y-transition" v-model="expand">
                 <template v-slot:activator="{ on }">
                     <span><!-- This span is needed to fix a positioning issue with the menu -->
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -22,7 +22,7 @@
             filingType="courtOrder"
             courtOrderNumberRequired="true"
         />
-        <div class="filing-item__actions mb-2">
+        <div class="filing-item__actions">
             <v-menu offset-y left transition="slide-y-transition" v-model="expand">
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/StaffNotation.vue
```diff
@@ -22,7 +22,7 @@
             filingType="courtOrder"
             courtOrderNumberRequired="true"
         />
-        <div class="filing-item__actions">
+        <div class="filing-item__actions mb-2">
```
severinbeauvais, line 25: It's usually better to add "top margin" to the element below this one, instead of adding "bottom margin" to this one. (Principle: it's easier to know what's above you than what's below you.)

What are you really trying to do here -- align this menu within the header?

image

It might work better to use negative margin at the top of this div to "pull it up" and aligned with the h2 in the header.

andrepestana-aot: >What are you really trying to do here -- align this menu within the header?

Yes.

>It might work better to use negative margin at the top of this div to "pull it up" and align it vertically with the h2 in the header.

The H2 in the header has bottom margin. I'm trying the negative margin/padding with no success. Let me see...

andrepestana-aot: It worked applying the negative margin to the main div. Thanks.

src/views/Dashboard.vue
```diff
@@ -41,6 +41,7 @@
                   v-if="visibleForStaff"
                   addScrollbarOffset="true"
                   @close="reloadDashboardIfNeeded($event)"
+                  class="mt-n3"
```
severinbeauvais, line 44: This is a different change that I didn't look at before 😛 

I see the H2 has `mb-3` (which would also work for your component here actually) but I still think it's better to have the lower element specify whitespace above it.

Is there any reason the bottom margin couldn't be removed from the header/h2, and top margin added to the element below the header?

severinbeauvais (approved):
Thanks. I appreciate you fixing something existing here :)

andrepestana-aot:
> Thanks. I appreciate you fixing an existing issue here :)

No worries. I got confused. Thanks!

PR #213, opened 2021-05-31
Added cursor to Staff Filing btn

cameron-freshworks:
*Issue #:* /bcgov/entity#7759

*Description of changes:*

* Added pointer to Staff Filing activator on hover.

### DIFF
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -119,6 +119,10 @@ export default class StaffNotation extends Vue {
     font-size: 0.875rem;
     color: $app-blue;
     border-right: 1px solid $gray3;
+
+    &:hover {
+      cursor: pointer;
+    }
 }
 
 // This class will be applied when addScrollbarOffset prop is true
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/StaffNotation.vue
```diff
@@ -119,6 +119,10 @@ export default class StaffNotation extends Vue {
     font-size: 0.875rem;
     color: $app-blue;
     border-right: 1px solid $gray3;
+
+    &:hover {
```
severinbeauvais, line 123: Just out of curiousity, what do you get if you don't specify the hover state?

severinbeauvais (approved):

src/components/Dashboard/StaffNotation.vue
cameron-freshworks: No difference in this case!

severinbeauvais: Oh right, the hover state would be to change the look of this "button"... not used in this case.

cameron-freshworks: Totally, I guess i did that out of habit 😁

PR #214, opened 2021-05-31
1password Keys Integration.

pwei1018:
*Issue #:*
https://github.com/bcgov/entity/issues/7849
https://github.com/bcgov/entity/issues/6801

*Description of changes:*
1. change the key new to match new configuration of 1password.
2. change to read the session keys directly instead of read from auth-api-config session key.
3. Update sbc-common-components from 2.4.18 to 2.4.19.


devops/vaults.json
```diff
@@ -13,10 +13,39 @@
             "address-key"
         ]
     },
+    {
```
severinbeauvais, line 16: I think I can guess what the purpose of this configuration is and how it works, but is it documented anywhere?

For example, if someone not familiar with this wants to add a "class" of config, or create a new app, or whatever, what do they need to do?

src/utils/fetch-config.ts
```diff
@@ -80,10 +73,14 @@ export async function fetchConfig (): Promise<void> {
...
   addressCompleteKey && console.info('Set Address Complete Key.')
 
-  const ldClientId = response.data['LD_CLIENT_ID']
+  const ldClientId = response.data['BUSINESS_FILING_LD_CLIENT_ID']
   window['ldClientId'] = ldClientId
   ldClientId && console.info('Set Launch Darkly Client ID.')
 
+  const sentryEnable = response.data['SENTRY_ENABLE'];
+  (<any>window).sentryEnable = sentryEnable
+  console.log('Set Sentry DSN Enable.')
```
severinbeauvais, line 82: Same as other PR: display value or delete this line.

devops/vaults.json
pwei1018: We have to be careful to add any values to this file. It should be done by Senior developer or SRE team member.

severinbeauvais (approved):

andrepestana-aot (approved):
Looks good to me!

PR #217, opened 2021-06-01
Update sbc-common-components version

eve-git:
*Issue #:* /bcgov/entity#6682

*Description of changes:*
Fix bug and disable browser autocomplete on address input

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15294,9 +15294,9 @@
       }
     },
     "sbc-common-components": {
-      "version": "2.4.16",
-      "resolved": "https://registry.npmjs.org/sbc-common-components/-/sbc-common-components-2.4.16.tgz",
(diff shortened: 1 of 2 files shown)


severinbeauvais (approved):
1. have the changes in the library been tested in this app?
2. I am OK with not updating the app version (since the "i" icon also displays the sbc-common-components version)
~3. we need to be careful about whether to merge this right now due to releases we're currently trying to push through~ _see response below_

severinbeauvais:
> 3. we need to be careful about whether to merge this right now due to releases we're currently trying to push through

I think we can merge in this PR when ready. (There are no current Edit UI releases that this would conflict with.) Please add the correct release label to the subject ticket (follow up with James and Linda if needed).

severinbeauvais:
> 3. we need to be careful about whether to merge this right now due to releases we're currently trying to push through

I think we can merge in this PR when ready. However, it will have to be in the NEXT release (not the [current release](https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/reports/release?release=6077649947db3a9f6444a55f)). Please add the correct release label to the subject ticket (follow up with James and Linda if needed).

PR #218, opened 2021-06-01
Added auth config key loading

severinbeauvais:
*Issue #:* /bcgov/entity#7694

*Description of changes:*
- added Auth Api Config workaround
- added local configuration
- updated app version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.9.10",
+  "version": "3.9.11",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
--- a/src/utils/fetch-config.ts
+++ b/src/utils/fetch-config.ts
@@ -31,6 +31,18 @@ export async function fetchConfig (): Promise<void> {
     return Promise.reject(new Error('Could not fetch configuration.json'))
   })
 
+  /**
+   * This is a workaround to fix the sbc-common-components that expect their own session keys.
+   * Ref: #6801
+   */
+  const authApiConfig = {
+    VUE_APP_AUTH_ROOT_API: response.data['SBC_CONFIG_AUTH_API_URL'],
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.9.10",
+  "version": "3.9.11",
```
severinbeauvais, line 3: I want to merge this in before the date PR.

src/utils/fetch-config.ts
```diff
@@ -31,6 +31,18 @@ export async function fetchConfig (): Promise<void> {
     return Promise.reject(new Error('Could not fetch configuration.json'))
   })
 
+  /**
+   * This is a workaround to fix the sbc-common-components that expect their own session keys.
+   * Ref: #6801
+   */
+  const authApiConfig = {
+    VUE_APP_AUTH_ROOT_API: response.data['SBC_CONFIG_AUTH_API_URL'],
+    VUE_APP_STATUS_ROOT_API: response.data['VUE_APP_STATUS_ROOT_API']
```
andrepestana-aot, line 40: Where is it used? I don't see where VUE_APP_STATUS_ROOT_API is used...

package.json
severinbeauvais: ... and before the sbc-common-components PR.

(Because this ticket is low risk and doesn't need much testing, whereas the other 2 tickets need actual QA.)

andrepestana-aot (changes requested):

src/utils/fetch-config.ts
severinbeauvais: Good eye! It is not used by our code. It is used by the SbcHeader and PaySystemAlert common components.

That's why I wrote that comment above this code :)

You don't need to read 6801 but basically it acknowledges that a work-around is needed at the moment.

andrepestana-aot: ok then :)

andrepestana-aot (approved):

cameron-freshworks (approved):
Looks good!

PR #221, opened 2021-06-03
Refactored Filings History List for new filings data

severinbeauvais:
*Issue #:* /bcgov/entity#7357

*Description of changes:*
- tbd


severinbeauvais:
Note that Code Coverage will improve to a satisfactory level when I complete the FHL unit tests.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.10.2",
+  "version": "4.0.0",
```
severinbeauvais, line 3: This feature is definitely a large, non-compatible change, so I think a major release change is appropriate.

src/App.vue
```diff
@@ -50,7 +50,8 @@
     </v-fade-transition>
 
     <SbcHeader />
-    <PaySystemAlert />
+    <!-- TODO: revert before final commit -->
+    <PaySystemAlert v-if="false" />
```
severinbeauvais, line 54: I'll go through all the TODOs in the next round (before final-final commit).

(This component generates a warning/error which was really annoying during development. Yes I created a ticket for this.)

package.json
cameron-freshworks: I agree, i think it's large enough to warrant version: `10.0.0` 😆

src/components/Dashboard/StaffNotation.vue
```diff
@@ -1,25 +1,25 @@
...
     <div id="staff-notation" class="pr-6" :class="{ 'add-scrollbar-offset': addScrollbarOffset }">
-        <add-staff-notation-dialog
+        <AddStaffNotationDialog
             :dialog="isAddingRegistrarsNotation"
             @close="hideRegistrarsNotationDialog($event)"
             attach="#staff-notation"
-            itemName="Registrar's Notation"
-            filingType="registrarsNotation"
+            displayName="Registrar's Notation"
+            name="registrarsNotation"
```
severinbeauvais, line 8: Cleanup to keep property names consistent.

src/components/StandaloneDirectorChange/CODDate.vue
```diff
@@ -107,6 +106,7 @@ export default class CodDate extends Mixins(DateMixin) {
    * @returns The minimum date that can be entered.
    */
   private get minDate (): string {
+    const lastCodFilingDate = this.dateToDateString(this.lastCodFilingDate)
```
severinbeauvais, line 109: `lastCodFilingDate` is now a state property and a Date.

src/components/StandaloneDirectorChange/CODDate.vue
```diff
@@ -76,11 +76,10 @@ export default class CodDate extends Mixins(DateMixin) {
 
   // Local definitions of computed properties for static type checking.
   // Use non-null assertion operator to allow use before assignment.
-  readonly currentDate!: string
```
severinbeauvais, line 79: This was already specified by an imported mixin and caused an error to re-declare it here.

src/components/common/Directors.vue
```diff
@@ -582,12 +583,10 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
 
   // Local definitions of computed properties for static type checking.
   // NB: use non-null assertion operator to allow use before assignment
-  readonly isBComp!: boolean
```
severinbeauvais, line 585: As above, this is already declared in an imported mixin.

src/components/common/Directors.vue
```diff
@@ -814,7 +814,7 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
           this.original.forEach((director, i) => {
             director.id = i + 1
             director.isFeeApplied = (director.isFeeApplied !== undefined) ? director.isFeeApplied : false
-            director.isDirectorActionable = (director.cessationDate === null)
+            director.isDirectorActionable = !director.cessationDate
```
severinbeauvais, line 817: This was a bug identified in the unit tests: `cessationDate` in this case was undefined (not null) so the logic didn't work.

Ditto on line 994 below.

src/interfaces/document-interface.ts
```diff
@@ -0,0 +1,10 @@
+import { DocumentTypes } from '@/enums'
```
severinbeauvais, line 1: I expect this interface will be used when we are able to fetch a list of documents from the API (soon).

src/interfaces/state-interface.ts
```diff
@@ -0,0 +1,42 @@
...
+  entityStatus: EntityStatus
+  entityBusinessNo: string
+  entityIncNo: string
+  entityFoundingDate: string
+  businessEmail: string
+  businessPhone: string
+  lastAnnualReportDate: string
+  lastFilingDate: Date
+  lastCoaFilingDate: Date
+  lastCodFilingDate: Date
```
severinbeauvais, line 23: Note that these are Javascript Dates. Until now we have mostly being using strings. Just be careful with using dates in this project.

src/components/Dashboard/TodoList.vue
```diff
@@ -701,16 +701,17 @@ export default {
...
     /** Loads a NEW Annual Report todo. */
-    loadAnnualReportTodo (task) {
+    loadAnnualReportTodo (task: any) {
       const todo: FilingIF = task.task.todo
       if (todo?.header && todo?.business) {
         const ARFilingYear: number = todo.header.ARFilingYear
         const subtitle: string = task.enabled && !this.isBComp ? '(including Address and/or Director Change)' : null
 
-        this.taskItems.push({
+        const item: TodoItemIF = {
```
severinbeauvais, line 710: I did it this was to type the object. (It identified some issues, which I fixed.)

src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,42 @@
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+import axios from '@/axios-auth'
+
+/** Mixin that provides integration with the Auth API. */
+@Component({})
+export default class AuthApiMixin extends Vue {
```
severinbeauvais, line 7: So I thought about creating a unit test file for this (and the other 2) API mixins, but they really don't do much so we really wouldn't be testing much (other than our test code). So I didn't.

src/mixins/date-mixin.ts
```diff
@@ -1,20 +1,46 @@
...
+export default class DateMixin extends Mixins(CommonMixin) {
   readonly currentDate!: string
 
+  /**
+   * Fetches and returns the web server's current date (in UTC).
+   * Used to bypass the user's local clock/timezone.
+   * Ref: https://www.npmjs.com/package/serverdate
+   * @returns a promise to return a Date object
+   */
+  async getServerDate (): Promise<Date> {
```
severinbeauvais, line 23: Copied from some of our other projects!

src/mixins/resource-lookup-mixin.ts
```diff
@@ -8,10 +8,10 @@ import { FilingCodes } from '@/enums'
  */
 @Component({
   computed: {
-    ...mapState(['configObject', 'isBComp'])
```
severinbeauvais, line 11: Bug! `isBComp` is not a state. I don't know how this worked before because it was always falsy ...?

src/store/actions.ts
```diff
@@ -14,7 +14,7 @@ export default {
   setNextARDate ({ commit }, nextARDate: string) {
     commit('nextARDate', nextARDate)
   },
-  setNameRequest ({ commit }, nameRequest: object) {
+  setNameRequest ({ commit }, nameRequest: any) {
```
severinbeauvais, line 17: It turns out objects have expected properties and methods. What we really want here is an "untyped object". `Any` seems to work fine and not cause type errors.

src/store/actions.ts
```diff
@@ -56,25 +56,34 @@ export default {
...
+  setDirectors ({ commit }, directors: Array<DirectorIF>) {
     commit('directors', directors)
   },
-  setLastAnnualReportDate ({ commit }, lastAnnualReportDate: string) {
-    commit('lastAnnualReportDate', lastAnnualReportDate)
+  setLastAnnualReportDate ({ commit }, date: string) {
+    commit('lastAnnualReportDate', date)
   },
-  setConfigObject ({ commit }, configObject: object) {
+  setLastFilingDate ({ commit }, date: Date) {
```
severinbeauvais, line 77: This is a good eample of some dates being Dates and others (above) being strings. Just be careful when using them. Sometimes they appear compatible (eg, printing a Date gives you a string) but they generally are not.

src/store/getters.ts
```diff
@@ -1,150 +1,95 @@
...
-  reportState (state): string {
+  reportState (state: StateIF): string {
     switch (state.currentFilingStatus) {
       case FilingStatus.NEW: return ''
       case FilingStatus.DRAFT: return 'Draft'
     }
     return state.currentFilingStatus
   },
 
-  /** Returns date of last Change of Directors filing from list of past filings. */
```
severinbeauvais, line 76: These are temporarily moved to App.vue until the API provides them (where we will fetch them in App.vue).

src/store/getters.ts
```diff
@@ -1,150 +1,95 @@
...
   /**
    * This is used to show Legal Obligations only for a new business
-   * that hasn't filed anything yet (except IA) and has no tasks.
+   * that has no tasks and hasn't filed anything yet (except their IA).
    **/
-  isBusinessWithNoMaintenanceFilings (state): boolean {
+  isBusinessWithNoMaintenanceFilings (state: StateIF): boolean {
     return (
+      // no todo items
+      state.tasks.length === 0 &&
```
severinbeauvais, line 84: Because most languages -- including JS -- perform "short circuit evaluation", if the first condition of a && is false, it doesn't check the rest. (Similarly, if the first condition of a || is true, it doesn't check the rest.) Therefore, checking easy/fast stuff first makes the code just a little bit faster. It's an easy optimization :)

src/store/state.ts
```diff
@@ -1,39 +1,43 @@
...
+import { StateIF } from '@/interfaces'
 
-export default {
+const state: StateIF = {
   // tombstone data
-  keycloakRoles: [] as Array<string>,
-  authRoles: [] as Array<string>,
-  username: null as string,
-  currentDate: null as string, // 'today' as YYYY-MM-DD
+  keycloakRoles: [],
```
severinbeauvais, line 5: I had attempted typing in the old code but it was never carried throughout the code. Now it works correctly (in class syntax files).

You can mouse over any of these properties to see their type.

tests/unit/AGMDate.spec.ts
```diff
@@ -13,7 +13,7 @@ Vue.use(Vuetify)
 Vue.use(Vuelidate)
 
 const vuetify = new Vuetify({})
-const store = getVuexStore()
+const store = getVuexStore() as any // remove typings for unit tests
```
severinbeauvais, line 16: ... Because we want the tests to be fragile (eg, use strings instead of types in tests so that we end up testing the types too).

tests/unit/App.spec.ts
```diff
@@ -1478,7 +1481,6 @@ describe('App as a Paid Incorporation Application', () => {
 
   it('fetches IA filings properly', () => {
     expect(vm.$store.state.entityIncNo).toBe('T123456789')
-    expect(vm.$store.state.entityType).toBe('BEN')
```
severinbeauvais, line 1481: This is not provided in a filing atm.

tests/unit/TodoList.spec.ts
```diff
@@ -505,7 +505,7 @@ describe('TodoList - UI', () => {
             'header': {
               'name': 'correction',
               'ARFilingYear': 2019,
-              'status': 'PENDING_CORRECTION',
```
severinbeauvais, line 508: That's right, a PENDING_CORRECTION is now a "PENDING" "correction" task :)

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -7,22 +7,22 @@ import flushPromises from 'flush-promises'
 import axios from '@/axios-auth'
```
severinbeauvais, line 7: This file is still under construction (while you review the rest of the changes :) ).

It's because I disabled a bunch of these tests that the code coverage is low for this module.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1,25 +1,35 @@
...
-    <load-correction-dialog
+    <LoadCorrectionDialog
       :dialog="loadCorrectionDialog"
       @exit="loadCorrectionDialog=false"
       attach="#filing-history-list"
     />
 
-    <div class="scrollable-container" style="max-height: 56rem">
+    <!-- Alternate Loading Spinner -->
+    <v-fade-transition>
```
severinbeauvais, line 23: I added this to show while fetching documents and detail comments. I tried simply changing buttons to their loading state but it was overly complicated and didn't look right. This spinner seems to work well (IMHO) (locally anyway).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1042 +376,665 @@
...
 
-    ...mapState(['filings', 'entityIncNo', 'entityName', 'entityType']),
+  private loadData (): void {
+    this.historyItems = []
 
-    /** The Pay API URL string. */
-    payApiUrl (): string {
-      return sessionStorage.getItem('PAY_API_URL')
-    },
+    // create history items from 'filings' array from API
```
severinbeauvais, line 473: I considered using just one list (ie, the original ledgers/filings array from the API) instead of creating a second array -- and I synced the property names between them to be closer to that -- but it would be overly complicated... so there are still 2 arrays.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1042 +376,665 @@
...
+    if (needReload) {
+      // find the filing in the list
+      const item = this.historyItems.find(item => (item.filingId === this.currentFilingId))
 
-            if (!iaFiling) {
-              throw new Error('Invalid API response')
-            }
+      // TODO: test "isBusy" spinner when we have commentsLink
+      if (item && item.commentsLink) {
+        this.isBusy = true // TODO: test this
```
severinbeauvais, line 879: Yes, I will go through all TODOs before my final submit. Some of this functionality is still blocked by API work.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1042 +376,665 @@
...
-
-      // Notice of articles or certificate will come in as a document report type
-      if (document.reportType) {
-        url += `?type=${document.reportType}`
-      }
+  /** Whether to disable correction for this history item. */
+  private disableCorrection (item: HistoryItemIF): boolean {
+    return (
+      this.disableChanges ||
+      item.availableOnPaperOnly ||
```
severinbeauvais, line 945: Paper filings now cannot be corrected. (Previously they could but Colin filings couldn't.... and now these are merged.) This is still an outstanding question for Linda in the ticket description and I will update this if needed.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1042 +376,665 @@
...
+
+  // enums for template
+  readonly DocumentTypes = DocumentTypes
+  readonly FilingTypes = FilingTypes
+
+  readonly isBComp!: boolean
+  readonly isRoleStaff!: boolean
+  readonly nrNumber!: string
+  readonly filings!: Array<LedgerIF>
+  readonly entityName!: string
```
cameron-freshworks, line 440: Are these computed properties just declared on the class? Is this the alternative because we're not using vuex-class decorators here?
Sorry my memory is short 😆

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1042 +376,665 @@
...
-        console.log('ERROR - missing section in filing =', filing)
+  /** Loads a general filing into the historyItems list. */
+  private loadOtherFiling (filing: LedgerIF): void {
+    try {
+      const effectiveDate = new Date(filing.effectiveDate)
+      const submittedDate = new Date(filing.submittedDate)
+
+      // TODO: remove when API already has this in display name
+      let displayName = filing.displayName ||
+        this.filingTypeToName(filing.name) // TODO: delete when API handles unknown filings
```
cameron-freshworks, line 634: Could this be 1 line or do you prefer a break on an OR?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: This is the alternate syntax for getters (which are declared in the component declaration. Actually I prefer the other syntax so I'll change it :)

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: With the comment, it didn't all fit on 1 line.

Anyway, line 634 will go away when the API always returns a displayName (which it currently doesn't), so this expression will then be 1 line.

Actually the whole block should go away when the API gives us the entire displayName.

severinbeauvais: You will find other "TODO"s in the code with a comment such as, "remove when API...".

Sorry, this is currently in an indeterminate state where the API doesn't yet provide all needed functionality, but the current code is usable/testable. I don't expect large changes when the API is complete; most of those TODOs should be easy to resolve.

src/mixins/enum-mixin.ts
```diff
@@ -55,46 +65,74 @@ export default class EnumMixin extends Vue {
...
-    return (item.filingType === FilingTypes.NAME_REQUEST)
+    return (item.name === FilingTypes.NAME_REQUEST)
+  }
+
+  /** Returns True if filing is a Transition. */
+  isTypeTransition (item: any): boolean {
+    return (item.name === FilingTypes.TRANSITION)
+  }
+
+  /** Returns True is filing is a Staff Only filing. */
```
cameron-freshworks, line 129: Probably my typo here, but a type none the less 😁  `Returns true if..`

severinbeauvais: Fixed.

src/mixins/auth-api-mixin.ts
cameron-freshworks: Good call. On a regular mount i believe tests can run through code which utilize the mixins, which in a way tests them anywho right? If not always in isolation but at least in practice.

cameron-freshworks (commented):
Soo many changes it is hard to grasp the big picture but i've no doubt you've done a great job with such far reaching changes. 
It looks good to me, and i'm sure if we might find some edge case oddities, we can just tackle them then! 👍

src/interfaces/state-interface.ts
```diff
@@ -0,0 +1,42 @@
...
+  arMaxDate: string // YYYY-MM-DD
+
+  tasks: Array<any> // tasks from API
+  filings: Array<LedgerIF> // filings from API
+  registeredAddress: OfficeAddressIF
+  recordsAddress: OfficeAddressIF
+  directors: Array<DirectorIF>
+
+  currentFilingStatus: FilingStatus
+  configObject: any
```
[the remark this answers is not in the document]
severinbeauvais: This was pre-existing. It contains entity-specific resources.

src/mixins/auth-api-mixin.ts
severinbeauvais: Haha, I've actually partly changed my mind on this, in order to increase code coverage!

Ideally, in components that use these API mixins, we should mock the mixin methods. But I'm not changing that now :)

severinbeauvais:
> Soo many changes it is hard to grasp the big picture but i've no doubt you've done a great job with such far reaching changes.
> It looks good to me, and i'm sure if we might find some edge case oddities, we can just tackle them then! 👍

Yeah, sorry. It's a big change. Bigger than I expected, for sure.

The big picture?
- the filings endpoint now returns a list of different objects
- the tasks endpoint is unchanged
- fetching a single filing is unchanged
- fetching a single task is unchanged
- a filing doesn't come with comments or documents; a new endpoint needs to be called to fetch them
- typing is much improved (ie, LedgerIF and StateIF)
- the FHL doesn't need to compute titles or filenames (the API will provide them)
- Last Filing Dates doesn't need to be computed in the UI (the API will provide them)

src/mixins/date-mixin.ts
```diff
@@ -1,20 +1,46 @@
...
+    const input = `${window.location.origin}/${process.env.VUE_APP_PATH}/`
+    const init: RequestInit = { cache: 'no-store', method: 'HEAD' }
+
+    // don't call fetch() during Jest tests
+    // because it's not defined
+    if (this.isJestRunning) return new Date()
```
argush3, line 29: This feels off.  Maybe this should be something that is mocked out in the tests.

severinbeauvais: Agreed. And I did try for a while then gave up and did this instead. What do you think?

argush3: I'm not really familiar with Jest but i just figured we would be able to mock or stub out `getServerDate` in the date-mixin where required.   I think I see sinon being used already in some of the tests.  What kind of problems were you seeing trying to mock/stub things out?

Anyways, not a deal breaker but it would be nice to remove that dependency at some point.

severinbeauvais: Jest does allow you to mock a function, but this function is in a mixin and I can't seem to mock it where it's used (in App.vue tests).

```
jest.spyOn(vm, 'getServerDate').mockImplementation(() => new Date()) // method is not overridden
```

The root issue I have is that `fetch()` is undefined within the Jest tests, and I'm trying to avoid the `console.warn()` in the catch block.

severinbeauvais: I managed to mock `fetch()`... see new commit.

argush3: nice! 👍

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Updated.

src/App.vue
severinbeauvais: This is now reverted.

src/App.vue
```diff
@@ -577,22 +536,41 @@ export default {
...
-          this.setFilings([{ filing }])
+          // build display name
+          const type: string = this.getCorpTypeDescription(this.entityType)
+          const name: string = this.filingTypeToName(FilingTypes.INCORPORATION_APPLICATION)
+          const desc: string = this.entityName || this.getCorpTypeNumberedDescription(this.entityType)
+
+          const effectiveDate: Date = this.apiToDate(filing.header.effectiveDate)
+
+          // add this as a filing (for Filing History List)
+          const filingItem: LedgerIF = {
```
severinbeauvais, line 547: I created [# 8692](https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/8692) to refactor this.

src/interfaces/historyItem-interface.ts
```diff
@@ -1,48 +1,47 @@
-import { CorpTypeCd, FilingTypes } from '@/enums'
+import { FilingTypes } from '@/enums'
```
severinbeauvais, line 1: These are all basically the same properties as before, but now * organized *.

src/mixins/legal-api-mixin.ts
```diff
@@ -1,26 +1,90 @@
...
+    return axios.get(url)
+  }
+
+  /**
+   * Fetches filings list for the specified business ID.
+   */
+  async fetchFilings (businessId: string): Promise<any> {
+    // *** TODO: revert before final commit
+    // const url = `businesses/${businessId}/filings`
+    const url = `https://legal-api-dev-pr.apps.silver.devops.gov.bc.ca/api/v1/businesses/${businessId}/filings`
```
severinbeauvais, line 33: Temporary URL - will be reverted before final commit.

src/mixins/legal-api-mixin.ts
```diff
@@ -75,4 +139,91 @@ export default class LegalApiMixin extends Vue {
...
+        return documents
+      })
+  }
+
+  /**
+   * Fetches one document and prompts browser to open/save it.
+   * See also PayApiMixin::fetchOneReceipt().
+   * @param document the document info object
+   */
+  async fetchOneDocument (entityIncNo: string, document: any): Promise<any> {
```
severinbeauvais, line 184: Most of these methods -- like this one -- previously existed somewhere else. I moved them here cause this is where they belong!

src/interfaces/todoItem-interface.ts
```diff
@@ -0,0 +1,43 @@
+import { FilingTypes, PaymentMethod } from '@/enums'
+import { PaymentErrorIF } from '@/interfaces'
+
+/** A Todo List item. */
+export interface TodoItemIF {
```
severinbeauvais, line 5: A "todo item" is what is in our local list and displayed in the Todo List.

Conversely, a "task item" is what got from the Lear "tasks" endpoint (ie, API object).

tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "strict": false,
     "jsx": "preserve",
     "importHelpers": true,
+    "resolveJsonModule": true,
```
severinbeauvais, line 8: This is needed to be able to load a JSON file (in FilingHistoryList2.spec.ts), ie,
`import { filings } from './filings.json'`

src/App.vue
```diff
@@ -127,12 +125,7 @@ export default {
         CorpTypeCd.BC_ULC_COMPANY,
         CorpTypeCd.BC_CCC,
         CorpTypeCd.COOP
-      ],
-
-      // enums
-      EntityStatus,
```
severinbeauvais, line 133: This is only needed if the enum is used in the template... which these are not.

tests/unit/EntityInfo.spec.ts
```diff
@@ -253,15 +253,7 @@ describe('EntityInfo - Click Tests - Alterations', () => {
     // store.state.entityType = 'LTD' // FUTURE: uncomment this
 
     const router = mockRouter.mock()
-    const wrapper = mount(EntityInfo, {
-      vuetify,
-      store,
-      router,
-      computed: {
-        // mock this getter to override FF check
-        viewChangeInfoEnabled () { return true }
```
severinbeauvais, line 262: I removed the FF check.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -705,16 +705,16 @@ export default {
...
+    // FUTURE move this to Legal API mixin
     async hasTasks (businessId) {
       let hasPendingItems = false
       if (this.filingId === 0) {
         const url = `businesses/${businessId}/tasks`
         await axios.get(url)
           .then(response => {
-            if (response && response.data && response.data.tasks) {
+            if (response?.data?.tasks) {
               response.data.tasks.forEach((task) => {
```
severinbeauvais, line 716: This could probably be improved with a `find()` (or `some()`) so it doesn't iterate over all tasks.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -174,8 +174,8 @@
             </v-btn>
             </div>
           </template>
-          <span>Ensure all of your information is entered correctly before you File.<br>
```
severinbeauvais, line 177: Span tags are not needed in a tooltip's content.

src/interfaces/api-header-interface.ts
```diff
@@ -0,0 +1,23 @@
...
+  email: string // *** TODO: remove if obsolete
+  filingId: number
+  isFutureEffective: boolean // *** TODO: remove if obsolete
+  name: FilingTypes
+  paymentMethod: any
+  paymentStatusCode: string
+  paymentToken: any // NB: may be UUID in future
+  priority?: boolean // alterations and corrections only
+  status: FilingStatus
+  submitter: string // *** TODO: remove if obsolete
```
severinbeauvais, line 22: Need to wait and see what new Legal API provides (and test, and adjust accordingly).

src/components/common/Directors.vue
```diff
@@ -894,7 +894,8 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
     }
 
     // set form to initial director data again
-    this.newDirector = { ...EmptyDirector, appointmentDate: this.asOfDate }
+    this.newDirector = cloneDeep(EmptyDirector)
```
severinbeauvais, line 897: Need to use `cloneDeep()` to copy the sub-objects too (not assign them).

src/components/common/Directors.vue
```diff
@@ -621,7 +620,7 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
   private messageIndex = -1
 
   /** V-model for new director form. */
-  private newDirector: DirectorIF = { ...EmptyDirector }
+  private newDirector = cloneDeep(EmptyDirector)
```
severinbeauvais, line 623: Need to use `cloneDeep()` to copy the sub-objects too (not assign them).

src/components/EntityInfo.vue
```diff
@@ -149,24 +157,28 @@ export default class EntityInfo extends Mixins(CommonMixin, EnumMixin) {
...
 
   readonly axios = axios // for template
 
   /** Whether to show the hover style. */
   private showHoverStyle = false
 
   /** True if View and Change Company Info button should be rendered. */
-  private get viewChangeInfoEnabled (): boolean {
-    return (this.isBComp || this.isBcCompany || this.isUlc) && getFeatureFlag('alteration-ui-enabled')
+  private get showViewChangeInfoBtn (): boolean {
```
severinbeauvais, line 176: Renamed so it's not confused with whether the button is disabled.

severinbeauvais:
Note that the code coverage is lower because I have currently disabled the App.vue tests while I wait to hear what the temp reg number filing (IA) will look like.

src/components/Dashboard/FilingHistoryList/CompletedAlteration.vue
```diff
@@ -2,10 +2,10 @@
   <div v-if="filing" class="completed-alteration-details body-2">
     <h4>Alteration Complete</h4>
 
-    <p v-if="filing.newLegalType !== filing.oldLegalType">
-      {{entityName || 'This company'}} was successfully altered
-      from a {{filing.oldLegalType}} to a {{filing.newLegalType}}
-      on {{filing.effectiveDateTime}} Pacific Time.
+    <p v-if="filing.toLegalType !== filing.fromLegalType">
```
cameron-freshworks, line 5: This is nice clarity. Is this a direct representation of property value names from the api?

severinbeauvais: Yes. In Filing History List, we have:
```
item.toLegalType = filing.data.alteration?.toLegalType || null
item.fromLegalType = filing.data.alteration?.fromLegalType || null
```

However, this raises the idea of using the list from the API directly in the template, instead of converting it into `historyItems`. I thought that would be too much churn so I didn't pursue it.

severinbeauvais: The main drivers for this refactor are:
- API provides more ready-to-use data (so UI doesn't need lots of business logic)
- fetching the data is faster since less data is provided up front (eg, the comments and documents are loaded on demand)
- probably lots of API cleanup (and, obviously, UI cleanup too 😀 )

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
```diff
@@ -22,50 +22,51 @@
...
+  @Getter getEntityName!: string
 
   /** The subject filing. */
-  @Prop({ required: true }) private filing: HistoryItemIF
+  @Prop({ required: true })
+  readonly filing: HistoryItemIF
 
   /** Data for the subject filing. */
-  private get _ (): any {
+  get _ (): any {
```
cameron-freshworks, line 42: I'm unfamiliar with the syntax here, could you explain for my know-how :)

src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /**
+   * Fetches entity info.
+   * @param businessId the business identifier (aka entity inc no)
+   */
+  async fetchEntityInfo (businessId: string): Promise<any> {
+    const url = `${this.authApiUrl}entities/${businessId}`
+    return axios.get(url)
+  }
```
cameron-freshworks, line 39: Are these used in the App yet, or for future use?

cameron-freshworks (commented):
A lot of good clean up and useful additions. 
Obviously it's high risk due to the amount of changes but we won't find everything in a code review especially considering the heavy dependancy on the LEAR update. 
IMO when that api update is ready to go, we get them both in DEV and begin regression testing!

src/components/Dashboard/FilingHistoryList/FutureEffective.vue
severinbeauvais: There's nothing fancy with the syntax. I thought I'd copy the Lodash example of importing their methods into a class named `_` (eg, _.cloneDeep). So this getter is named `_`.

The idea is to have something simple, that isn't `this`. I could have named it `data` but that seemed weird and repetitive instead of clean.

src/mixins/auth-api-mixin.ts
severinbeauvais: Yes, these were in App.vue previously. They are still called from there.

src/components/common/DetailsList.vue
```diff
@@ -34,36 +35,36 @@
...
-  /** Filing containing correction information. */
-  @Prop({ default: () => { return { comments: [] } } })
-  private filing: object
+  /** The filing containing comments. */
+  @Prop({ default: () =>
+    ({
+      comments: [],
+      filingId: null
+    })
+  })
```
severinbeauvais, line 51: @cameron-freshworks Here's some new syntax for your know-how. (It was new to me!)

-> for arrow functions, instead of `() => return { xxx }`, just use `() => ({ xxx })`

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1065 +349,576 @@
...
-        }
-      }
-      return false
-    },
 
-    /** Whether this filing(alteration) is pursuant to a plan of arrangement  */
-    isPlanOfArrangement (courtOrder: CourtOrderIF): boolean {
-      if (!courtOrder || !courtOrder.effectOfOrder) {
-        return false
+      // add properties for correction filings
```
argush3, line 518: line 518 and 524 comments should be swapped?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -374,1065 +349,576 @@
...
-        }
-
-        this.historyItems.push(item)
-      } else {
-        // eslint-disable-next-line no-console
-        console.log('ERROR - missing section in filing =', filing)
+      // special handling for paper-only filings
+      if (filing.availableOnPaperOnly) {
+        // clear URL so we don't try to load documents
+        item.documentsLink = null
```
argush3, line 515: so the API will return a documents link even for paper filings?

src/App.vue
```diff
@@ -507,90 +462,120 @@ export default {
...
+
+    storeFiledIa (filing: any): void {
+      // NB: this was already validated in storeIncorpApp()
+      const incorporationApplication = filing.data.incorporationApplication
+
+      // set temporary addresses
+      this.storeAddresses({ data: incorporationApplication.offices || [] })
+
+      // set temporary directors
+      const directors = incorporationApplication.parties?.filter(party => party.roles.filter(role =>
```
argush3, line 553: readability a bit difficult for this.  not a big thing but just something i thought when reading this code

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I think this is actually correct:
- a Correction Filing has the id of the filing that it corrects
- A Corrected Filing has the id of the filing that corrects it

argush3: ok 👍

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I don't know yet :)

This code clears this property so that we don't try to fetch documents for this filing (when the user clicks to expand this item).

I'm torn whether I should leave this (as a safety) or delete it (if the API doesn't provide it). Thoughts?

src/App.vue
severinbeauvais: Agreed. Is this better? (If not then I'll break it up.)
```
      const directors = incorporationApplication.parties?.filter(party =>
        party.roles.filter(role =>
          role.roleType === Roles.DIRECTOR
        ).length !== 0
      )
```

argush3: looks good!

src/components/Dashboard/FilingHistoryList.vue
argush3: I was just curious.  I think we can keep it just to be safe.

src/components/StandaloneDirectorChange/CODDate.vue
```diff
@@ -67,7 +67,7 @@ import { DateMixin } from '@/mixins'
 export default class CodDate extends Mixins(DateMixin) {
   // Prop passed into this component.
   @Prop({ default: '' })
-  private initialCODDate: string
+  readonly initialCODDate: string
```
argush3, line 70: `Cod` seems to be used in other places.  do we want to do that here too?

severinbeauvais: OK, I changed variables with "COD" to "Cod" (and same for COA).

src/store/getters.ts
```diff
@@ -62,14 +77,24 @@ export default {
...
+  isPendingDissolution (state: StateIF): boolean {
+    return (state.entityStatus === EntityStatus.PENDING_DISSOLUTION)
+  },
+
+  /** Is True if business is not in compliance. */
+  isNotInCompliance (state: StateIF): boolean {
+    return (state.entityStatus === EntityStatus.NOT_IN_COMPLIANCE)
   },
 
   nrNumber (state: StateIF): string {
```
argush3, line 95: a lot of properties are prefixed with `get` in this getters file.  Do we want to update the other ones to be prefixed with `get`?  i.e. nrNumber, reportState, entityIncNo, coaEffectiveDate

argush3:
I've taken a pass and all I can say is great job tackling this huge refactor/cleanup of the ledger!  I've gone through the code and things look good to me.  Just the things I commented on.  Like Cameron was saying, we'll just need to test this as a team when we get it into our enviornments as it's a pretty big change both on FE and BE.

src/store/getters.ts
severinbeauvais: I tried to add some consistency to the getters, as you mention, but it's not perfect. For example, we have a lot of getters with the name `isXxx..`. Should these be changed to `getIsXxx..`? I don't love that. And either way, getters could be confused with local (component) getters or properties. So the current style is not perfect.

That said, there are some low hanging fruit (the ones you mention) that I will change over :)

severinbeauvais:
> I've taken a pass and all I can say is great job tackling this huge refactor/cleanup of the ledger! I've gone through the code and things look good to me. Just the things I commented on. Like Cameron was saying, we'll just need to test this as a team when we get it into our enviornments as it's a pretty big change both on FE and BE.

Thanks so much to you and Cam for your thorough review!

src/store/getters.ts
severinbeauvais: Done.

PR #222, opened 2021-06-04
Updated Staff Comment component import

severinbeauvais:
*Issue #:* /bcgov/entity#7079

*Description of changes:*
- updated staff comment import (to use new date mixin)
- updated app version

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "3.10.0",
+  "version": "3.10.1",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "@bcrs-shared-components/court-order-poa": "1.0.11",
     "@bcrs-shared-components/enums": "1.0.8",
     "@bcrs-shared-components/interfaces": "1.0.17",
-    "@bcrs-shared-components/staff-comments": "1.0.3",
+    "@bcrs-shared-components/staff-comments": "1.1.0",
```
severinbeauvais, line 23: FYI, I went from 1.0.x to 1.1.0 due to a change that may be incompatible in some apps. (Ie, the app has to provide the updated date mixin, so the minor version change should be a trigger to verify this when importing the updated component.)

eve-git (approved):
Looks good.

andrepestana-aot (approved):
👍

PR #226, opened 2021-09-02
8150 Refactor Todo List to use AR min/max dates

severinbeauvais:
*Issue #:* /bcgov/entity#8150

*Description of changes:*
- app version: 3.10.4
- fixed some typings
- fixed "colin only" prop change
- continue to get todo AR min/max dates from header
- use AR max date for BCOMP due date
- updated data checks
- continue to use draft AR date fallbacks
- added missing Filing Status values
- renamed Header Interface -> Filing Header Interface and updated props
- added header type in Filing Interface
- updated Task Item Interface
- added Todo Header Interface
- misc cleanup
- deleted obsolete Cache ID
- no longer set AR Filing Year for draft CODs and COAs


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -477,7 +477,7 @@ export default {
         const filing = this.filings[i].filing
         if (filing?.header?.date) {
           let filingDate = filing.header.date.slice(0, 10)
-          if (filing.header.availableInColinOnly) {
+          if (filing.header.availableInColinOnly || filing.header.inColinOnly) {
```
severinbeauvais, line 480: I'm not sure when this was changed but now the UI handles both cases.

@thorwolpert COLIN filings go away in the new ledge, right?

src/components/common/NameRequestInfo.vue
```diff
@@ -91,7 +91,7 @@ export default class NameRequestInfo extends Mixins(DateMixin, EnumMixin, NameRe
 
   /** The entity title  */
   private entityTypeDescription (): string {
-    return this.getCorpTypeDescription(this.parsedNameRequest.entityType)
+    return this.getCorpTypeDescription(this.parsedNameRequest.entityType as CorpTypeCd)
```
severinbeauvais, line 94: This was OK previously, so it's either:
- a change in BCRS Shared Component
- an improvement in lint / Vue tools

src/enums/filingStatus.ts
```diff
@@ -1,11 +1,15 @@
...
+  EPOCH = 'EPOCH', // soon to be obsolete
+  ERROR = 'ERROR', // soon to be obsolete
+  NEW = 'NEW', // for todo tasks only
   PAID = 'PAID',
   PENDING = 'PENDING',
-  PENDING_ALTERATION = 'PENDING_ALTERATION',
-  PENDING_CORRECTION = 'PENDING_CORRECTION',
+  PENDING_ALTERATION = 'PENDING_ALTERATION', // soon to be obsolete
+  PENDING_CORRECTION = 'PENDING_CORRECTION', // soon to be obsolete
+  WITHDRAWN = 'WITHDRAWN'
```
severinbeauvais, line 14: I added missing types according to:

filing-states

src/interfaces/filing-header-interface.ts
```diff
@@ -0,0 +1,28 @@
+import { FilingStatus, FilingTypes } from '@/enums'
+
+/** A filing's header object from the API, ie, `filing.header`. */
+export interface FilingHeaderIF {
```
severinbeauvais, line 4: This used to be Header Interface but now there's a different interface for Filing Header and Todo Header (even though both come in tasks from the API).

src/views/AnnualReport.vue
```diff
@@ -674,6 +674,8 @@ export default {
           }
         }
 
+        // NB: ARFilingYear, arMaxDate and arMinDate were set in the store by the TodoList component
```
severinbeauvais, line 677: This seems kind of weird but is needed when starting a new AR.

A better way might be for the dashboard to create a draft AR with the initial data, and then THIS component could always load the draft and have the data it needs... future (if ever).

src/components/Dashboard/TodoList.vue
```diff
@@ -469,7 +469,7 @@
           </template>
 
           <template v-else-if="isStatusPending(task)">
-            <PaymentPendingOnlineBanking v-if="isPayMethodOnlineBanking(task)" :filing=task />
+            <PaymentPendingOnlineBanking v-if="isPayMethodOnlineBanking(task)" :filing=task class="mb-6" />
```
severinbeauvais, line 472: This is the only one I noticed but there are probably more :(

src/components/Dashboard/TodoList.vue
```diff
@@ -979,22 +994,21 @@ export default {
         case FilingTypes.ANNUAL_REPORT:
           // resume this Annual Report
           this.setARFilingYear(task.ARFilingYear)
-          this.setArMinDate(task.arMinDate)
-          this.setArMaxDate(task.arMaxDate)
+          this.setArMinDate(task.arMinDate) // COOP only
+          this.setArMaxDate(task.arMaxDate) // COOP only
+          this.setNextARDate(task.nextArDate) // BCOMP only
```
severinbeauvais, line 999: This is actually not used (yet) because BCOMP ARs cannot be saved as drafts, and therefore cannot be resumed. But now it's ready in case BCOMP ARs do get saved/resumed :)

src/components/Dashboard/TodoList.vue
```diff
@@ -979,22 +994,21 @@ export default {
...
           this.setCurrentFilingStatus(FilingStatus.DRAFT)
           this.$router.push({ name: Routes.ANNUAL_REPORT, params: { filingId: task.id } })
           break
 
         case FilingTypes.CHANGE_OF_DIRECTORS:
           // resume this Change Of Directors
-          this.setARFilingYear(task.ARFilingYear)
```
severinbeauvais, line 990: This is not used for COD (or COA).

src/components/Dashboard/TodoList.vue
```diff
@@ -713,24 +717,28 @@ export default {
...
+          // NB: get min/max AR dates from header object (not business object)
+          // same as loading a draft AR
+          arMinDate: header.arMinDate, // COOP only
+          arMaxDate: header.arMaxDate, // COOP only
+          status: header.status || FilingStatus.NEW,
           enabled: Boolean(task.enabled),
           order: task.order,
           nextArDate: this.nextArDate(todo.business.nextAnnualReport), // BCOMP only
-          arDueDate: this.arDueDate(todo.business.nextAnnualReport) // BCOMP only
+          arDueDate: this.formatDateString(header.arMaxDate) // BCOMP only
```
andrepestana-aot, line 728: OK. Can we remove `todo.business.arMaxDate` and `todo.business.arMinDate` from the response from the legal API then?

andrepestana-aot (approved):

src/components/Dashboard/TodoList.vue
severinbeauvais: Yes. The UI does not use these.

(The API may still want to calculate these values to validate the AR date when the AR is filed.)

Do you have a ticket in mind already, or do we need a new ticket?

src/views/AnnualReport.vue
severinbeauvais: Another way would be for the task object to be passed to the Annual Report component. FWIW

src/components/Dashboard/TodoList.vue
severinbeauvais: @thorwolpert Is removing business.arMaxDate and business.arMinDate something you want to do in your ledger refactor work?

andrepestana-aot: > Do you have a ticket in mind already, or do we need a new ticket?

We don't have any ticket for that atm. That would be good.

severinbeauvais: https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/8827

eve-git (approved):
LGTM

src/components/common/NameRequestInfo.vue
cameron-freshworks: I've noticed this as well

cameron-freshworks (approved):
Oh! Already merged :)

PR #227, opened 2021-09-10
8922 update to use NR State - CONSUMED, EXPIRED

vysakh-menon-aot:
*Issue #:* /bcgov/entity#8922

*Description of changes:*

### DIFF
--- a/src/mixins/name-request-mixin.ts
+++ b/src/mixins/name-request-mixin.ts
@@ -35,12 +35,6 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
       throw new Error('getNrState() : no NR provided')
     }
 
-    // If the NR is expired, it is not consumable.
-    const expireDays = this.daysFromToday(nr.expirationDate)
-    if (isNaN(expireDays) || expireDays < 1) {
-      return NameRequestStates.EXPIRED
-    }
-
     // If the NR is awaiting consent, it is not consumable.
     // null = consent not required
     // R = consent received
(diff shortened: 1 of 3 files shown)


src/mixins/name-request-mixin.ts
```diff
@@ -51,17 +45,14 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
     }
 
     // If the NR's root state is not APPROVED or CONDITIONAL, it is not consumable.
-    if (nr.state !== NameRequestStates.APPROVED && nr.state !== NameRequestStates.CONDITIONAL) {
+    // EXPIRED or CONSUMED should not return NOT_APPROVED.
+    if (![NameRequestStates.APPROVED, NameRequestStates.CONDITIONAL,
```
cameron-freshworks, line 49: Can conditional be approved (ie conditionally approved)? is this supposed to be `! NameRequestStates.CONDITIONAL`

vysakh-menon-aot: Yes. `NameRequestStates.CONDITIONAL` means conditionally approved and it can be consumed. Is that what you are asking?

severinbeauvais (approved):

src/mixins/name-request-mixin.ts
cameron-freshworks: For sure, the question was mostly around the ! but you answered that in the other PR. 

But yea, conditional from Name Request is a little tricky because it has an accompanying property which indicates whether the Consent has been received, waived or is still REQUIRED. 

So the state could be CONDITIONAL but without consent received which would result in not quite approved yet. 
Not sure how to handle that here just some food for thought.

cameron-freshworks (approved):

PR #228, opened 2021-09-21
9049 validate NR only when IA filing is in DRAFT or PENDING status

vysakh-menon-aot:
*Issue #:* /bcgov/entity#9049

*Description of changes:*

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -594,21 +594,21 @@ export default {
     },
 
     storeNrData (nr: any, ia: any): void {
-      if (ia.filing.header.status in [FilingStatus.DRAFT, FilingStatus.PENDING]) {
-        // check if NR is valid
-        if (!this.isNrValid(nr)) {
-          this.nameRequestInvalidDialog = true
-          throw new Error('Invalid NR data')
-        }
+      // check if NR is valid
+      if (!this.isNrValid(nr)) {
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -593,26 +593,28 @@ export default {
...
     },
 
-    storeNrData (nr: any): void {
-      // check if NR is valid
-      if (!this.isNrValid(nr)) {
-        this.nameRequestInvalidDialog = true
-        throw new Error('Invalid NR data')
-      }
+    storeNrData (nr: any, ia: any): void {
+      if (ia.filing.header.status in [FilingStatus.DRAFT, FilingStatus.PENDING]) {
```
severinbeauvais, line 597: I'm trying to think what states the filing could be in, and what should be done in various cases...
Here, you've determined that we should check the NR only if the filing is Draft or Pending. What about the other states?
A couple of options to make this code "fuller": add a comment explaining the business rule here; or use a switch-case (with comments) to itemize what happens for all states. Or ...?

severinbeauvais: Or change this condition to "don't check the NR if the filing is Completed"?

What do you think?

vysakh-menon-aot: I think it should be fine to check !Completed. Will change it

severinbeauvais (approved):
Looks good to me!

PR #230, opened 2021-09-23
Update to use Nr type dynamically

cameron-freshworks:
*Issue #:* /bcgov/entity#9199

*Description of changes:*
* Small update to take/use the NR type dynamically.

### DIFF
--- a/src/mixins/name-request-mixin.ts
+++ b/src/mixins/name-request-mixin.ts
@@ -64,8 +64,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
     return {
       // workaround for old or new property name
       nrNumber: nr.nrNum || nr.nrNumber,
-      // FUTURE: Update entityType to use nr.requestTypeCd when namex supports our entity types
-      entityType: CorpTypeCd.BENEFIT_COMPANY,
+      entityType: nr.legalType,
       filingId: filingId,
       applicant: {
         // Address Information


src/mixins/name-request-mixin.ts
```diff
@@ -64,8 +64,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
     return {
       // workaround for old or new property name
       nrNumber: nr.nrNum || nr.nrNumber,
-      // FUTURE: Update entityType to use nr.requestTypeCd when namex supports our entity types
-      entityType: CorpTypeCd.BENEFIT_COMPANY,
+      entityType: nr.legalType,
```
severinbeauvais, line 67: Don't we have a table or something to convert from "type in NR" to "type in Entity" ?

severinbeauvais: Or are the types the same between namex and Lear?

severinbeauvais: Also, what's the difference between `legalType` and `requestTypeCd` ?

cameron-freshworks: ah you are right, there could be a conversion mismatch for BC. I forgot, good catch!

severinbeauvais: Or was it CR?

severinbeauvais: Lol, CR (in NR) is BC in Lear, so we're both right.

cameron-freshworks: the legalType is the entityType and the requestTypeCd as namex calls it, is the code for describing what the user wants to use the NR for afaik.
ie new business, restoration change of name etc.

severinbeauvais: https://github.com/bcgov/bcrs-shared-components/blob/06785da8bbeb1594100243a30d5639d60e0952df/src/modules/corp-type-module/corp-type-module.ts#L19

cameron-freshworks: I'm wrong, i'm not actually sure..

severinbeauvais: > the legalType is the entityType and the requestTypeCd as namex calls it, is the code for describing what the user wants to use the NR for afaik.
> ie new business, restoration change of name etc.

I looked at some sample NRs and both were the same value 🤷‍♂️

cameron-freshworks: oh no, i was right. The `requestTypeCd` is the code that existed to describe the combination of entity type, request type AND location. 
So these codes mapped to a specific type of NRs. It just so happens that the legalType code often matches the request Type code for basic NEW business requests of specific entity types, in this case a COOP.

Clear as mud eh hahaha
Basically a legacy code.

severinbeauvais: Should this be updated too?
https://github.com/bcgov/business-filings-ui/blob/da20983945e2da3e5d1bcadbcb10b83991cbf921/src/App.vue#L603

severinbeauvais:
Make sure you test some NRs (including edge cases).

src/mixins/name-request-mixin.ts
cameron-freshworks: I'm not sure, i don't believe `requestTypeCd` and `entityType` are always the same thing. I'm pretty sure if we look at requestTypeCd mappings in Namex they will describe a combination of values, where entity type is just entity type. 
I'm not too sure about uncommenting this at the moment.

cameron-freshworks: After looking at this further, we are just parsing this NR for display purposes and the conversion IS happening in computed value `entityTypeDescription` :) 
Good for bringing up though, had to check!

severinbeauvais: OK, so requestTypeCd is different from entityType.

Should that check in App.vue be uncommented, using `nr.legalType` ?

The original idea was to verify that the NR type matches the entity type. Is this guaranteed some other way? If not then this check should be enabled.

severinbeauvais: ^^ If the the check in App.vue is not needed then please delete it (and add a comment indicate why it's not needed). Eg,
> don't need to verify NR type because ...

cameron-freshworks: Hmm i really don't know. Not sure i'm comfortable with either at this point. Would need to look at it for bit longer than i have time for at this moment.

severinbeauvais: OK, go with what you have now (plus CR/BC conversion), test it well, and then make a note to look at it later.

cameron-freshworks:
> Make sure you test some NRs (including edge cases).

The tests still pass after removing the hard coded Benefit Company value 👍  
Assuming good testing was already in place, we should be fine with making this change.

severinbeauvais:
> > Make sure you test some NRs (including edge cases).
> 
> The tests still pass after removing the hard coded Benefit Company value 👍
> Assuming good testing was already in place, we should be fine with making this change.

(cringe) Maybe quickly verify that the tests even look at this.

src/mixins/name-request-mixin.ts
cameron-freshworks: That conversion is happening prior to this point, can confirm. When we pass in the NR to the mixin, it's already reflecting the correct type. 
ie a Benefit Company NR(BC) is of the correct Legal Type for Lear (BEN) prior to the point of converting to the display name.

cameron-freshworks:
> > > Make sure you test some NRs (including edge cases).
> > 
> > 
> > The tests still pass after removing the hard coded Benefit Company value 👍
> > Assuming good testing was already in place, we should be fine with making this change.
> 
> (cringe) Maybe quickly verify that the tests even look at this.

Can confirm, we have some testing tech debt.

Screen Shot 2021-09-23 at 2 59 49 PM

cameron-freshworks:
Created a ticket to capture this tech debt in our testing suites.
I propose parameterization to support all the varying entity types to come! 

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/9210

severinbeauvais (approved):
Go ahead with this (but test before merging please).

PR #263, opened 2021-12-03
10159 Disable PWA

severinbeauvais:
*Issue #:* /bcgov/entity#10159

_This is a bit experimental because it's not testable locally (PWA is not used locally). However, the cached HEAD call reproduces for me in Dev, so I will be able to verify this fix after it's merged._

*Description of changes:*
- added plugin to remove existing service workers on users' systems
- removed plugin to implement Vue PWA
- updated Vue config
- deleted obsolete manifest file
- deleted obsolete icons
- removed service worker registration
- deleted obsolete service worker scripts
- app version: 4.2.0


package.json
```diff
@@ -41,7 +41,8 @@
     "vue2-filters": "^0.10.0",
     "vuelidate": "^0.7.5",
     "vuetify": "^2.3.9",
-    "vuex": "^3.5.1"
+    "vuex": "^3.5.1",
+    "webpack-remove-serviceworker-plugin": "^1.0.0"
```
severinbeauvais, line 45: You can read up on this in the references I put in vue.config.js.

TLDR; You need to register a new Service Worker to replace the one currently on users' machines. But this new SW self-destroys. After some time (a year?) we can delete this package, since all users would have "wiped" their computers by then.

vue.config.js
```diff
@@ -11,6 +12,10 @@ const aboutText2 = (sbcName && sbcVersion) ? `${sbcName} v${sbcVersion}` : ''
 module.exports = {
   configureWebpack: {
     plugins: [
+      // this is needed to remove existing service workers on users' systems
+      // ref: https://www.npmjs.com/package/webpack-remove-serviceworker-plugin
+      // ref: https://github.com/NekR/self-destroying-sw/tree/master/packages/webpack-remove-serviceworker-plugin
+      new RemoveServiceWorkerPlugin(),
```
severinbeauvais, line 18: No filename is specified here because the default one is OK. After a `npm run build`, the dist/ folder includes file `sw.js` that is installed by the plugin.

package.json
severinbeauvais: New ticket to remove this plugin in 1 year: https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/10370

lekshmimallika-aot (approved):

package.json
cameron-freshworks: cool!

cameron-freshworks (approved):
Nothing better than a PR that REMOVES code 😁

PR #267, opened 2021-12-08
7635 Implemented Allowable Actions mixin

severinbeauvais:
*Issue #:* /bcgov/entity#7635 and /bcgov/entity#9350

*Description of changes:*

The intent of these changes is to prepare the code (framework) for when the UI gets the list of allowable actions from the API (ie, that business logic will no longer exist in the UI).

- created Allowable Actions mixin
- created Allowable Actions enum
- moved "allow" logic to new mixin
- created unit tests for new mixin
- updated other unit tests
- app version: 4.2.1

- also added commit to align Entity Info buttons vertically

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -178,8 +178,8 @@ export default {
       return sessionStorage.getItem('BUSINESS_ID')
     },
 
-    /** The Business Create URL string. */
-    createUiUrl (): string {
+    /** The Create URL string. */
+    createUrl (): string {
       return sessionStorage.getItem('CREATE_URL')
     },
 
@@ -719,7 +719,7 @@ export default {
         throw new Error('Invalid API response')
       }
 
-      const url = `${this.createUiUrl}define-dissolution?id=${this.getEntityIncNo}`
+      const url = `${this.createUrl}define-dissolution?id=${this.getEntityIncNo}`
       window.location.assign(url) // assume URL is always reachable
     },
 
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -185,7 +185,7 @@
                 </v-btn>
 
                 <!-- the drop-down menu -->
-                <v-menu offset-y left transition="slide-y-transition" v-if="isRoleStaff && !tempRegNumber">
+                <v-menu offset-y left transition="slide-y-transition" v-if="isRoleStaff && !!businessId">
                   <template v-slot:activator="{ on }">
                     <v-btn text v-on="on" class="menu-btn app-blue pa-1" click.stop>
                       <v-icon>mdi-menu-down</v-icon>
@@ -193,7 +193,7 @@
                   </template>
                   <v-list dense>
                     <v-list-item-group color="primary">
-                      <v-list-item v-if="!filing.isTypeStaff" :disabled="disableCorrection(filing)">
+                      <v-list-item :disabled="disableCorrection(filing)">
                         <v-list-item-icon>
                           <v-icon class="app-blue">mdi-file-document-edit-outline</v-icon>
                         </v-list-item-icon>
@@ -315,11 +315,11 @@
     <!-- No Results Message -->
     <v-card class="no-results" flat v-if="!historyItems.length">
       <v-card-text>
-        <template v-if="tempRegNumber">
+        <template v-if="!!tempRegNumber">
           <div class="no-results__subtitle">Complete your filing to display</div>
         </template>
 
-        <template v-else>
+        <template v-if="!!businessId">
           <div class="no-results__title">You have no filing history</div>
           <div class="no-results__subtitle">Your completed filings and transactions will appear here</div>
         </template>
@@ -348,9 +348,9 @@ import { DetailsList } from '@/components/common'
 import { AddCommentDialog, DownloadErrorDialog, LoadCorrectionDialog } from '@/components/dialogs'
 
 // Enums, Interfaces and Mixins
-import { FilingTypes, Routes } from '@/enums'
+import { AllowableActions, FilingTypes, Routes } from '@/enums'
 import { ActionBindingIF, ApiFilingIF, DocumentIF, HistoryItemIF, LegalFilingIF } from '@/interfaces'
-import { DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
+import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
 
 @Component({
   components: {
@@ -374,9 +374,8 @@ import { DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
   }
 })
 export default class FilingHistoryList extends Mixins(
-  DateMixin, EnumMixin, FilingMixin, LegalApiMixin
+  AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin, LegalApiMixin
 ) {
-  @Prop({ default: false }) readonly disableChanges: boolean
   @Prop({ default: null }) readonly highlightId: number
 
   @Getter isBComp!: boolean
@@ -407,6 +406,11 @@ export default class FilingHistoryList extends Mixins(
   /** The IA's Temporary Registration Number string. */
   private get tempRegNumber (): string {
     return sessionStorage.getItem('TEMP_REG_NUMBER')
+  }
+
+  /** The Business ID string. */
+  private get businessId (): string {
+    return sessionStorage.getItem('BUSINESS_ID')
   }
 
   private loadData (): void {
@@ -842,7 +846,7 @@ export default class FilingHistoryList extends Mixins(
   /** Whether to disable correction for this history item. */
   private disableCorrection (item: HistoryItemIF): boolean {
     return (
-      this.disableChanges ||
+      !this.isAllowed(AllowableActions.FILE_CORRECTION) ||
       item.availableOnPaperOnly ||
       item.isTypeStaff ||
       item.isFutureEffective ||
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -48,17 +48,17 @@
                 </template>
                 <v-list dense>
                     <v-list-item-group color="primary">
-                        <v-list-item @click="showRegistrarsNotationDialog()">
+                        <v-list-item @click="showRegistrarsNotationDialog()" :disabled="disabled">
                             <v-list-item-title>
                                 <span class="app-blue">Add Registrar's Notation</span>
                             </v-list-item-title>
                         </v-list-item>
-                        <v-list-item @click="showRegistrarsOrderDialog()">
+                        <v-list-item @click="showRegistrarsOrderDialog()" :disabled="disabled">
                             <v-list-item-title>
                                 <span class="app-blue">Add Registrar's Order</span>
                             </v-list-item-title>
                         </v-list-item>
-                        <v-list-item @click="showCourtOrderDialog()">
+                        <v-list-item @click="showCourtOrderDialog()" :disabled="disabled">
                             <v-list-item-title>
                                 <span class="app-blue">Add Court Order</span>
                             </v-list-item-title>
@@ -86,6 +86,9 @@ export default class StaffNotation extends Vue {
 
   /** Prop for the scrollbar offset to be added. */
   @Prop() readonly addScrollbarOffset: string
+
+  /** Prop for disabling the menu items. */
+  @Prop({ default: false }) readonly disabled: boolean
 
   showRegistrarsNotationDialog (): void {
     this.isAddingRegistrarsNotation = true
@@ -129,8 +132,8 @@ export default class StaffNotation extends Vue {
     }
 }
 
-// This class will be applied when addScrollbarOffset prop is true
-// This is necessary to align with FilingHistoryList component that may have an active scroll bar
+// This class will be applied when addScrollbarOffset prop is true.
+// This is necessary to align with FilingHistoryList component that may have an active scroll bar.
 .add-scrollbar-offset {
     overflow-y: scroll;
     scrollbar-color: transparent transparent; // FireFox uses this property
@@ -141,7 +144,7 @@ export default class StaffNotation extends Vue {
     }
 }
 
-// Fix the transparent added by .add-scrollbar-offset (Firefox only)
+// Fix the transparent added by .add-scrollbar-offset (Firefox only).
 ::v-deep .add-notation-dialog {
     scrollbar-color: auto;
 }
@@ -155,7 +158,12 @@ export default class StaffNotation extends Vue {
 #app > div.v-menu__content {
     margin: 0.625rem 0 0 0;
 }
+
 .v-btn.active .v-icon {
     transform: rotate(-180deg);
+}
+
+::v-deep .theme--light.v-list-item--disabled {
+  opacity: 0.38 !important;
 }
 </style>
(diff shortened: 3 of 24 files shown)


src/App.vue
```diff
@@ -178,8 +178,8 @@ export default {
       return sessionStorage.getItem('BUSINESS_ID')
     },
 
-    /** The Business Create URL string. */
-    createUiUrl (): string {
+    /** The Create URL string. */
+    createUrl (): string {
```
severinbeauvais, line 182: Renamed to be consistent with other places this is used.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -185,15 +185,15 @@
                 </v-btn>
 
                 <!-- the drop-down menu -->
-                <v-menu offset-y left transition="slide-y-transition" v-if="isRoleStaff && !tempRegNumber">
+                <v-menu offset-y left transition="slide-y-transition" v-if="isRoleStaff && !!businessId">
```
severinbeauvais, line 188: `tempRegNumber` and `businessId` are mutually exclusive: we either get one or the other.

I changed this line of code because I think it makes more sense to "do this if we have a businessId" (instead of "do this if we don't have a tempRegNumber").

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -185,15 +185,15 @@
...
                   <template v-slot:activator="{ on }">
                     <v-btn text v-on="on" class="menu-btn app-blue pa-1" click.stop>
                       <v-icon>mdi-menu-down</v-icon>
                     </v-btn>
                   </template>
                   <v-list dense>
                     <v-list-item-group color="primary">
-                      <v-list-item v-if="!filing.isTypeStaff" :disabled="disableCorrection(filing)">
+                      <v-list-item :disabled="disableCorrection(filing)">
```
severinbeauvais, line 196: Always render this item (but disable it) as needed.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -48,17 +48,17 @@
                 </template>
                 <v-list dense>
                     <v-list-item-group color="primary">
-                        <v-list-item @click="showRegistrarsNotationDialog()">
+                        <v-list-item @click="showRegistrarsNotationDialog()" :disabled="disabled">
```
severinbeauvais, line 51: Show this drop-down menu but disable the items as needed.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -155,7 +158,12 @@ export default class StaffNotation extends Vue {
 #app > div.v-menu__content {
     margin: 0.625rem 0 0 0;
 }
+
 .v-btn.active .v-icon {
     transform: rotate(-180deg);
 }
+
+::v-deep .theme--light.v-list-item--disabled {
+  opacity: 0.38 !important;
```
severinbeauvais, line 167: To make items look disabled. (Copied from elsewhere.)

src/components/Dashboard/TodoList.vue
```diff
@@ -658,6 +659,15 @@ export default class TodoList extends Mixins(DateMixin, EnumMixin, FilingMixin,
     return this.requiresAlteration ? 'Alter Now' : 'Resume'
   }
 
+  /** Whether the File Annual Report button should be disabled. */
+  private isFileAnnualReportDisabled (item: TodoItemIF, index: number): boolean {
+    return (
+      !item.enabled ||
+      (this.isBComp && !this.enableCheckbox[index]) ||
+      !this.isAllowed(AllowableActions.FILE_ANNUAL_REPORT)
+    )
```
severinbeauvais, line 668: Pretty much the same logic that was in the template except for `isAllowed()` (and `isCoaPending` is now part of `isAllowed` for this action).

src/components/EntityInfo.vue
```diff
@@ -222,27 +221,13 @@ export default class EntityInfo extends Mixins(CommonMixin, DateMixin, EnumMixin
...
-  /** True if Download Summary button should be rendered. */
-  private get showDownloadSummaryBtn (): boolean {
-    return getFeatureFlag('download-summary-enabled')
-  }
-
-  /** True if Dissolution button should be rendered. */
-  private get showDissolutionBtn (): boolean {
-    // is the current entity type enabled for dissolutions?
-    return getFeatureFlag('supported-dissolution-entities')?.includes(this.getEntityType)
-  }
```
severinbeauvais, line 244: These all got refactored into the mixin for the applicable action.

src/components/EntityInfo.vue
```diff
@@ -316,10 +301,7 @@ export default class EntityInfo extends Mixins(CommonMixin, DateMixin, EnumMixin
 
   /** Redirects the user to the Auth UI to update their business profile. */
   private editBusinessProfile (): void {
-    // safety checks
-    if (!this.hasBlocker && !this.isHistorical) {
```
severinbeauvais, line 320: No need for safety checks: it's now clear in the the template that we can't get here unless allowed.

src/interfaces/state-interface.ts
```diff
@@ -27,6 +27,7 @@ export interface StateIF {
   lastAnnualReportDate: string // YYYY-MM-DD
   lastAddressChangeDate: string // YYYY-MM-DD
   lastDirectorChangeDate: string // YYYY-MM-DD
+  isHistorical: boolean
```
severinbeauvais, line 30: This is probably temporary ... until we get this some other way. I added it here so I could fiddle with it in the unit tests.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Getter isUlc!: boolean
+  @Getter isHistorical!: boolean
+  @Getter isRoleStaff!: boolean
+
+  /**
+   * Returns True if the specified action is allowed, else False.
+   * @param action the action to check
+   */
+  isAllowed (action: AllowableActions): boolean {
+    const businessId = sessionStorage.getItem('BUSINESS_ID')
```
severinbeauvais, line 21: This CANNOT be a getter as it would not be reactive. I found this out the hard way.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      case AllowableActions.FILE_STAFF_NOTATION: {
+        return (!this.isHistorical && !this.hasBlocker && !!businessId && this.isRoleStaff)
+      }
+
+      case AllowableActions.VIEW_CHANGE_COMPANY_INFO: {
+        return (!this.isHistorical && (this.isBComp || this.isBcCompany || this.isUlc))
+      }
+
+      default: return null
```
severinbeauvais, line 64: I didn't want to return False so we could tell the difference between "not allowed" and "unknown" (although this is unused atm except for in the unit tests).

src/mixins/allowable-actions-mixin.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Getter getEntityType!: CorpTypeCd
+  @Getter hasBlocker!: boolean
+  @Getter isBComp!: boolean
+  @Getter isBcCompany!: boolean
+  @Getter isUlc!: boolean
+  @Getter isHistorical!: boolean
+  @Getter isRoleStaff!: boolean
+
+  /**
+   * Returns True if the specified action is allowed, else False.
```
severinbeauvais, line 17: In other words, this is the business logic for what actions are allowed. For now.

src/views/Dashboard.vue
```diff
@@ -24,7 +24,6 @@
               <LegalObligation/>
               <TodoList
                 :highlightId="filingId"
-                :disableChanges="disableChanges"
```
severinbeauvais, line 27: I moved this inside TodoList. Same with FilingHistoryList below.

src/views/Dashboard.vue
```diff
@@ -35,15 +34,15 @@
                   <span>Recent Filing History</span>&nbsp;<span class="gray6">({{historyCount}})</span>
                 </h2>
                 <StaffNotation
-                  v-if="isRoleStaff && !tempRegNumber"
+                  v-if="isRoleStaff && !!businessId"
```
severinbeauvais, line 37: = don't render if it's not applicable

tests/unit/Dashboard.spec.ts
```diff
@@ -62,39 +60,93 @@ describe('Dashboard - UI', () => {
...
-    expect(wrapper.find('#standalone-directors-button').attributes('disabled')).toBeUndefined()
+    // re-mount the component since setting session storage is not reactive
+    sessionStorage.setItem('BUSINESS_ID', 'CP1234567')
+    const localWrapper: Wrapper<Vue> = shallowMount(Dashboard, { store, vuetify, mocks: { $route } })
+    const localVm: any = localWrapper.vm
+
+    store.state.hasBlockerTask = false
+    store.state.hasBlockerFiling = false
+    store.state.isCoaPending = false
+    expect(localVm.hasBlocker).toEqual(false)
```
severinbeauvais, line 71: TLDR; What is allowed in each test here depends on more things.

tests/unit/allowable-actions-mixin.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+import * as FeatureFlags from '@/utils/feature-flags'
+import MixinTester from './mixin-tester.vue'
+
+Vue.use(Vuetify)
+
+const vuetify = new Vuetify({})
+const store = getVuexStore() as any // remove typings for unit tests
+
+// mock global function
+const getFeatureFlag = jest.spyOn(FeatureFlags as any, 'getFeatureFlag')
```
severinbeauvais, line 14: Make a note of this! It took me a while to figure out (Google was no help).

This is needed because `getFeatureFlag` is a global function -- not attached to window, global, Vue, etc.

And then you can mock the return value (or implementation)... see below (eg, line 56).

tests/unit/allowable-actions-mixin.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      { isHistorical: true, hasBlocker: false, businessId: null, expected: false },
+      // only third condition:
+      { isHistorical: true, hasBlocker: true, businessId: 'CP1234567', expected: false },
+      // all conditions:
+      { isHistorical: false, hasBlocker: false, businessId: 'CP1234567', expected: true }
+    ]
+
+    for (let test of tests) {
+      store.state.isHistorical = test.isHistorical
+      jest.spyOn(vm, 'hasBlocker', 'get').mockReturnValue(test.hasBlocker)
```
severinbeauvais, line 108: Make a note of this! It took me a while to figure out (Google was of little help).

This is how you mock a getter. ((Though in this case it's on `vm` object.)

tests/unit/EntityInfo.spec.ts
```diff
@@ -375,15 +375,19 @@ describe('EntityInfo - Click Tests - Alterations', () => {
 })
 
 describe('EntityInfo - Click Tests - Dissolutions', () => {
+  // mock the mixin to always return True
+  const AllowableActionsMixin: any = {
+    methods: {
+      isAllowed: () => true
+    }
+  }
```
severinbeauvais, line 383: Or I could have used `spyOn` to mock just the method and mocked a return value (or implementation).

Well, now we have a choice for future tests :)

tests/unit/date-mixin.spec.ts
```diff
@@ -14,7 +14,6 @@ describe('Date Mixin', () => {
 
   beforeAll(async () => {
     // mount the component and wait for everything to stabilize
-    // (this can be any component since we are not really using it)
```
severinbeauvais, line 17: Obsolete comment. Now we use MixinTester :)

severinbeauvais:
@ all : as you review today, I'm going to look again at where each button is and try to determine if we should conditionally render the button or conditionally disable the button. (Currently, except for some places I called out, the functionality is the same as before my changes.) This may require UX input. Do you have an opinion on the various buttons? lmk

src/components/Dashboard/FilingHistoryList.vue
cameron-freshworks: Right on.
I think we did the same in the create ui. A common computed value.

cameron-freshworks (commented):
Looks really good!
I like the concept of consolidating the variable disabling states in 1 location, seems super helpful and easy to manage. 
I think that definitely simplifies things!

I'm curious how the api allowable actions ties in, and if that would add a bit too much complexity. For instance, i know that Name Request has similar mechanisms to provide allowable actions from the API and at times, seems a bit more complex to maintain.

src/interfaces/state-interface.ts
lekshmimallika-aot: This will be in a property called state. The value will be 'ACTIVE' or 'HISTORICAL' . It will also have a reason property that indicates what changed this entity to historical state.You will also get the filing id that caused this. This is still pending approval from Thor.

severinbeauvais:
> Looks really good! I like the concept of consolidating the variable disabling states in 1 location, seems super helpful and easy to manage. I think that definitely simplifies things!
> 
> I'm curious how the api allowable actions ties in, and if that would add a bit too much complexity. For instance, i know that Name Request has similar mechanisms to provide allowable actions from the API and at times, seems a bit more complex to maintain.

In the RFC, Thor expounds the idea of the API using "constraint tables" and enums, like Namex, [to inform UI actions](https://github.com/bcgov/entity/pull/10366/files#r762131698). The idea -- which is good design -- is that there should be as little business logic in the UI as possible (especially if we plan to make the Legal API open to others).

I consolidated the Allowable Actions business logic with the expectation that the UI would just get a list of actions similar to Namex, and then the new mixin could be refactored accordingly (probably very easily). And if not, well, you're quite right: it definitely simplifies things. I like it :)

lekshmimallika-aot (approved):
Looks good

src/interfaces/state-interface.ts
severinbeauvais: Thanks for looking at this with the API in mind. This should be easy to refactor as needed.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: By the way, I'm using `!!` something more and more to convert from a non-boolean to a boolean. Like here, where "businessId" is a string.

tests/unit/allowable-actions-mixin.spec.ts
severinbeauvais: PS Trivia: `vm` stands for ViewModel in the [MVVM software architectural pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel).

cameron-freshworks:
> > Looks really good! I like the concept of consolidating the variable disabling states in 1 location, seems super helpful and easy to manage. I think that definitely simplifies things!
> > I'm curious how the api allowable actions ties in, and if that would add a bit too much complexity. For instance, i know that Name Request has similar mechanisms to provide allowable actions from the API and at times, seems a bit more complex to maintain.
> 
> In the RFC, Thor expounds the idea of the API using "constraint tables" and enums, like Namex, [to inform UI actions](https://github.com/bcgov/entity/pull/10366/files#r762131698). The idea -- which is good design -- is that there should be as little business logic in the UI as possible (especially if we plan to make the Legal API open to others).
> 
> I consolidated the Allowable Actions business logic with the expectation that the UI would just get a list of actions similar to Namex, and then the new mixin could be refactored accordingly (probably very easily). And if not, well, you're quite right: it definitely simplifies things. I like it :)

Cool, well i look forward to seeing the implementation of this! 
I know in Namex we have a very centralized component that utilizes the list of allowable actions, where in our filings UI, it feels* a little more dispersed. That being said, i'm sure it will work great!

cameron-freshworks (approved):

PR #269, opened 2021-12-09
Custodian accordion + global updates

cameron-freshworks:
*Issue #:* /bcgov/entity#8822 and /bcgov/entity#9282

*Description of changes:*
* Implemented Custodian of Records Accordion
* Global style updates 
* - Hidden scroll bars
* - Removed DASHBOARD title header
* Basic Getter placeholder for custodians

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "4.2.1",
+  "version": "4.2.2",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -77,8 +77,8 @@
 .scrollable-container {
   overflow-y: hidden;
   &:hover {
-  overflow-y: auto;
-}
+    overflow-y: auto;
+  }
 }
 
 // Hide scrollbar for Chrome, Safari and Opera
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -75,7 +75,7 @@
 
 // Scrollable Container
 .scrollable-container {
-  overflow-y:scroll;
+  overflow: auto;
 }
 
 // Hide scrollbar for Chrome, Safari and Opera
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -75,7 +75,10 @@
 
 // Scrollable Container
 .scrollable-container {
-  overflow: auto;
+  overflow-y: hidden;
+  &:hover {
+  overflow-y: auto;
+}
 }
 
 // Hide scrollbar for Chrome, Safari and Opera
--- a/src/components/Dashboard/AddressListSm.vue
+++ b/src/components/Dashboard/AddressListSm.vue
@@ -235,12 +235,16 @@ export default class AddressListSm extends Mixins(CommonMixin, CountriesProvince
     return (this.showCompleteYourFilingMessage || this.showGrayedOut)
   }
 
+  /** Value indicating open or closed state of accordion. */
   private get expansionValue (): Array<number> {
+    const OPEN_PANEL = 0
+    const CLOSE_PANEL = 1
+
     if (this.isHistorical) {
-      return []
+      return [] // All panels closed by default
     } else if (this.disabled) {
-      return [0, 1]
-    } else return [0]
+      return [OPEN_PANEL, CLOSE_PANEL]
+    } else return [OPEN_PANEL]
   }
 }
 </script>
--- a/src/components/Dashboard/AddressListSm.vue
+++ b/src/components/Dashboard/AddressListSm.vue
@@ -1,7 +1,7 @@
 <template>
   <div id="address-list-sm">
     <!-- when "disabled", expand all panels and disable expansion -->
-    <v-expansion-panels accordion multiple :value="disabled ? [0,1] : [0]" :disabled="disabled">
+    <v-expansion-panels accordion multiple :value="expansionValue" :disabled="disabled">
       <!-- Registered Office -->
       <v-expansion-panel id="registered-office-panel"
         class="align-items-top"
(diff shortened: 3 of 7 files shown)


src/assets/styles/layout.scss
```diff
@@ -75,7 +75,7 @@
 
 // Scrollable Container
 .scrollable-container {
-  overflow-y:scroll;
+  overflow: auto;
```
cameron-freshworks, line 78: I couldn't stand those scroll bars visible on elements that didn't require scrolling
This will display them if they are required, hide them if not.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "4.2.0",
+  "version": "4.2.1",
```
severinbeauvais, line 3: We're already at 4.2.1.

severinbeauvais: Please also include lock file :)

src/views/Dashboard.vue
```diff
@@ -50,7 +46,18 @@
           </v-col>
 
           <v-col cols="12" md="3" style="position: relative">
-            <!-- FUTURE: add "Liquidator or Custodian of Records" here -->
+            <section v-if="isHistorical && getCustodians.length >= 1">
```
cameron-freshworks, line 49: A means of hiding this until we actually set and use Custodians, which would only be in the event of a future historical state.

package.json
cameron-freshworks: oh, i'm behind locally!

src/assets/styles/layout.scss
severinbeauvais: What happens to horizontal scrolling on narrow screens?

src/components/Dashboard/AddressListSm.vue
```diff
@@ -226,13 +226,22 @@ export default class AddressListSm extends Mixins(CommonMixin, CountriesProvince
...
   private get disabled (): boolean {
     return (this.showCompleteYourFilingMessage || this.showGrayedOut)
   }
+
+  private get expansionValue (): Array<number> {
+    if (this.isHistorical) {
+      return []
+    } else if (this.disabled) {
+      return [0, 1]
+    } else return [0]
```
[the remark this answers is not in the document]
severinbeauvais: Or you could self-document it using constants.

Example:
https://github.com/bcgov/namerequest/blob/b116b9bcb389679760d052d35b0cdd9a45c91636/src/components/dialogs/renew.vue#L101

src/components/Dashboard/CustodianListSm.vue
```diff
@@ -0,0 +1,139 @@
+<template>
+  <div id="custodian-list-sm">
```
severinbeauvais, line 2: Consider a v-if here to display nothing if the custodians array is empty. (Or it that already checked in the parent?)

severinbeauvais: I see you already check that in the Dashboard, so all good.

src/assets/styles/layout.scss
cameron-freshworks: Good call, will check. 

If you go down to a mobile size, a horizontal bar will display. 
I could `overflow-y: auto` & `overflow-x: hidden` if you think this is something we should support.

src/components/Dashboard/AddressListSm.vue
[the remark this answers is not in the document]
cameron-freshworks: Hmm, i do like that for readability 👍 
Thanks!

src/assets/styles/layout.scss
severinbeauvais: Weird, I don't get horiz scrollbars when I make Chrome super narrow.

I'm thinking what you have now is OK.

Actually, there might be a ticket for hiding the scrollbars. Check Zenhub.

cameron-freshworks: Cool, 2 birds with 1 stone!

cameron-freshworks: Updated 👍  Moved #9282

src/assets/styles/layout.scss
```diff
@@ -75,7 +75,10 @@
 
 // Scrollable Container
 .scrollable-container {
-  overflow-y:scroll;
+  overflow-y: hidden;
+  &:hover {
+  overflow-y: auto;
```
severinbeauvais, line 80: indent

src/components/Dashboard/AddressListSm.vue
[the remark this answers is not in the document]
severinbeauvais: Your change is OK with me.
You can resolve this conversation if you like (since you created it).

src/views/Dashboard.vue
```diff
@@ -9,13 +9,9 @@
...
     <v-container id="dashboard-container" class="view-container">
       <article id="dashboard-article">
-        <header>
-          <h1 data-test-id="dashboard-title">Dashboard</h1>
-        </header>
-
         <v-row>
           <v-col cols="12" md="9">
-            <section>
+            <section v-if="!isHistorical">
```
severinbeauvais, line 14: Where did you see that the Todo List is hidden when the entity is historical?

src/store/getters.ts
```diff
@@ -166,5 +166,10 @@ export default {
   /** The entity registered office address. */
   getRegisteredOfficeAddress (state: StateIF): OfficeAddressIF {
     return state.registeredAddress
+  },
+
+  /** The custodians list. */
+  getCustodians (state: StateIF): any[] {
+    return []
```
severinbeauvais, line 173: When this gets implemented, we might want to make sure it's always at least `[]` (never null) in the store. This will allow us to confidently check its length without conditional chaining (which doesn't work in the template anyway).
There are already some arrays in the store that are done this way. In my PR, I've done complianceWarnings this way as I think it makes for cleaner code.

src/assets/styles/layout.scss
cameron-freshworks: Fixed next commit

src/views/Dashboard.vue
cameron-freshworks: https://projects.invisionapp.com/share/XY11SF2SDP27#/screens/458580685

src/store/getters.ts
cameron-freshworks: For sure, definitely agree! We can initialize it as an empty array. 
Will do this when we retrieve and apply the information to store :)

src/views/Dashboard.vue
cameron-freshworks: Screen Shot 2021-12-10 at 10 40 10 AM

Sorry, in the event the design is locked behind request access.

severinbeauvais: Perfect, thanks.

severinbeauvais (approved):

severinbeauvais:
In part 2 of this ticket, please create unit tests for the new component.

cameron-freshworks:
> In part 2 of this ticket, please create unit tests for the new component.

100%, will do.

PR #272, opened 2021-12-17
Updated Dashboard.vue

severinbeauvais:
*Issue #:* None

*Description of changes:*
- removed extra 2.25rem whitespace at the top of the dashboard

Screenshot:

image

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -9,7 +9,7 @@
 
     <v-container id="dashboard-container" class="view-container">
       <article id="dashboard-article">
-        <v-row>
+        <v-row class="mt-n9">
           <v-col cols="12" md="9">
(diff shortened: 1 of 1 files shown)


severinbeauvais:
This change can go in any time, which is why I didn't update the app version.

src/views/Dashboard.vue
```diff
@@ -9,7 +9,7 @@
 
     <v-container id="dashboard-container" class="view-container">
       <article id="dashboard-article">
-        <v-row>
+        <v-row class="mt-n9">
```
severinbeauvais, line 12: This was simpler than removing the top margin on the sections in each column (some of which are conditional) (and they use common CSS, so it would be weird to override that IMO).

cameron-freshworks (approved):
👍

PR #273, opened 2021-12-20
Implements Shared BreadCrumb

cameron-freshworks:
*Issue #:* /bcgov/entity#10232

*Description of changes:*
* Implements Shared BreadCrumb
* Added default routes
* Removed obsolete code/tests


lekshmimallika-aot (approved):

src/resources/breadcrumb-resources.ts
```diff
@@ -0,0 +1,11 @@
+import { BreadcrumbIF } from '@/interfaces'
+
+export const dashboardBreadcrumb: BreadcrumbIF = {
+  text: 'BC Registries Dashboard',
+  href: `${sessionStorage.getItem('BUSINESSES_URL')}business`
+}
+
+export const staffDashboardBreadcrumb: BreadcrumbIF = {
+  text: 'Staff Dashboard',
+  href: `${sessionStorage.getItem('BUSINESSES_URL')}staff`
```
severinbeauvais, line 10: This URL doesn't work for me. It's this, right?
https://dev.bcregistry.ca/business/staff

src/interfaces/breadcrumb-interface.ts
```diff
@@ -1,7 +1,5 @@
 export interface BreadcrumbIF {
```
severinbeauvais, line 1: interfaces/index.ts should import the shared IF instead of using this file, right?

src/App.vue
```diff
@@ -69,6 +69,7 @@
     <div class="app-body">
       <!-- only show pages while signing in or once the data is loaded -->
       <main v-if="isSigninRoute || dataLoaded">
+        <BreadCrumb :breadcrumbs="breadcrumbs" />
```
severinbeauvais, line 72: You know how words sometimes look funny? "crumb" is doing it for me atm 🤣

src/App.vue
```diff
@@ -93,6 +94,7 @@ import { updateLdUser } from '@/utils'
 import PaySystemAlert from 'sbc-common-components/src/components/PaySystemAlert.vue'
 import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
 import SbcFooter from 'sbc-common-components/src/components/SbcFooter.vue'
+import { BreadCrumb } from '@bcrs-shared-components/bread-crumb'
```
severinbeauvais, line 97: Not a big deal, but see if you can move this import to src/components/common/index.ts and then, here, import from there. (Purpose is to remove external imports from our components.)

src/resources/breadcrumb-resources.ts
severinbeauvais: Oops, no, it's this (which works): https://dev.bcregistry.ca/business/auth/staff

src/interfaces/breadcrumb-interface.ts
cameron-freshworks: mmm yea! You are right, we have access to that now if i update the imported IF version.

src/App.vue
cameron-freshworks: haha i know, it's such a weird word. I'm not sure if it's meant to be one or two words..

src/App.vue
cameron-freshworks: 100% i was looking at doing that but there was no default index export for the components so i didn't want to break the mould per se of Filings UI. 
Can definitely add that in!

severinbeauvais: That's why I thought "common" might be OK.

cameron-freshworks: Definitely, good call. Will do!

cameron-freshworks: Done in next commit

src/interfaces/breadcrumb-interface.ts
cameron-freshworks: oof, when i update the IF's version something goes gravely wrong with todolist and todolist.spec interfaces. Looking into it 👀

cameron-freshworks: I have no idea the relation to the shared components Interfaces and the TODO Component.

severinbeauvais: These?
https://github.com/bcgov/business-filings-ui/blob/fdf168ad3036f7459bb601c398d5532724081803/src/components/Dashboard/TodoList.vue#L541

cameron-freshworks: It's specifically the ConfirmDialogType, it is somehow breaking that component. 
When the shared Interfaces are used in a ts file, no issues, but when used directly on a Vue Component causing issues.

severinbeauvais: This is tech debt associated with the updated shared interfaces ... I sympathize ... but it would be better to not have multiple copies of this IF and to not carry forward more tech debt.

severinbeauvais: Weird, the ConfirmDialogType is the same in Filings UI as in the shared library.

cameron-freshworks: I'm wondering if it's related to the Vue version used by the shared interface vs our vue version locally

severinbeauvais: I'd be surprised. Basically Filings UI imports the code for the shared component and compiles it using Filing UI's Vue libraries.

severinbeauvais (approved):

src/interfaces/breadcrumb-interface.ts
cameron-freshworks: Updated Enum version as well 👍

PR #276, opened 2021-12-23
Parties Requests, Custodian Population and Misc Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#10474

*Description of changes:*
* Refactored Directors request to parties request
* Populate store with Custodians data
* Assigned named route/path to Dashboard
* Updated roles enum to be in sync with back end
* Update fixes 'Completing Party' sometimes appearing in the Directors list as duplicate
* Updated / Added unit testing


src/App.vue
```diff
@@ -633,14 +653,15 @@ export default {
...
-      // set directors
-      // (ie, parties that have a role of Director)
-      const directors = incorporationApplication.parties?.filter(party =>
-        party.roles.filter(role =>
-          role.roleType === Roles.DIRECTOR
-        ).length !== 0
+      // Format party roles
+      incorporationApplication.parties.forEach(
+        party => party.roles?.forEach(
+          (role, index) => { party.roles[index] = role.roleType?.toLowerCase() }
```
cameron-freshworks, line 659: This is required, as the `roleType` value we receive from IA filings has a capitalized first letter, ie `Completing Party` but when requesting the role type directly from the parties, we receive all lower case. 
This is also why i refactored our ROLES enum, to be in alignment with the back end.
The parsing of filing json is much more the exception than fetching Directors or Custodians information in the Filings UI imo.

To elaborate, this functionality called here is a much rarer occurrence, in my experience it is only called after a user completes an IA and returns the filings UI with that TEMP REG, and this allows us to populate the DASHBOARD with that filing json.

src/App.vue
```diff
@@ -731,18 +752,29 @@ export default {
...
 
-    storeDirectors (response: any): void {
-      const directors = response?.data?.directors
-      if (directors) {
-        const directorsSorted = directors.sort(this.fieldSorter(['lastName', 'firstName', 'middleName']))
-        for (var i = 0; i < directorsSorted.length; i++) {
-          directorsSorted[i].id = i + 1
-          directorsSorted[i].isNew = false
-          directorsSorted[i].isDirectorActive = true
+    storeParties (response: any): void {
```
cameron-freshworks, line 755: Updated to handle both custodians and directors. 
This also filters our some instances where completing parties were returned as duplicates of the directors in the directors request.

src/enums/roles.ts
```diff
@@ -1,6 +1,7 @@
 export enum Roles {
-  COMPLETING_PARTY = 'Completing Party',
-  INCORPORATOR = 'Incorporator',
-  DIRECTOR = 'Director',
-  SUBSCRIBER = 'Subscriber',
+  COMPLETING_PARTY = 'completing party',
+  INCORPORATOR = 'incorporator',
+  DIRECTOR = 'director',
+  SUBSCRIBER = 'subscriber',
+  CUSTODIAN = 'custodian'
```
cameron-freshworks, line 6: As commented above, this is to keep in alignment with the data in our database vs filing json values.

src/routes.ts
```diff
@@ -114,6 +114,6 @@ export default [
   {
     // default/fallback route
     path: '*',
-    redirect: '/'
+    redirect: '/dashboard'
```
cameron-freshworks, line 117: We refer to dashboard in a few spots but haven't defined the route as such. Therefore navigation to `dashboard` was always caught by the fallback route and subsequently root. 
Since this UI has several named sub-roots, it would be a good idea to be explicit about our dashboard.

cameron-freshworks:
I chose to refactor the Directors request in this PR as it made more sense to leave less TechDebt. 
If there is a good reason to keep Directors request in parallel with Parties than all good. 
I did notice there was a Directors request in our Legal API mixin as well as our Directors filing component.. so i thought it might be good to clean up a portion of it here.

lekshmimallika-aot (approved):
Looks good

src/App.vue
```diff
@@ -731,18 +751,29 @@ export default {
...
+      const parties = response?.data?.parties
+      if (parties) {
+        const directorsList = parties?.filter(partiesArr => partiesArr.roles?.includes(Roles.DIRECTOR))
```
severinbeauvais, line 757: FYI, inside the filter call, the variable is single element from the array, in this case just a "party".
No need to change.

cameron-freshworks: oh no, thats a good call i will update. 
I was wondering what to call it when trying to differentiate from parties.

src/App.vue
```diff
@@ -731,18 +751,29 @@ export default {
...
+            directorsListSorted[i].id = i + 1
+            directorsListSorted[i].isNew = false
+            directorsListSorted[i].isDirectorActive = true
+          }
+          this.setDirectors(directorsListSorted)
+        }
+
+        // Custodians
+        if (custodianList) {
+          this.setCustodians(custodianList)
```
severinbeauvais, line 773: Instead of 3 different arrays, would it make sense to store just the parties and then provide getters to extract just the directors or custodians?

src/components/Dashboard/CustodianListSm.vue
```diff
@@ -67,11 +67,19 @@ import { Component, Mixins, Prop } from 'vue-property-decorator'
...
 export default class CustodianListSm extends Mixins(CommonMixin, CountriesProvincesMixin) {
-  @Prop({ default: [] })
-  readonly custodians: Array<object>
+  @Prop({ default: () => [] })
+  readonly custodians: Array<PartiesIF>
+
+  /** Return the officer person or org name */
+  private officerName (party: PartiesIF): string {
+    return party.officer.firstName
+      ? `${party.officer.firstName} ${party.officer.lastName}`
```
severinbeauvais, line 80: Any thoughts about whether middle initial should be in here, too? I think there's an outstanding ticket for that.

src/routes.ts
severinbeauvais: That was there so the root URL would be the dashboard (not "/dashboard" in the brower URL). Does this still work?

src/App.vue
cameron-freshworks: Possibly
The directors are structured a bit different, with some additional properties assigned locally as you can see above. Which is why we also preserved that existing directors interface as well. 
They could be stored in the same place but the functionality to isolate the types between each other has to be done somewhere i suppose, so it makes sense to do it up front.

tests/unit/App.spec.ts
```diff
@@ -1344,7 +1365,7 @@ describe('App as a PAID (pending) Incorporation Application', () => {
     expect(vm.$store.getters.getEntityName).toBe('My Name Request')
   })
 
-  it('fetches IA filing properly', () => {
+  it.only('fetches IA filing properly', () => {
```
severinbeauvais, line 1368: Don't commit this :)

src/components/Dashboard/CustodianListSm.vue
cameron-freshworks: Good question, this is sort of a port from DirectorsListSm template. 
Is the middle initial something we want to include for both Directors and (Person) Custodians?

tests/unit/App.spec.ts
cameron-freshworks: Oh no!

src/App.vue
severinbeauvais: OK.

src/components/Dashboard/CustodianListSm.vue
severinbeauvais: Search for that other ticket. If you can't find it then the answer is no. Otherwise you could kill 2 🐦 with one change.

tests/unit/App.spec.ts
cameron-freshworks: I was wondering why the code cov was unhappy..

src/routes.ts
cameron-freshworks: It would display as `BC0870718/dashboard` by naming the route. 
I can remove this though, will look into this a little deeper, as i'm worried the fallback route is called vs our root route sometimes. Trying to make sense of it yet :)

src/App.vue
cameron-freshworks: updated!

src/components/Dashboard/CustodianListSm.vue
cameron-freshworks: I found ticket#: 10452, but it's pretty vague. Either way, seems like there is a desire to include middle names, in this case the middle initial. Will try to fit that in today!

tests/unit/App.spec.ts
cameron-freshworks: Fixed in next commit

src/components/Dashboard/CustodianListSm.vue
cameron-freshworks: Put the middle name into the directors accordion as well as custodians accordion. 
1 snag though, currently the custodians party object is only returning first and last name, when i explicitly put it into the filing. It might not be passed through at some point in the Dissolutions flow perhaps?
Either way, something to look for there, but this UI will support/display it once we have that value in the response.

cameron-freshworks:
@severinbeauvais Ive made all the updates recommended and with both of us being away next week, i will proceed with the merge and proceed forward with the ticket. That being said, if you have any further thoughts on the code going in, please don't hesitate to reach out and we can update!

src/routes.ts
cameron-freshworks: Removed this 👍

src/App.vue
severinbeauvais: At some point, probably a good idea to restore the conditional chaining on role type.

PR #278, opened 2022-01-06
9530 Fixed filing id not saved on pay error

severinbeauvais:
*Issue #:* /bcgov/entity#9530

*Description of changes:*
- fixed filing id not saved on pay error, which causes retry to fail
- app version = 4.3.2

### DIFF
--- a/src/views/AnnualReport.vue
+++ b/src/views/AnnualReport.vue
@@ -795,6 +795,12 @@ export default class AnnualReport extends Mixins(
       return error?.response?.data?.filing || null
     })
 
+    const filingId = +this.savedFiling?.header?.filingId || 0
+    if (filingId > 0) {
+      // save filing ID for possible future updates
+      this.filingId = filingId
+    }
+
     // if there was no error, finish save process now
     // otherwise, dialog may finish this later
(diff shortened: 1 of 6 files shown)


src/views/AnnualReport.vue
```diff
@@ -803,13 +809,10 @@ export default class AnnualReport extends Mixins(
...
 
   onClickSaveFinish (): void {
-    const filingId = +this.savedFiling?.header?.filingId
     // safety check
-    if (filingId > 0) {
+    if (this.filingId > 0) {
       // changes were saved, so clear flag
       this.haveChanges = false
-      // save filing ID for future updates
-      this.filingId = filingId
```
severinbeauvais, line 812: In the old code, `this.filingId` was only updated when the "finish" part of the step ran.

The bug is that we want to save the new filingId even if we get a save/pay error, so that we can retry using the new filingId.

severinbeauvais, line 812: In the old code, `this.filingId` was only updated when the "finish" part of the step ran.

The bug fix is that we want to save the new filingId even if we get a save/pay error, so that we can retry using the new filingId (which becomes a PUT instead of a POST).

argush3 (approved):
LGTM

PR #279, opened 2022-01-06
Added dependency package so it doesn't need to be installed globally

severinbeauvais:
*Issue #:* None

*Description of changes:*

Depending on which UI projects we work on, a different version of Node may be required. Multiple versions can be managed using [Node Version Manager](https://github.com/nvm-sh/nvm).

However, previously we may have had some packages installed globally, resulting in this warning when installing nvm:
```
=> You currently have modules installed globally with `npm`. These will no
=> longer be linked to the active version of Node when you install a new node
=> with `nvm`; and they may (depending on how you construct your `$PATH`)
=> override the binaries of modules installed with `nvm`:

/usr/local/lib
├── @vue/cli@4.5.13
├── jshint@2.13.0
├── lerna@4.0.0
├── tshint@0.0.2
└── typescript@4.3.5
=> If you wish to uninstall them at a later point (or re-install them under your
=> `nvm` Nodes), you can remove them from the system Node as follows:

     $ nvm use system
     $ npm uninstall -g a_module
```

After uninstalling all global packages, they can be reinstalled locally (as dev dependencies in package.json). I did some testing locally and found that I wasn't using `jshint` or `tshint`, so I simply didn't reinstall those. `lerna` is already specified in the package.json for bcrs-shared-components (the only project that uses it as far as I know). And `typescript` is already specified in the package.json for the various UI projects. Here is what my system looks like now:
```
$ nvm use system
Now using system version of node: v10.19.0 (npm v6.14.4)

$ npm list -g --depth 0
/usr/local/lib
└── (empty)

$ nvm use 12

$ npm list -g --depth 0
/home/severin/.nvm/versions/node/v12.22.8/lib
└── npm@6.14.15
```

Hence, this PR is to install `@vue/cli` as a dev dependency for THIS project. (I may also need to do this for the other UI projects but I'll cross those bridges when I get there.)

**Also note that THIS project can be built/linted/tested using Node v12 (as per example above), so I encourage you to update when you get a chance.**

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0)._

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -37,7 +37,7 @@ http {
    add_header Pragma "no-cache";
 
    # Content Security Policy
-   add_header Content-Security-Policy "default-src 'self'; frame-src 'self' *.gov.bc.ca *.hotjar.com; script-src 'self' 'unsafe-eval' *.gov.bc.ca *.hotjar.com; style-src 'self' 'unsafe-inline' *.cloudflare.com; font-src 'self' *.gov.bc.ca *.hotjar.com *.cloudflare.com; img-src 'self' *.postescanada-canadapost.ca data:; connect-src 'self' *.gov.bc.ca *.launchdarkly.com *.hotjar.com *.postescanada-canadapost.ca; manifest-src 'self';";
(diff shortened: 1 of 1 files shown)


severinbeauvais:
### How to Install and Use NVM

`curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash`

Then you can install different version of node like this:
`nvm install 12`

You can list your installed versions like this:
`nvm list`

You can change to a different (installed) version like this:
`nvm use 12`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "4.3.2",
+  "version": "4.3.3",
```
severinbeauvais, line 3: It's good practise to update this in every PR, to more easily track features or breakages 😉

argush3 (approved):

vysakh-menon-aot (approved):

severinbeauvais:
Anyone opposed to merging this?

PR #280, opened 2022-01-14
10599_add_security_headers

ozamani9:
*Issue #:* /bcgov/entity#10599

*Description of changes:*

Updated the nginx.conf file to include security header information (i.e. Content Security Policy, etc.)

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -14,8 +14,31 @@ http {
  fastcgi_temp_path     /tmp/fastcgi_temp;
  uwsgi_temp_path       /tmp/uwsgi_temp;
  scgi_temp_path        /tmp/scgi_temp;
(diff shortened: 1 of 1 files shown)


rarmitag (approved):

nginx.conf
```diff
@@ -36,7 +59,6 @@ http {
    }
 
    location @index {
-     add_header Cache-Control no-cache;
```
severinbeauvais, line 39: @ozamani9 Why did you remove this?

ozamani9: it was added in the header section instead.  so its like i moved it, not removed it.. does that make sense

image

severinbeauvais: Yes, perfect.

PR #281, opened 2022-01-14
Connect Business Summary

cameron-freshworks:
*Issue #:* /bcgov/entity#10619 & /bcgov/entity#10150

*Description of changes:*
* Connects Download business summary btn to request for document
* Adds unit tests for btn
* Relocate summary btn to present for all business states (other than Temp Reg), ie active & historical
* Remove the 'Goto Dashboard' Btn and functionality from the pending filing in the filing history list

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -840,7 +840,7 @@ export default {
       this.showSpinner = true
       const summaryDocument: DocumentIF = {
         title: 'Summary',
-        filename: `${this.businessId} - Summary - ${this.getCurrentDate}.pdf`,
+        filename: `${this.businessId} Summary - ${this.getCurrentDate}.pdf`,
         link: `businesses/${this.businessId}/documents/summary`
       }
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -8,6 +8,12 @@
       attach="#app"
     />
 
+    <DownloadErrorDialog
+      :dialog="downloadErrorDialog"
+      @close="downloadErrorDialog=false"
+      attach="#app"
+    />
+
     <BusinessAuthErrorDialog
       :dialog="businessAuthErrorDialog"
       @exit="onClickExit()"
@@ -73,6 +79,7 @@
         <EntityInfo
           @confirmDissolution="confirmDissolutionDialog = true"
           @notInGoodStanding="nigsMessage = $event; notInGoodStandingDialog = true"
+          @downloadBusinessSummary="downloadBusinessSummary()"
         />
         <router-view />
       </main>
@@ -102,6 +109,7 @@ import {
   BusinessAuthErrorDialog,
   ConfirmDissolution,
   DashboardUnavailableDialog,
+  DownloadErrorDialog,
   NameRequestAuthErrorDialog,
   NameRequestInvalidDialog,
   NotInGoodStandingDialog
@@ -121,7 +129,7 @@ import {
   LegalApiMixin,
   NameRequestMixin
 } from '@/mixins'
-import { ApiFilingIF, ApiTaskIF, BreadcrumbIF, BusinessIF, TaskTodoIF } from '@/interfaces'
+import { ApiFilingIF, ApiTaskIF, BreadcrumbIF, BusinessIF, DocumentIF, TaskTodoIF } from '@/interfaces'
 import {
   CorpTypeCd,
   DissolutionTypes,
@@ -146,6 +154,7 @@ export default {
   data () {
     return {
       confirmDissolutionDialog: false,
+      downloadErrorDialog: false,
       dataLoaded: false,
       dashboardUnavailableDialog: false,
       businessAuthErrorDialog: false,
@@ -178,6 +187,7 @@ export default {
     BreadCrumb,
     ConfirmDissolution,
     DashboardUnavailableDialog,
+    DownloadErrorDialog,
     BusinessAuthErrorDialog,
     NameRequestAuthErrorDialog,
     NameRequestInvalidDialog,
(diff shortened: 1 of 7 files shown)


src/App.vue
```diff
@@ -823,6 +833,23 @@ export default {
...
       }
+    },
+
+    /** Request and Download Business Summary Document. */
+    async downloadBusinessSummary (): Promise<void> {
+      this.showSpinner = true
+      const summaryDocument: DocumentIF = {
+        title: 'Summary',
+        filename: `${this.businessId} - Summary - ${this.getCurrentDate}.pdf`,
+        link: `businesses/${this.businessId}/documents/summary`
```
cameron-freshworks, line 844: With the introduction of the summary document, a new endpoint was created `businesses/${this.businessId}/documents` which exists in parallel with our filings documents endpoint.
ie `businesses/${this.businessId}/filings/${someFilingId}/documents`. 

The response of the new endpoint returns `{ summary: 'url' }` where the url is essentially the same url as the new one listed above but with `/summary` tacked onto the end, so i opted to just make this method specific and retrieve the summary document directly. 

IF in the future they decide to include more document/document types on the business documents endpoint, we would need 2 requests. I opted not to make 2 requests for a single document in this instance.

cameron-freshworks: To elaborate, the 1 request would be to retrieve all the document links. As it stands right now, it's a single object containing the `summary` key, with value `url/summary`.

cameron-freshworks:
SonarCloud is unhappy with code duplication on 4 files i did not touch.. so not sure 🤷

src/App.vue
severinbeauvais: I'm OK with this.

src/components/EntityInfo.vue
```diff
@@ -286,10 +285,9 @@ export default class EntityInfo extends Mixins(AllowableActionsMixin, CommonMixi
     window.location.assign(this.businessProfileUrl) // assume URL is always reachable
   }
 
-  /** Downloads business summary PDF. */
-  private downloadBusinessSummary (): void {
-    // FUTURE
-  }
+  // Pass prompt downloads business summary event to parent. */
+  @Emit('downloadBusinessSummary')
+  private downloadBusinessSummary (): void {}
```
cameron-freshworks, line 290: Just thinking ahead here, eventually our EntitiesInfo will be shared, so the less functionality it does the better. 
Hence, why we are having the parent (App in this case) tackle the request.

src/App.vue
```diff
@@ -823,6 +833,23 @@ export default {
...
         this.fetchData()
       }
+    },
+
+    /** Request and Download Business Summary Document. */
+    async downloadBusinessSummary (): Promise<void> {
+      this.showSpinner = true
+      const summaryDocument: DocumentIF = {
+        title: 'Summary',
+        filename: `${this.businessId} - Summary - ${this.getCurrentDate}.pdf`,
```
severinbeauvais, line 843: Remove hyphen between Business ID and title.

Examples:
CP1002494 Certificate - 2022-01-13.pdf
CP1002494 Receipt - 2022-01-13.pdf

src/components/EntityInfo.vue
```diff
@@ -31,7 +31,7 @@
               <v-chip class="primary mt-n1 ml-4 pointer-events-none" small label text-color="white">
                 <strong>HISTORICAL</strong>
               </v-chip>
-              <span class="font-14 ml-3">{{reasonText || 'Unknown Reason'}}</span>
+              <span class="font-14 ml-3 mr-3">{{reasonText || 'Unknown Reason'}}</span>
```
severinbeauvais, line 34: Might as well use `mx-3" then 😁

lekshmimallika-aot (approved):

tests/unit/EntityInfo.spec.ts
```diff
@@ -416,3 +416,43 @@ describe('EntityInfo - Click Tests - Dissolutions', () => {
...
   })
 })
+
+describe('EntityInfo - Click Tests - Business Summary', () => {
+  // mock the mixin to always return True
+  const AllowableActionsMixin: any = {
+    methods: {
+      isAllowed: () => true
+    }
+  }
```
severinbeauvais, line 426: This is one way to do it (globally) and it's fine here.

You could also do this if you need to control the return value:
https://github.com/bcgov/business-filings-ui/blob/40a77d9928b713c0c6dee7da600ba6e1bd6d8fcd/tests/unit/FilingHistoryList.spec.ts#L369

severinbeauvais:
> SonarCloud is unhappy with code duplication on 4 files i did not touch.. so not sure 🤷

See https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/10572

src/components/EntityInfo.vue
severinbeauvais: If we share the EntityInfo component then we'll probably need to add support for a template slot for the buttons anyway (since not all apps will need the same buttons).

I'm OK with this approach for now. The risk is making App.vue too large. I think we're there now but I don't have a slick way to rearchitect it.

severinbeauvais (approved):
Just a couple of small things otherwise LGTM.

src/components/EntityInfo.vue
cameron-freshworks: Ah yea, i like that idea a lot.

src/App.vue
cameron-freshworks: ty, will do!

src/components/EntityInfo.vue
cameron-freshworks: 🤦  Was staring me right the face!

tests/unit/EntityInfo.spec.ts
cameron-freshworks: I like it!

src/App.vue
cameron-freshworks: > > SonarCloud is unhappy with code duplication on 4 files i did not touch.. so not sure 🤷
> 
> See https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/10572

ahhh ok, so this isn't the first time. Thank you for sharing!

cameron-freshworks:
> > SonarCloud is unhappy with code duplication on 4 files i did not touch.. so not sure 🤷
> 
> See https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/10572

Ah ok, so this isn't the first time. Thank you for sharing!

tests/unit/EntityInfo.spec.ts
severinbeauvais: No need to change this.

The mixin already has unit tests that verify its response to feature flags. THIS component is one level up so it can just trust that the mixin is working correctly.

cameron-freshworks: Copy that, ty!

PR #288, opened 2022-02-01
10853 Login and logout redirects

severinbeauvais:
*Issue #:* /bcgov/entity#10853

*Description of changes:*
- initialize token service in main instead of App (needed to sync session before routing)
- redirect to BC Registry for login, then return here
- after logout, redirect to login page
- app version = 4.3.6

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "4.3.5",
+  "version": "4.3.6",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
--- a/src/App.vue
+++ b/src/App.vue
@@ -93,7 +93,6 @@
 <script lang="ts">
 // Libraries
 import { mapActions, mapGetters } from 'vuex'
-import KeycloakService from 'sbc-common-components/src/services/keycloak.services'
 import * as Sentry from '@sentry/browser'
 import { updateLdUser } from '@/utils'
 
@@ -173,9 +172,6 @@ export default {
 
       /** Whether to show the alternate loading spinner. */
       showSpinner: false,
-
-      /** Whether the token refresh service is initialized. */
-      tokenService: false,
 
       /** Currently supported entity types in Filings UI. */
       supportedEntityTypes: [
@@ -298,7 +294,6 @@ export default {
     if (!this.isAuthenticated) return
 
     // ...otherwise proceed
-    await this.startTokenService()
     await this.fetchData()
   },
 
@@ -318,36 +313,6 @@ export default {
       'setLastAnnualReportDate', 'setNameRequest', 'setLastAddressChangeDate', 'setLastDirectorChangeDate',
       'setConfigObject', 'setReasonText', 'setEntityState', 'setAdminFreeze', 'setComplianceWarnings',
       'setGoodStanding']),
-
-    /** Starts token service to refresh KC token periodically. */
-    async startTokenService (): Promise<void> {
-      // only initialize once
-      // don't start during Jest tests as it messes up the test JWT
-      if (this.tokenService || this.isJestRunning) return Promise.resolve()
-
-      try {
-        console.info('Starting token refresh service...') // eslint-disable-line no-console
-        await KeycloakService.initializeToken()
-        this.tokenService = true
-      } catch (error) {
-        // happens when the refresh token has expired in session storage
-
-        // eslint-disable-next-line no-console
-        console.log('Error initializing token refresher =', error)
-
-        // clear old session variables and reload page to get new tokens
-        this.clearKeycloakSession()
-        location.reload()
-      }
-    },
-
-    /** Clears Keycloak token information from session storage. */
-    clearKeycloakSession (): void {
-      sessionStorage.removeItem(SessionStorageKeys.KeyCloakToken)
-      sessionStorage.removeItem(SessionStorageKeys.KeyCloakRefreshToken)
-      sessionStorage.removeItem(SessionStorageKeys.KeyCloakIdToken)
-      sessionStorage.removeItem(SessionStorageKeys.CurrentAccount)
-    },
 
     /** Fetches business data / incorp app data. */
     async fetchData (): Promise<void> {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
     "provinces": "^1.11.0",
     "regenerator-runtime": "^0.13.7",
     "register-service-worker": "^1.7.1",
-    "sbc-common-components": "2.4.19",
+    "sbc-common-components": "2.4.38",
```
severinbeauvais, line 38: This was updated in another PR (still waiting to be merged). I will need to rebase.

src/App.vue
```diff
@@ -298,7 +294,6 @@ export default {
     if (!this.isAuthenticated) return
 
     // ...otherwise proceed
-    await this.startTokenService()
```
severinbeauvais, line 301: The old code started the token service only after the user was logged in, but that's a catch-22 because it syncs the session, which can log in (or log out) the user.

Also, the session needs to be synced (across domains) before any routing happens (since the router may redirect to a signin page).

So, this is moved to main.ts.

src/views/auth/Signin.vue
```diff
@@ -1,27 +1,17 @@
...
 export default class Signin extends Vue {
-  /** Called when Keycloak session is ready. */
-  private async onReady () {
-    // navigate to dashboard
-    this.$router.push({ name: Routes.DASHBOARD })
+  created () {
+    // redirect to BC Registry login page then return to this app
+    const loginUrl = sessionStorage.getItem('REGISTRY_HOME_URL') + 'login'
+    const returnUrl = encodeURIComponent(sessionStorage.getItem('BASE_URL'))
+    window.location.assign(`${loginUrl}?return=${returnUrl}`)
```
severinbeauvais, line 14: I could have moved this code inside router.ts, but this is cleaner ;)

src/views/auth/Signin.vue
```diff
@@ -1,27 +1,17 @@
 <template>
-  <sbc-signin @sync-user-profile-ready="onReady()" />
+  <div />
```
shabeeb-aot, line 2: can remove full template if not needed.

src/views/auth/Signin.vue
shabeeb-aot: onReady will get called after user authenticated. Hope its already handled and return URL will be dashboard

src/views/auth/Signout.vue
```diff
@@ -15,9 +12,9 @@ import SbcSignout from 'sbc-common-components/src/components/SbcSignout.vue'
   }
 })
 export default class Signout extends Vue {
-  /** The BCROS Home URL string. */
-  get bcrosHomeUrl (): string {
-    return sessionStorage.getItem('BUSINESSES_URL')
+  get redirectUrl (): string {
+    // after signout, redirect to BC Registries login page
+    return sessionStorage.getItem('REGISTRY_HOME_URL') + 'login'
```
shabeeb-aot, line 17: Better to give fall back also, if no value in session

shabeeb-aot (approved):

src/App.vue
cameron-freshworks: Does this apply to our other UI's as well?
My hunch is that we similar elsewhere as well.

cameron-freshworks (approved):
Looks good!

src/views/auth/Signin.vue
pwei1018: How do you handle the 'bookmark' with this page?

src/App.vue
severinbeauvais: Yes, this will apply to our other UIs where we init the token service in App instead of main.

src/views/auth/Signin.vue
severinbeauvais: > onReady will get called after user authenticated. Hope its already handled and return URL will be dashboard

This page no longer includes the SbcSignin component, so there is no longer an `onReady` event handler. This page is just a redirect to the BC Registy login page, which performs the login and then redirects to the `return` URL specified here.

severinbeauvais: > How do you handle the 'bookmark' with this page?

This page is transient and doesn't even display anything. The user may have bookmarked the "base" URL (which is the dashboard for this entity, eg, https://dev.bcregistry.ca/business/BC0870720), or they may bookmark the BC Registy login page (eg, https://yfthig-dev.web.app/login/), or perhaps a Business Registry link (marketing page or dashboard).

So if the user bookmarked the base URL and then are logged in, they just see the entity's dashboard. If they are not logged in then they will be redirected to the generic login page (BC Registry login) first.

src/views/auth/Signout.vue
severinbeauvais: Generally, yes... In this case, it's an environment variable set when the app starts up. That would be a serious error that would manifest in many other things not working before the user gets to this signout page.

Also, a fallback would also need to use an environment variable, which also might not work.

So, I'm thinking, this is not something we need to worry about. What do you think?

src/App.vue
severinbeauvais: Actually, Namerequest UI is already done this way. (That's what I copied from.)

src/views/auth/Signin.vue
severinbeauvais: Let me know if this is still now clear. I could discuss it at the next Tech Demo.

severinbeauvais:
@pwei1018 @shabeeb-aot Are you OK with my changes and responses above? OK to merge this PR?

shabeeb-aot:
I am good

src/views/auth/Signin.vue
pwei1018: I mean if someone bookmark this page "https://dev.bcregistry.ca/business/BC0870720/standalone-directors", after the user login it will go to the main dashboard and won't go to director page. The business pages might be ok. It won't have lots of layer.

severinbeauvais: Good example... but it will not work anyway because [this filing expects some app tombstone data](https://github.com/bcgov/business-filings-ui/blob/1c4f31189fad1ee53db895b818ed7dc63cf50592/src/views/StandaloneDirectorsFiling.vue#L468), which is very contextual (the app needs to have loaded all other entity data first). All filings are like this, so you cannot restore a bookmark to any filing. This is by design: we would not want a user to create an AR filing if one isn't due, etc.

severinbeauvais: ... And if you're wondering about the account ID URL parameter, that's [a different ticket](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/10999) :)

PR #290, opened 2022-02-04
10999 Handle account id sync

severinbeauvais:
*Issue #:* /bcgov/entity#10999

*Description of changes:*
- stored account id in session
- created "navigate" function to automatically add accountid parameter
- replaced window.location.assign with new function
- fixed date/time conversions broken by Node 14
- updated breadcrumb resources to append account id
- fixed invalid URLs in unit tests
- app version = 4.3.7

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -525,6 +525,7 @@ import { Component, Mixins, Prop, Watch } from 'vue-property-decorator'
 import { Action, Getter, State } from 'vuex-class'
 import axios from '@/axios-auth'
 import Vue2Filters from 'vue2-filters' // needed for orderBy
+import { navigate } from '@/utils'
 
 // Dialogs and Components
 import { AddCommentDialog, CancelPaymentErrorDialog, ConfirmDialog, DeleteErrorDialog } from '@/components/dialogs'
@@ -1164,9 +1165,9 @@ export default class TodoList extends Mixins(
 
       case FilingTypes.CORRECTION:
         if (item.correctedFilingType === FilingNames.INCORPORATION_APPLICATION) {
-          // redirect to Edit web app to correct this Incorporation Application
-          const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction?correction-id=${item.filingId}`
-          window.location.assign(correctionUrl) // assume URL is always reachable
+          // navigate to Edit web app to correct this Incorporation Application
+          const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction/?correction-id=${item.filingId}`
+          navigate(correctionUrl)
         } else {
           // resume this Correction Filing
           this.setCurrentFilingStatus(FilingStatus.DRAFT)
@@ -1177,21 +1178,21 @@ export default class TodoList extends Mixins(
         break
 
       case FilingTypes.INCORPORATION_APPLICATION:
-        // redirect to Create web app to resume this Incorporation Application
+        // navigate to Create web app to resume this Incorporation Application
         const incorpAppUrl = `${this.createUrl}?id=${this.tempRegNumber}`
-        window.location.assign(incorpAppUrl) // assume URL is always reachable
+        navigate(incorpAppUrl)
         break
 
       case FilingTypes.ALTERATION:
-        // redirect to Edit web app to alter this company
-        const alterationUrl = `${this.editUrl}${this.getIdentifier}/alteration?alteration-id=${item.filingId}`
-        window.location.assign(alterationUrl) // assume URL is always reachable
+        // navigate to Edit web app to alter this company
+        const alterationUrl = `${this.editUrl}${this.getIdentifier}/alteration/?alteration-id=${item.filingId}`
+        navigate(alterationUrl)
         break
 
       case FilingTypes.DISSOLUTION:
-        // redirect to Create web app to dissolve this company
-        const dissolutionUrl = `${this.createUrl}define-dissolution?id=${this.getIdentifier}`
-        window.location.assign(dissolutionUrl) // assume URL is always reachable
+        // navigate to Create web app to dissolve this company
+        const dissolutionUrl = `${this.createUrl}define-dissolution/?id=${this.getIdentifier}`
+        navigate(dissolutionUrl)
         break
 
       default:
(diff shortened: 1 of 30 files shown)


severinbeauvais:
Please ignore SonarCloud code duplication errors. Those are due to similar code in AR/COA/COD/Corr views. Fixing that is not in the scope of this ticket.

src/components/Dashboard/TodoList.vue
```diff
@@ -1164,9 +1165,9 @@ export default class TodoList extends Mixins(
 
       case FilingTypes.CORRECTION:
         if (item.correctedFilingType === FilingNames.INCORPORATION_APPLICATION) {
-          // redirect to Edit web app to correct this Incorporation Application
-          const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction?correction-id=${item.filingId}`
-          window.location.assign(correctionUrl) // assume URL is always reachable
+          // navigate to Edit web app to correct this Incorporation Application
+          const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction/?correction-id=${item.filingId}`
```
severinbeauvais, line 1169: Having this / at the end of the route and before the params is "more correct" but both work.

src/mixins/date-mixin.ts
```diff
@@ -76,14 +76,17 @@ export default class DateMixin extends Vue {
     // safety check
     if (!isDate(date) || isNaN(date.getTime())) return null
 
-    const dateStr = date.toLocaleDateString('en-CA', {
+    // NB: some versions of Node have only en-US locale
+    // so use that and convert results accordingly
+    const dateStr = date.toLocaleDateString('en-US', {
```
severinbeauvais, line 81: I needed to do this because I'm using Node 12 locally, and a bunch of unit tests that compared dates failed. This is the fix (and also line 157, below).

Ref: https://github.com/nodejs/node/issues/8500

src/resources/BreadcrumbResources.ts
```diff
@@ -1,16 +1,29 @@
...
+    text: 'My Business Registry',
+    href: `${sessionStorage.getItem('BUSINESSES_URL')}business${getParams()}`
+  }
+}
+
+export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
+  return {
+    text: 'Staff Dashboard',
+    href: `${sessionStorage.getItem('BUSINESSES_URL')}staff${getParams()}`
+  }
```
severinbeauvais, line 28: I had to change these to functions so they are evaluated when called (where ACCOUNT_ID exists), not during compile time (where it doesn't).

src/resources/business-config.ts
```diff
@@ -12,7 +12,7 @@ export class Business {
   flows: Array<Flow>
 }
 
-export const configJson = [
+export const ConfigJson = [
```
severinbeauvais, line 15: Since this is a const, I like the convention of title-casing it.

src/utils/fetch-config.ts
```diff
@@ -11,11 +11,19 @@ export async function fetchConfig (): Promise<void> {
...
   if (!processEnvBaseUrl || !windowLocationPathname || !windowLocationOrigin) {
     return Promise.reject(new Error('Missing environment variables'))
   }
 
+  // get and store account id, if present
+  const accountId = new URLSearchParams(windowLocationSearch).get('accountid')
+  if (accountId) {
+    sessionStorage.setItem('ACCOUNT_ID', accountId)
+    console.log('Set Account ID to: ' + accountId)
+  }
```
severinbeauvais, line 25: Note that SbcHeader has code to strip "accountid=xxx" out of the URL (so you don't see it) but THIS code works fine because it's run before SbcHeader is created.

src/utils/navigate.ts
```diff
@@ -0,0 +1,25 @@
...
+    // assume URL is always reachable
+    window.location.assign(url)
+
+    return true
+  } catch (error) {
+    console.log('Error navigating =', error) // eslint-disable-line no-console
+
+    return false
+  }
+}
```
severinbeauvais, line 25: There's a sexier way to parse the URL using `new URL(xxx)` and to get/set the parameters using `new URLSearchParams(yyy)` but it doesn't work consistently in the unit tests, so I rolled my own code.

tests/unit/AnnualReport.spec.ts
```diff
@@ -858,8 +858,8 @@ describe('Annual Report - Part 3 - Submitting', () => {
   it('saves a new filing and redirects to Pay URL when this is a new AR and the File & Pay button ' +
     'is clicked', async () => {
     // set necessary session variables
-    sessionStorage.setItem('BASE_URL', `${process.env.VUE_APP_PATH}/`)
-    sessionStorage.setItem('AUTH_WEB_URL', 'auth/')
+    sessionStorage.setItem('BASE_URL', 'https://base.url/')
+    sessionStorage.setItem('AUTH_WEB_URL', 'https://auth.web.url/')
```
severinbeauvais, line 862: I changed this to "real" URLs so that `new URL(xxx)` works (since it expects a valid URL), and then didn't end up using that, but I think this is still better code so I left it.

src/mixins/date-mixin.ts
```diff
@@ -76,14 +76,17 @@ export default class DateMixin extends Vue {
...
       timeZone: 'America/Vancouver',
       month: 'numeric', // 12
       day: 'numeric', // 31
       year: 'numeric' // 2020
     })
 
-    return dateStr
+    const [ mm, dd, yyyy ] = dateStr.split('/')
+    return `${yyyy}-${mm.padStart(2, '0')}-${dd.padStart(2, '0')}`
```
cameron-freshworks, line 89: A comment might be useful here!

cameron-freshworks (approved):
Interesting updates!
Looks good 👍

src/mixins/date-mixin.ts
severinbeauvais: Fixed in new commit.

src/components/Dashboard/TodoList.vue
argush3: oh...interesting.  I never thought about it deeply before but I've always used it without the `/` at the end.

src/mixins/date-mixin.ts
argush3: nice, i like.  very slick.

src/components/Dashboard/TodoList.vue
severinbeauvais: The way I see it, `.../correction/` is a route, so make it look like a route.

Conversely, `.../correction` could be a route with a parameter at the end. (Yes, a route parameter, but still a bit different IMHO).

Another way to look at it is that we specify various URLs in our config/env and most all end with a /, so we're following convention here.
```
  "AUTH_WEB_URL": "https://dev.bcregistry.ca/business/auth/",
  "DASHBOARD_URL": "https://dev.bcregistry.ca/business/",
```

argush3: makes sense

argush3 (approved):
looks good!

PR #297, opened 2022-02-14
Hide Show Documents Button in the Filing History for filing with no documents

lekshmimallika-aot:
*Issue #:* /bcgov/entity#10934

*Description of changes:*

- Updated the filing history list component to show the view documents button only if documentsLink exists and has a non null value
- Updated tests

### DIFF
--- a/tests/unit/FilingHistoryList.spec.ts
+++ b/tests/unit/FilingHistoryList.spec.ts
@@ -1338,6 +1338,7 @@ describe('Filing History List - paper only and other filings', () => {
         status: 'PAID',
         submittedDate: '2020-03-24 19:20:05 GMT',
         submitter: 'Cameron',
+        documentsLink: 'http://test',
         data: {
           alteration: {
             fromLegalType: 'BC',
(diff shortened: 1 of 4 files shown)


tests/unit/FilingHistoryList.spec.ts
```diff
@@ -1956,3 +1973,32 @@ describe('Filing History List - detail comments', () => {
     wrapper.destroy()
   })
 })
+
+describe('Filing History List - documents', () => {
+  const FILING_WITH_DOCUMENTS_LINK = {
```
severinbeauvais, line 1978: WITHOUT documents link?

severinbeauvais (commented):
Looks OK.

Please update the app version (in package.json, then run `npm i`) so you can track where your change is deployed.

tests/unit/FilingHistoryList.spec.ts
lekshmimallika-aot: ya :) will change it

severinbeauvais (approved):

PR #304, opened 2022-02-24
Updated Firm Redirect

cameron-freshworks:
*Issue #:* /bcgov/entity#10613

*Description of changes:*
* Small QOL update for Firms to redirect to the appropriate route when starting a Change Filing vs an Alteration Filing

### DIFF
--- a/src/components/EntityInfo.vue
+++ b/src/components/EntityInfo.vue
@@ -269,7 +269,9 @@ export default class EntityInfo extends Mixins(AllowableActionsMixin, CommonMixi
     if (!this.isGoodStanding) {
       this.emitNotInGoodStanding(NigsMessage.CHANGE_COMPANY_INFO)
     } else {
-      const url = `${this.editUrl}${this.getIdentifier}/alteration`
+      let url = `${this.editUrl}${this.getIdentifier}`
+      // Append appropriate route based on entity type
+      url += (this.isFirm ? '/change' : '/alteration')
       navigate(url)
     }
   }
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -72,6 +72,11 @@ export default {
     return (state.entityType === CorpTypeCd.BC_CORPORATION)
   },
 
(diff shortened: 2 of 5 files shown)


src/store/getters.ts
```diff
@@ -72,6 +72,11 @@ export default {
     return (state.entityType === CorpTypeCd.BC_CORPORATION)
   },
 
+  /** Is True if entity is a Firm. */
+  isFirm (state: StateIF): boolean {
+    return (state.entityType === CorpTypeCd.SOLE_PROP || state.entityType === CorpTypeCd.PARTNERSHIP)
+  },
```
cameron-freshworks, line 78: 1 getter for now as i haven't seen anything differentiating the two with regards to this UI. Can be split up when/if required.

src/components/EntityInfo.vue
```diff
@@ -269,7 +269,9 @@ export default class EntityInfo extends Mixins(AllowableActionsMixin, CommonMixi
     if (!this.isGoodStanding) {
       this.emitNotInGoodStanding(NigsMessage.CHANGE_COMPANY_INFO)
     } else {
-      const url = `${this.editUrl}${this.getIdentifier}/alteration`
+      let url = `${this.editUrl}${this.getIdentifier}`
+      // Append appropriate route based on entity type
+      url += (this.isFirm ? '/change' : '/alteration')
```
cameron-freshworks, line 274: Fairly certain Firms are/will be the exception. Majority of entity types will be alterations.

severinbeauvais (approved):

src/components/EntityInfo.vue
cameron-freshworks: I think so yea
We were explicit with Alterations, so i figured we could do the same with change filings. 
This could change in the future but this seems pretty straightforward for now :)

severinbeauvais: Sounds good.

PR #306, opened 2022-03-01
Support Change of Registration Drafts

cameron-freshworks:
*Issue #:* /bcgov/entity#10613

*Description of changes:*
* Implements handlers for Change of Registration Draft Filings

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -1180,7 +1180,6 @@ export default class TodoList extends Mixins(
     const header = filing.header
     const changeOfRegistration = filing.changeOfRegistration
 
-    // NB: don't check "changeOfRegistration" as it may be empty
     if (header) {
       const title = `Change to ${this.getCorpTypeDescription(this.getEntityType)} Registration`
 
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -848,6 +848,9 @@ export default class TodoList extends Mixins(
         case FilingTypes.REGISTRATION:
           await this.loadRegistration(task)
           break
+        case FilingTypes.CHANGE_OF_REGISTRATION:
+          await this.loadChangeOfRegistration(task)
+          break
         default:
           // eslint-disable-next-line no-console
           console.log('ERROR - invalid name in filing header =', header)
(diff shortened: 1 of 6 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -1172,6 +1175,54 @@ export default class TodoList extends Mixins(
     }
   }
 
+  private async loadChangeOfRegistration (task: ApiTaskIF): Promise<void> {
+    const filing = task.task.filing
+    const header = filing.header
+    const changeOfRegistration = filing.changeOfRegistration
+
+    // NB: don't check "changeOfRegistration" as it may be empty
```
severinbeauvais, line 1183: Is this true or did you clone it from another function?

src/components/Dashboard/TodoList.vue
```diff
@@ -1172,6 +1175,54 @@ export default class TodoList extends Mixins(
...
+      if (this.isStatusDraft(header)) {
+        if (this.nameRequest) {
+          subtitle = `NR APPROVED - ${this.expiresText(this.nameRequest)}`
+        } else {
+          subtitle = 'DRAFT'
+        }
+      }
+
+      const paymentStatusCode = header.paymentStatusCode || null
+      const payErrorObj = paymentStatusCode && await this.getPayErrorObj(paymentStatusCode)
```
severinbeauvais, line 1198: Does `getPayErrorObj()` never throw an error (that needs to be handled here)? (It might catch its own errors and always return something sane, I dunno.)

src/enums/filingNames.ts
```diff
@@ -17,5 +17,6 @@ export enum FilingNames {
   REGISTRARS_ORDER = 'Registrar\'s Order',
   SPECIAL_RESOLUTION = 'Special Resolution',
   TRANSITION_APPLICATION = 'Transition Application',
-  VOLUNTARY_DISSOLUTION = 'Voluntary Dissolution'
+  VOLUNTARY_DISSOLUTION = 'Voluntary Dissolution',
+  CHANGE_OF_REGISTRATION = 'Change of Registration'
```
severinbeauvais, line 21: Please sort this (and also filingTypes below).

severinbeauvais (approved):
Looks good.

src/components/Dashboard/TodoList.vue
cameron-freshworks: mmmmm no, it shouldn't be true actually. 
It's a copy - paste - clone error.
Removed!

src/components/Dashboard/TodoList.vue
cameron-freshworks: It's an interesting request. 
It will make a request to fetch any errors associated with the code, return them if there is errors. 
Catches itself is there is an error and ignores them. 
So on this level it would be payErrorObj  would be `false`

src/enums/filingNames.ts
cameron-freshworks: Will do, thanks!
Updated

src/components/Dashboard/TodoList.vue
severinbeauvais: Perfect. Thanks for confirming!

src/enums/filingTypes.ts
```diff
@@ -17,6 +17,7 @@ export enum FilingTypes {
   SPECIAL_RESOLUTION = 'specialResolution',
   TRANSITION = 'transition',
   VOLUNTARY_DISSOLUTION = 'voluntaryDissolution',
+  CHANGE_OF_REGISTRATION = 'changeOfRegistration',
```
severinbeauvais, line 20: Please put this is sorted order, too :)

src/components/Dashboard/TodoList.vue
cameron-freshworks: Good idea 👍

PR #312, opened 2022-04-14
Fixed Role Check

cameron-freshworks:
*Issue #:* /bcgov/entity#11898

*Description of changes:*
* fixed role check to adhere to api update

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -644,17 +644,6 @@ export default {
       // set addresses
       this.storeAddresses({ data: application.offices || [] })
 
-      // Format party roles from IA
-      application.parties.forEach( // Check each party for roles
-        party => party.roles?.forEach( // Check each role for roleType
-          (role, index) => {
-            if (role.roleType) { // If roleType exists, assign it to parent role index
-              party.roles[index] = role.roleType.toLowerCase()
-            }
-          }
-        )
-      )
-
       // Set parties
       this.storeParties({ data: { parties: application.parties || [] } })
 
@@ -749,7 +738,6 @@ export default {
 
     storeParties (response: any): void {
       const parties = response?.data?.parties
-      console.log(parties)
       if (parties) {
         const directorsList = parties?.filter(party => party.roles?.some(role => role.roleType === Roles.DIRECTOR))
         const custodianList = parties?.filter(party => party.roles?.some(role => role.roleType === Roles.CUSTODIAN))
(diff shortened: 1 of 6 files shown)


tests/unit/App.spec.ts
```diff
@@ -140,11 +140,11 @@ const BCOMP_PARTIES = [
     roles: [
       {
         appointmentDate: '2020-07-06',
-        roleType: 'completing party'
+        roleType: 'Completing party'
```
vysakh-menon-aot, line 143: p should be Caps

cameron-freshworks: Updated, thanks!

src/App.vue
```diff
@@ -644,17 +644,6 @@ export default {
...
-      // Format party roles from IA
-      application.parties.forEach( // Check each party for roles
-        party => party.roles?.forEach( // Check each role for roleType
-          (role, index) => {
-            if (role.roleType) { // If roleType exists, assign it to parent role index
-              party.roles[index] = role.roleType.toLowerCase()
-            }
-          }
-        )
-      )
```
severinbeauvais, line 656: No longer needed?

severinbeauvais (approved):
LGTM

vysakh-menon-aot (approved):

src/App.vue
cameron-freshworks: I don't believe so no, but to verify, what state is this triggered. 
Is this after a user completes a Filed IA, that interm state before they have to reload there dashboard?

cameron-freshworks: Going to try and verify the structure in the IA. This could still be an edge case

cameron-freshworks: Ok, verified 👍

severinbeauvais: Did you file an IA to test this?

severinbeauvais: > I don't believe so no, but to verify, what state is this triggered. Is this after a user completes a Filed IA, that interm state before they have to reload there dashboard?

Yeah, looks like that scenario.

When the entity is opened using a permanent ID then the IA filing will naturally appear in the ledger.

cameron-freshworks: Exactly, you bet. The state where it’s still a temp reg but with a filing in the history.

severinbeauvais: You da man.
💯 
😎

PR #317, opened 2022-04-22
Display business address for firms in filings dashboard

lekshmimallika-aot:
*Issue #:* /bcgov/entity#9840

*Description of changes:*
Component for showing firms address as the layout itself is different
Changes in dashboard component to redirect to edit UI, to show the correct header etc.
Test updates
 


src/components/Dashboard/AddressListSm.vue
```diff
@@ -1,15 +1,19 @@
 <template>
   <div id="address-list-sm">
+    <template v-if="isFirm">
+        <firms-address-list
```
severinbeauvais, line 4: If you use the PascalCase name here then VS Code lets you click F12 to display the component.

src/store/getters.ts
```diff
@@ -72,6 +72,11 @@ export default {
     return (state.entityType === CorpTypeCd.BC_CORPORATION)
   },
 
+  /** Is True if entity is a Sole Proprietorship or General Partnership. */
+  isFirm (state: StateIF): boolean {
+    return (state.entityType === CorpTypeCd.SOLE_PROP || state.entityType === CorpTypeCd.PARTNERSHIP)
+  },
```
severinbeauvais, line 78: I'd prefer to see as a local getter in the component that uses it.

If, in the future, we find ourselves using this multiple times, then we can refactor it.

Also, should this check for ULC, CCC, etc?

src/views/Dashboard.vue
```diff
@@ -60,7 +60,7 @@
 
             <section>
               <header class="aside-header mb-3">
-                <h2 data-test-id="dashboard-addresses-subtitle">Office Addresses</h2>
+                <h2 data-test-id="dashboard-addresses-subtitle">{{addressHeader}}</h2>
```
severinbeauvais, line 63: I'm thinking line 84 (`:disabled`) should also check the hasBlocker getter. Or is there something else that disables this button if there's a blocker filing?

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -0,0 +1,172 @@
+<template>
+  <div id="address-list-sm">
```
severinbeauvais, line 2: Update this id to match the component name.

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -0,0 +1,172 @@
...
+import { State } from 'vuex-class'
+
+// Mixins
+import { CommonMixin, CountriesProvincesMixin } from '@/mixins'
+
+// Interfaces
+import { OfficeAddressIF } from '@/interfaces'
+
+@Component({})
+export default class FirmsAddressList extends Mixins(CommonMixin, CountriesProvincesMixin) {
```
severinbeauvais, line 105: This all looks substantially similar to the existing template/script in AddressListSm.vue. Couldn't you blend the changes into that instead of duplicating so much code?

src/views/Dashboard.vue
```diff
@@ -211,7 +222,12 @@ export default {
 
     /** Display COA warning if BCOMP else proceed to COA. */
     proceedCoa () {
-      this.isBComp ? this.toggleCoaWarning() : this.goToStandaloneAddresses()
+      if (this.isFirm) {
+        let url = `${this.editUrl}${this.getIdentifier}/change`
```
argush3, line 226: this can be a `const` instead of a `let`

src/components/Dashboard/FirmsAddressList.vue
severinbeauvais: OK, Lekshmi and I chatted and determined it's too messy to do this all in AddressListSm, so I'm OK with this sub-component.

src/views/Dashboard.vue
lekshmimallika-aot: I think value passed to the showGrayedOut prop already takes care of this 
image

src/store/getters.ts
lekshmimallika-aot: I think ULC,CCC etc comes under corporations. I added this in the getter because this was getting used in both dashboard and FirmsAddressList

src/views/Dashboard.vue
severinbeauvais: Thanks for checking. I see now that `isAllowed(AllowableActions.FILE_ADDRESS_CHANGE)` checks for hasBlocker.

src/components/Dashboard/AddressListSm.vue
severinbeauvais: PascalCase starts with a capital.

But I see it works in camelCase too :)

severinbeauvais (approved):
Looks good.

src/components/Dashboard/AddressListSm.vue
lekshmimallika-aot: Oh right :P . Updating it

argush3 (approved):

PR #322, opened 2022-05-05
12024 Filings UI - Display "Not Entered" for missing data on address and prop/part components

lewischenstudio:
*Issue #:* [/bcgov/entity#12024](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12024)

*Description of changes:*

When a user opens their entity dashboard for an SP/GP, if data is missing we need to:

- [x] Still display the appropriate headers
- [x] Show "Not Entered" if data fields are empty
  - When the field would normally not be empty for this entity type
  - This applies to firms only (so only those components need this change)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "business-filings-ui",
-      "version": "5.0.0",
+      "version": "5.0.1",
       "dependencies": {
         "@babel/compat-data": "^7.11.0",
         "@bcrs-shared-components/bread-crumb": "^1.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.0.0",
+  "version": "5.0.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/src/components/Dashboard/FirmsAddressList.vue
+++ b/src/components/Dashboard/FirmsAddressList.vue
@@ -34,14 +34,14 @@
         <div v-else class="pt-4 pb-4">
           <v-list class="pt-0 pb-0" v-if="businessAddress">
             <!-- Delivery Address -->
-            <v-list-item class="delivery-address-list-item" v-if="businessAddress.deliveryAddress">
+            <v-list-item class="delivery-address-list-item">
               <v-list-item-icon class="address-icon mr-0">
                 <v-icon color="primary">mdi-truck</v-icon>
               </v-list-item-icon>
               <v-list-item-content>
                 <v-list-item-title class="mb-2 address-title">Delivery Address</v-list-item-title>
                 <v-list-item-subtitle>
-                  <ul class="address-subtitle pre-line">
+                  <ul class="address-subtitle pre-line" v-if="businessAddress.deliveryAddress">
                     <li class="address-line1">{{ businessAddress.deliveryAddress.streetAddress }}</li>
                     <li class="address-line2">{{ businessAddress.deliveryAddress.streetAddressAdditional }}</li>
                     <li class="address-line3">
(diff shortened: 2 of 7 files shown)


lekshmimallika-aot (approved):

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.0.0",
+  "version": "5.0.1",
```
severinbeauvais, line 3: Please run `npm i` to update this file. Did you update it manually instead? It is missing some changes.

vysakh-menon-aot (approved):

thorwolpert (approved):

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -53,35 +53,41 @@
                       <span>{{ getCountryName(businessAddress.deliveryAddress.addressCountry) }}</span>
                     </li>
                   </ul>
+                  <ul class="address-subtitle pre-line" v-else>
+                    <li class="address-line1">Not Entered</li>
```
severinbeauvais, line 57: I recommend a new classname here. AFAIK these classnames are only used for selecting elements during testing.

It looks like your change broke a unit test, that you need to fix. Also it would be nice to add some test cases that verify your change versus normal behaviour.

src/components/Dashboard/ProprietorPartnersListSm.vue
```diff
@@ -20,46 +20,52 @@
...
+                  <ul class="address-subtitle pre-line" v-if="party.deliveryAddress">
                     <li class="address-line1">{{ party.deliveryAddress.streetAddress }}</li>
                     <li class="address-line2">{{ party.deliveryAddress.streetAddressAdditional }}</li>
                     <li class="address-line3">{{ party.deliveryAddress.addressCity }}
                                               {{ party.deliveryAddress.addressRegion }}
                                               {{ party.deliveryAddress.postalCode }}</li>
                     <li class="address-line4">{{ getCountryName(party.deliveryAddress.addressCountry) }}</li>
                   </ul>
+                  <ul class="address-subtitle pre-line" v-else>
+                    <li class="address-line1">Not Entered</li>
```
severinbeauvais, line 43: Same as above: please use new classname, and also add a unit test or 2 😃

lewischenstudio: I added new classes and 2 unit tests for the two related components.

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -34,14 +34,14 @@
...
-            <v-list-item class="delivery-address-list-item" v-if="businessAddress.deliveryAddress">
+            <v-list-item class="delivery-address-list-item">
               <v-list-item-icon class="address-icon mr-0">
                 <v-icon color="primary">mdi-truck</v-icon>
               </v-list-item-icon>
               <v-list-item-content>
                 <v-list-item-title class="mb-2 address-title">Delivery Address</v-list-item-title>
                 <v-list-item-subtitle>
-                  <ul class="address-subtitle pre-line">
+                  <ul class="address-subtitle pre-line" v-if="businessAddress.deliveryAddress">
```
severinbeauvais, line 44: Did you test with mock data? What does the data look like when there is no business address?
- does the app load or throw an error while fetching data?
- does the conditional on line 35 pass?

src/components/Dashboard/ProprietorPartnersListSm.vue
```diff
@@ -20,46 +20,55 @@
...
               </v-list-item-content>
             </v-list-item>
 
-            <v-list-item class="delivery-address-list-item" v-if="party.deliveryAddress">
+            <v-list-item class="delivery-address-list-item">
               <v-list-item-content>
                 <v-list-item-title class="mb-2 address-title">Delivery Address</v-list-item-title>
                 <v-list-item-subtitle>
-                  <ul class="address-subtitle pre-line">
+                  <ul class="address-subtitle pre-line" v-if="party.deliveryAddress">
```
severinbeauvais, line 36: I just realized that the ticket was not clear -- there may be no proprietor/partners at all and that's when "Not Entered" should display in this component.

severinbeauvais: ^^ I added a ticket comment to ask Linda for clarification

src/components/Dashboard/FirmsAddressList.vue
lewischenstudio: - When there is no business address, then It displays "Complete your filing to display" for all components.
- When there is data but delivery address or mailing address is null, then it displays "Not Entered".

I had the same question too. In what scenario do we ever run into "Not Entered" for delivery/mailing address, if create-ui is error-proof already.

lewischenstudio: I did test it with mock data and I am writing a unit tests for those components. I always forget to write unit test.

src/components/Dashboard/ProprietorPartnersListSm.vue
lewischenstudio: When there is no "proprietor/partners" or delivery/mailing address at all, it displays "Complete your filing to display".

src/components/Dashboard/FirmsAddressList.vue
severinbeauvais: 1. We may not have the info if the firm was loaded directly from the old system. That's why we need to do conversion filings.

2. Your test data assumes that there is a business object that contains a null delivery address and a null mailing address. Is that actually what is in the store when one of these incomplete firms is loaded in Filings UI?

src/components/Dashboard/ProprietorPartnersListSm.vue
severinbeauvais: Right.

Note that in both scenarios there may be no parties (orgpersons). You need to differentiate between those scenarios in your code. I assume you can check if the current entity is using a temp registration number (ie, it's still a registration) or is using an actual business number (ie, it's now a firm).

lewischenstudio: image

The current implementation from line 3 to 5 (before this branch) covers the scenario for no parties, and it displays "Complete your filing to display".

lambert-alex (approved):

src/components/Dashboard/FirmsAddressList.vue
lewischenstudio: Looking at Argus's file, I think `delivery_address` or `mailing_address` can be `null`. From that file, There are several cases where `delivery_address` is null but `mailing_address` is not null.

severinbeauvais: Could `businessAddress` be null?

src/components/Dashboard/ProprietorPartnersListSm.vue
severinbeauvais: I'm resolving this conversation now, but this code will need to be re-reviewed when we better understand what "missing data" scenarios we might encounter, and then we can look at what messages are displayed in each case.

severinbeauvais (changes requested):
Do not merge until we understand what "missing data" scenarios we might encounter and we handle each of them accordingly.

At the moment, I'm not convinced we display the right messages for the various scenarios (both incomplete business addresses and incomplete party data).

src/components/Dashboard/FirmsAddressList.vue
lewischenstudio: `businessAddress` can be null, but it is handled on layer up in line 35.

severinbeauvais: I see the conditional on line 35, but I don't see the "Not Entered" message in this situation.

lewischenstudio: I added the v-else for the case where businessAddress is null in my latest commit.

severinbeauvais: OK, this is very safe code now :)

Also, great job on the unit tests where you verify all possible scenarios 👍

src/components/Dashboard/ProprietorPartnersListSm.vue
```diff
@@ -20,46 +20,55 @@
```
severinbeauvais, line 20: What about the scenario where there are no parties?

lewischenstudio: I added the condition for no parties and included unit test in my latest commit.

severinbeauvais: Shouldn't it say "Not Entered" in this case?

Also, you don't need the "v-else" to protect the element at line 12: the v-for handles empty arrays just fine.

severinbeauvais (commented):
Looks good.

I'm OK with the codecov results as the files it's complaining about were not touched in this PR.

severinbeauvais (approved):

PR #323, opened 2022-05-09
 Invoking the dissolution(with the modal to confirm)#11663

shabeeb-aot:
*Issue #:* /bcgov/entity#11663

*Description of changes:*
* Confirmation modal changes for SP/GP/ dissolution
* update business address while sending to API


src/components/dialogs/ConfirmDissolution.vue
```diff
@@ -56,28 +57,64 @@ export default class ConfirmDissolution extends Vue {
...
+        title = 'Dissolution'
+        break
+      default:
+        entityTitle = 'CompanyCompany'
+        subTitle = 'company'
+        act = 'Business Corporations'
+        title = 'Voluntary Dissolution'
+    }
+    return { entityTitle, subTitle, act, title }
+  }
```
severinbeauvais, line 117: This project has not yet implemented the "resources concept" as per Create UI or Edit UI. We knew this was coming but I didn't know you would be the first to really benefit from this.

Would you be able to copy the resources concept here please? In a nutshell:
- in a "resources" folder are resource files for each entity type, eg: https://github.com/bcgov/business-create-ui/tree/main/src/resources/Incorporations (you probably don't need the sub-folder)
- in this component you can "find" the correct resources, eg: https://github.com/bcgov/business-create-ui/blob/12d332edb99ff14a4fb6c729c2be0e8926d2de59/src/App.vue#L774 (up to you whether the resources need to be global/store or not; we can refactor this later as needed)
- in this component, just display the labels from the resource

src/components/dialogs/ConfirmDissolution.vue
```diff
@@ -56,28 +57,64 @@ export default class ConfirmDissolution extends Vue {
...
   }
 
   // Pass click event to parent.
   @Emit() private close () { }
 
   // Pass click event to parent.
   @Emit() private proceed () { }
+
+  // returning  title subtitle, and act by entityType
+  private getEntityDetails (): any {
```
severinbeauvais, line 89: Did you try this as a getter (since there are no formal parameters)? It would probably have been reactive to the entity type. It might have simplified some syntax, and as a plug, getters are cached so often the code is faster :)

src/mixins/filing-mixin.ts
```diff
@@ -19,6 +19,7 @@ export default class FilingMixin extends Mixins(DateMixin) {
   @Getter getIdentifier!: string
   @Getter getEntityFoundingDate!: Date
   @Getter getRegisteredOfficeAddress!: OfficeAddressIF
+  @Getter gettBusinessAddress!: OfficeAddressIF
```
severinbeauvais, line 22: Typo? "gett"

src/store/getters.ts
```diff
@@ -206,5 +206,11 @@ export default {
   /** The parties list from the API. */
   getParties (state: StateIF): Array<PartyIF> {
     return state.parties
+  },
+
+  /** The entity busness address. */
+  gettBusinessAddress (state: StateIF): OfficeAddressIF {
+    return state.businessAddress
   }
```
severinbeauvais, line 214: Now that we have a nice getter for this, could you please update FirmsAddressList.vue:122 to replace the State with Getter?

src/components/dialogs/ConfirmDissolution.vue
shabeeb-aot: As discussed in offline moved to business-config

severinbeauvais: It looks waaaaay better :)

src/store/getters.ts
severinbeauvais: TY

src/components/Dashboard/TodoList.vue
```diff
@@ -877,7 +878,7 @@ export default class TodoList extends Mixins(
         name: FilingTypes.DISSOLUTION,
         filingId: header.filingId,
         legalType: corpTypeDescription,
-        title: 'Voluntary Dissolution',
+        title: this.isFirm ? 'Dissolution' : 'Voluntary Dissolution',
```
severinbeauvais, line 881: 1. I'd really prefer to see the resources used here instead of making this component specific for firms.

2. It's odd that this displays "Dissolution" for a firm and "VD" for everything else, but the resources specify the exact opposite. Typo?

severinbeauvais:
SonarCloud is complaining about "duplicate lines" in the resource file. This is actually OK.

I thought there was a way to login to SonarCloud and explicitly approve an exception here, but at the moment I can't figure out how to do that. If you can -- great. If not, don't worry about it.

severinbeauvais:
As for the code coverage, do you have time to create a test file for the dialog component? You can look at the other dialog test suites for example code.

shabeeb-aot:
> SonarCloud is complaining about "duplicate lines" in the resource file. This is actually OK.
> 
> I thought there was a way to login to SonarCloud and explicitly approve an exception here, but at the moment I can't figure out how to do that. If you can -- great. If not, don't worry about it.

we have to exclude this file in sonarCloud, I think we need admin access for it
https://docs.sonarcloud.io/advanced-setup/analysis-scope/#restrict-the-scope-of-analysis-in-general

shabeeb-aot:
> As for the code coverage, do you have time to create a test file for the dialog component? You can look at the other dialog test suites for example code.

Actually, i am pretty packed with work, I created a simple test file for dissolution component only for one type now

severinbeauvais:
> we have to exclude this file in sonarCloud, I think we need admin access for it

Yes, I think it's reasonable to exclude the resources folder from SonarCloud. I don't think I have admin access. Can you please ask Patrick? Maybe he can do this for us, or give us both admin access :)

src/components/Dashboard/TodoList.vue
severinbeauvais: Awesome! Thanks.

src/interfaces/todo-list-interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface TodoListResourceIF {
+  title: string
+
```
severinbeauvais, line 3: Does this blank line cause a lint error locally? `npm run lint`

src/resources/business-config.ts
```diff
@@ -73,6 +73,16 @@ export const ConfigJson = [
...
       ]
+    },
+    dissolutionConfirmation: {
+      entityTitle: 'Company',
+      subTitle: 'company',
+      act: 'Business Corporations',
+      modalTitle: 'Voluntary Dissolution',
+      confirmButtonText: 'Continue with Voluntary Dissolution'
+    },
+    TodoList: {
```
severinbeauvais, line 84: Now or in a future commit (no rush), please change this to [camelCase](https://en.wikipedia.org/wiki/Camel_case), as is our code style for variables.

severinbeauvais: PS I'm OK with this entire object (ConfigJson) being PascalCase because it's a "const".

tests/unit/ConfirmDissolution.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ConfigJson } from '@/resources'
+
+Vue.use(Vuetify)
+
+const vuetify = new Vuetify({})
+const store = getVuexStore() as any // remove typings for unit tests
+
+describe('ConfirmDissolution - Displays Confirmation messages', () => {
+  beforeAll(() => {
+    store.state.entityType = 'SP' // now added only for SP
```
severinbeauvais, line 15: In a future commit, you could iterate over entity types to test all variations.

Example: https://github.com/bcgov/business-create-ui/blob/main/tests/unit/AgreementType.spec.ts

src/interfaces/todo-list-interface.ts
shabeeb-aot: it does not show any error, still, it is good to fix. fixed it

PR #327, opened 2022-05-12
12024 Fix Filings UI - Display "Not Entered" for missing data on address and prop/part components

lewischenstudio:
*Issue #:* /bcgov/entity#12024

*Description of changes:*

**Completed UXA Observations:**

- [x] FM0614924 is an SP missing the Proprietor. The Proprietor section on the dashboard is empty. The placeholder text "Not Entered" should be displayed inside a box. (This works fine for a GP missing partners)
- [x] FM0614946 is an SP (DBA) missing the Proprietor. The Proprietor section on the dashboard is empty. The placeholder text "Not Entered" should be displayed inside a box.
- [x] Changed "Not Entered" to "(Not Entered)" with parentheses. 

**404 Handler**
- [x] Filing UI can load FM0272480 FM0272488 with 404 handler.

**New Ticket**
- [x] Edit UI can't load FM0614924/FM0613561/FM0272480/FM0272488: new ticket created [12206](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12206)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -372,6 +372,7 @@ export default {
       if (this.businessId) {
         try {
           await this.fetchBusinessData() // throws on error
+          await this.fetchAddressData() // business address
           this.dataLoaded = true
         } catch (error) {
           console.log(error) // eslint-disable-line no-console
@@ -402,18 +403,32 @@ export default {
         this.fetchBusinessInfo(this.businessId),
         this.fetchTasks(this.businessId),
         this.fetchFilings(this.businessId || this.tempRegNumber),
-        this.fetchAddresses(this.businessId),
         this.fetchParties(this.businessId)
       ])
 
-      if (!data || data.length !== 6) throw new Error('Incomplete business data')
+      if (!data || data.length !== 5) throw new Error('Incomplete business data')
 
       this.storeEntityInfo(data[0])
       await this.storeBusinessInfo(data[1])
       this.storeTasks(data[2])
       this.storeFilings(data[3])
-      this.storeAddresses(data[4])
-      this.storeParties(data[5])
+      this.storeParties(data[4])
+    },
+
+    async fetchAddressData ():Promise<void> {
+      let hasBAError = false
+      const data = await Promise.resolve(
+        this.fetchAddresses(this.businessId)
+      ).catch(error => {
+        if (error.response.status === 404 &&
+          error.response.config.url.includes('addresses') &&
+          error.response.data.message.includes('address not found')) hasBAError = true
+      })
+
+      let badAddress = { data: { businessOffice: null } }
+      if (!data && hasBAError) this.storeAddresses(badAddress) // if 404 and business address not found
+      if (!data && !hasBAError) throw new Error('Incomplete business data')
+      if (data) this.storeAddresses(data)
     },
 
     /** Fetches and stores the draft application data. */
--- a/src/App.vue
+++ b/src/App.vue
@@ -372,7 +372,7 @@ export default {
       if (this.businessId) {
         try {
           await this.fetchBusinessData() // throws on error
-          await this.fetchAddressData() // business address
+          await this.fetchStoreAddressData() // business address
           this.dataLoaded = true
         } catch (error) {
           console.log(error) // eslint-disable-line no-console
@@ -415,7 +415,7 @@ export default {
       this.storeParties(data[4])
     },
 
-    async fetchAddressData ():Promise<void> {
+    async fetchStoreAddressData ():Promise<void> {
       let hasBAError = false
       const data = await Promise.resolve(
         this.fetchAddresses(this.businessId)
@@ -425,10 +425,13 @@ export default {
           error.response.data.message.includes('address not found')) hasBAError = true
       })
 
-      let badAddress = { data: { businessOffice: null } }
-      if (!data && hasBAError) this.storeAddresses(badAddress) // if 404 and business address not found
-      if (!data && !hasBAError) throw new Error('Incomplete business data')
-      if (data) this.storeAddresses(data)
+      if (data) {
+        this.storeAddresses(data)
+      } else if (hasBAError) { // if 404 and business address not found
+        this.storeAddresses({ data: { businessOffice: null } })
+      } else {
+        throw new Error('Incomplete business data')
+      }
     },
 
     /** Fetches and stores the draft application data. */
--- a/src/components/Dashboard/FirmsAddressList.vue
+++ b/src/components/Dashboard/FirmsAddressList.vue
@@ -133,7 +133,10 @@ export default class FirmsAddressList extends Mixins(CommonMixin, CountriesProvi
   @Getter getBusinessAddress!: OfficeAddressIF
 
   // Business
-  get businessAddress (): OfficeAddressIF {
+  get businessAddress (): OfficeAddressIF|null {
+    if (this.getBusinessAddress && this.getBusinessAddress.deliveryAddress &&
+        Object.values(this.getBusinessAddress.deliveryAddress).length === 0) return null
+    if (this.getBusinessAddress && this.getBusinessAddress.deliveryAddress === null) return null
     return this.getBusinessAddress
   }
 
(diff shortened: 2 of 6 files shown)


src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -133,7 +133,10 @@ export default class FirmsAddressList extends Mixins(CommonMixin, CountriesProvi
   @Getter getBusinessAddress!: OfficeAddressIF
 
   // Business
-  get businessAddress (): OfficeAddressIF {
+  get businessAddress (): OfficeAddressIF|null {
+    if (this.getBusinessAddress && this.getBusinessAddress.deliveryAddress &&
```
severinbeauvais, line 137: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here to simplify things a bit.

Also on line 139.

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -133,7 +133,10 @@ export default class FirmsAddressList extends Mixins(CommonMixin, CountriesProvi
...
+        Object.values(this.getBusinessAddress.deliveryAddress).length === 0) return null
```
severinbeauvais, line 138: `Object.values()` should only be used in exceptional circumstances. What are you checking for here? Show me a sample object that this matches.

Also, could any of this this be simplified with [isEmpty](https://www.geeksforgeeks.org/lodash-_-isempty-method/)?

src/App.vue
```diff
@@ -402,18 +403,32 @@ export default {
...
       this.storeFilings(data[3])
-      this.storeAddresses(data[4])
-      this.storeParties(data[5])
+      this.storeParties(data[4])
+    },
+
+    async fetchAddressData ():Promise<void> {
+      let hasBAError = false
+      const data = await Promise.resolve(
+        this.fetchAddresses(this.businessId)
```
severinbeauvais, line 421: I think you can simplify this to:
```
const data = this.fetchAddresses(this.businessId)
   .catch(error => {
...
```

src/App.vue
```diff
@@ -372,6 +372,7 @@ export default {
       if (this.businessId) {
         try {
           await this.fetchBusinessData() // throws on error
+          await this.fetchAddressData() // business address
```
severinbeauvais, line 375: Since this does more than just fetching the data, I'd prefer to see it named "loadAddressData" instead.

src/App.vue
```diff
@@ -402,18 +403,32 @@ export default {
...
+      ).catch(error => {
+        if (error.response.status === 404 &&
+          error.response.config.url.includes('addresses') &&
+          error.response.data.message.includes('address not found')) hasBAError = true
+      })
+
+      let badAddress = { data: { businessOffice: null } }
+      if (!data && hasBAError) this.storeAddresses(badAddress) // if 404 and business address not found
+      if (!data && !hasBAError) throw new Error('Incomplete business data')
+      if (data) this.storeAddresses(data)
```
severinbeauvais, line 431: This is complicated to review. I'd prefer to see...
```
if situation 1 {
  handle situation 1
} else if situation 2 {
  handle situation 2
} else {
  handle fallback/common situation, if applicable
}
```

severinbeauvais: Other simplification ideas:

Since `fetchAddresses()` normally returns address data, why not simply return alternate address data in the catch block? Then you would only need to store the addresses after this block.

You could also move your `throw` into the catch block.

tests/unit/FirmsAddressList.spec.ts
```diff
@@ -163,12 +163,16 @@ describe('FirmsAddressList', () => {
...
   })
 
-  it('displays "Not Enter" for null address', async () => {
+  it('displays "(Not Entered)" for null address', async () => {
     // init store
     store.state.businessAddress = {
       deliveryAddress: null,
       mailingAddress: null
     }
+    store.state.getBusinessAddress = {
```
severinbeauvais, line 172: `getBusinessAddress` is not a state property.

Are you trying to override a store getter? See https://vuex.vuejs.org/guide/getters.html#getters

...Or you could mock it.

...Or, why not set the state the way you want it on lines 168-170?

src/App.vue
severinbeauvais: Or "fetchStoreAddressData"... whatever is consistent with the other functions in here.

src/components/Dashboard/FirmsAddressList.vue
lewischenstudio: `this.getBusinessAddress` will be one of the four samples:
`this.getBusinessAddress = null` (case 1)
`this.getBusinessAddress = { deliveryAddress: null }` (case 2)
`this.getBusinessAddress = { deliveryAddress: {} }` (case 3)
`this.getBusinessAddress = { deliveryAddress: { actions ... }` (case 4, correct data)

I am using `Object.values()` for the case 3.

tests/unit/FirmsAddressList.spec.ts
lewischenstudio: I will create another test case for getBusinessAddress and use adifferent state.businessAddress.

src/App.vue
lewischenstudio: I think it's because this method is a Promise. When I use .catch directly, it will not work.

src/components/Dashboard/FirmsAddressList.vue
severinbeauvais: Thanks for the scenario. Very helpful.

I counsel against using Object.values(). I think you should use `isEmpty()` for case 3.

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -133,7 +133,11 @@ export default class FirmsAddressList extends Mixins(CommonMixin, CountriesProvi
   @Getter getBusinessAddress!: OfficeAddressIF
 
   // Business
-  get businessAddress (): OfficeAddressIF {
+  get businessAddress (): OfficeAddressIF|null {
+    let deliveryAddress = this.getBusinessAddress?.deliveryAddress
+    if (this.getBusinessAddress === null) return null
```
severinbeauvais, line 138: 'deliveryAddress' will be undefined if the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) fails.

Also, these are good cases for just checking if [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) or [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).

src/components/Dashboard/FirmsAddressList.vue
```diff
@@ -133,7 +133,11 @@ export default class FirmsAddressList extends Mixins(CommonMixin, CountriesProvi
...
+    if (deliveryAddress === null) return null
+    if (Object.values(deliveryAddress).length === 0) return null
```
severinbeauvais, line 140: Again, please use isEmpty() instead of object values.

Rationale: Object.values() is looking deeply into the object's implementation, which we don't need/want to do at our "high level Typescript code" level.

src/App.vue
lewischenstudio: I changed it to "fetchStoreAddressData".

src/App.vue
severinbeauvais: OK, I'll look at this later.

src/App.vue
severinbeauvais: Thanks, much easier to review/understand :)

src/components/Dashboard/FirmsAddressList.vue
lewischenstudio: I agree. I changed it to isEmpty() and it's very powerful.

src/components/Dashboard/FirmsAddressList.vue
lewischenstudio: Thanks. I've changed it to isEmpty(). Simple, elegant and powerful.

src/components/Dashboard/FirmsAddressList.vue
severinbeauvais: Obsolete code. Resolving.

severinbeauvais (approved):
LGTM

PR #330, opened 2022-05-20
Update business dashboard/11660

meganawong:
*Issue #:* /bcgov/entity#11660

*Description of changes:*
* Changed the text from "voluntary dissolution" to "dissolution" for SP/GP in the business dashboard header, filing history and on respective buttons.


src/components/dialogs/ConfirmDissolution.vue
```diff
@@ -13,11 +13,14 @@
         {{getModalTitle}}
         </p>
         <p class="warning-text" id="dialog-text">
-          You are about to voluntarily dissolve <strong>{{ getEntityName }}</strong>;
+          You are about to {{ dissolutionType }} <strong>{{ getEntityName }}</strong>;
           once this process is completed and the required documents are filed, the {{ entityTitle }} will be
           struck from the register and dissolved, ceasing to be an incorporated {{ subEntityTitle }} under the
           {{ entityAct }} Act.
         </p>
+        <p class="warning-text" id="dialog-text">
```
shabeeb-aot, line 21: put inside v-if so, p tag wont be there when there is no additional label

src/interfaces/dissolution-confirmation-interface.ts
```diff
@@ -4,4 +4,6 @@ export interface DissolutionConfirmationResourceIF {
   act: string
   modalTitle: string
   confirmButtonText:string
+  additionalLabel: string
```
shabeeb-aot, line 7: use like additionalLabel?: string,  I think this will optional field for Sp/GP,
same for dissolutionType

src/components/Dashboard/TodoList.vue
```diff
@@ -372,7 +372,7 @@
                       >
                         <v-icon class="pr-1" color="primary" size="18px">mdi-delete-forever</v-icon>
                         <template v-if="isTypeDissolution(item)">
-                          <v-list-item-title>Delete Voluntary Dissolution</v-list-item-title>
+                          <v-list-item-title>Delete Dissolution</v-list-item-title>
```
shabeeb-aot, line 375: This need to come from config file

src/enums/dissolutionTypesNames.ts
```diff
@@ -28,6 +28,7 @@ export enum DissolutionNames {
   // COURT_ORDER_LIQUIDATION = 'Court Ordered Liquidation',
   // INVOLUNTARY = 'Involuntary Dissolution',
   VOLUNTARY = 'Voluntary Dissolution',
+  FIRM = 'Dissolution',
```
shabeeb-aot, line 31: instead of FIRM may be FIRM _DISSOLUTION

shabeeb-aot, line 31: instead of FIRM may be FIRM _DISSOLUTION

severinbeauvais:
Please update the app version and check in the 2 package files.

severinbeauvais:
Please update the app version and check in the 2 package files.

severinbeauvais:
Need to fix package file conflicts.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -474,7 +476,7 @@ export default class FilingHistoryList extends Mixins(
       const item: HistoryItemIF = {
         availableOnPaperOnly: filing.availableOnPaperOnly,
         commentsCount: filing.commentsCount,
-        displayName: filing.displayName,
+        displayName: this.isTypeDissolution(filing) ? this.getFilingHistoryListResource?.title : filing.displayName,
```
severinbeauvais, line 479: Don't do this. The API is supposed to provide the display name in all cases. Change the API instead.

src/mixins/enum-mixin.ts
```diff
@@ -227,9 +227,10 @@ export default class EnumMixin extends Vue {
    * @param type the dissolution type to convert
    * @returns the dissolution name
    */
-  dissolutionTypeToName (type: DissolutionTypes): string {
+  dissolutionTypeToName (type: DissolutionTypes, isFirm: boolean = false): string {
     switch (type) {
-      case DissolutionTypes.VOLUNTARY: return DissolutionNames.VOLUNTARY
+      case DissolutionTypes.VOLUNTARY:
+        return isFirm ? DissolutionNames.FIRM_DISSOLUTION : DissolutionNames.VOLUNTARY
```
severinbeauvais, line 233: Ideally there would be a new dissolution type for "firm voluntary", but I guess the API doesn't do that. So I'm OK with having this special case here.

However, I think you should import the `isFirm` getter in this mixin instead of passing it in as an optional parameter. I think it will be just a bit cleaner.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Rationale: Special cases are the killer of maintainable applications. The ledger design is for the API to say what to display here.

severinbeauvais (changes requested):
Most of my comments are small, but I strongly disagree with the special case design implemented here.

src/components/Dashboard/FilingHistoryList.vue
seeker25: https://github.com/bcgov/lear/blob/859dd918f06dbdb11b81622a2ef72a90b3e1011e/legal-api/src/legal_api/core/meta/filing.py#L167 I think?

severinbeauvais: Yes I think that's it.

I'm wondering if it would make sense to return a different code for SP and GP as well? (I'm assuming this is the dissolution type but I could be wrong.) For future conversion of type to name (in enum mixin).

@argush3 What do you think?

src/App.vue
```diff
@@ -210,7 +210,7 @@ export default {
   },
 
   computed: {
-    ...mapGetters(['getIdentifier', 'getEntityName', 'getEntityType', 'isRoleStaff']),
+    ...mapGetters(['getIdentifier', 'getEntityName', 'getEntityType', 'isRoleStaff', 'isFirm']),
```
severinbeauvais, line 213: No longer needed?

src/mixins/enum-mixin.ts
severinbeauvais: @meganawong Please let the author resolve their comments. That way they know to check back and see if the code change is good. (Otherwise reviewers need to review all the code again.)

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: PS Travis, the API also determines the name of the outputs (PDFs), which probably need the same change.

argush3: Yes, we should update the API for this.

src/App.vue
meganawong: removed

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I mean different reason text in the Entity Info header depending on the type of dissolution.

See src/mixins/enum-mixin.ts in this PR and [here where it's used](https://github.com/bcgov/business-filings-ui/blob/0c6eb4b3fc29f46a1fc7a6091e403070bf9af9ff/src/App.vue#L581).

src/components/dialogs/ConfirmDissolution.vue
meganawong: resolved

src/interfaces/dissolution-confirmation-interface.ts
meganawong: resolved

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Code was reverted.

severinbeauvais (approved):
I approve merging these changes.

shabeeb-aot (approved):

PR #333, opened 2022-05-25
Feature/11662

shabeeb-aot:
*Issue #:* /bcgov/entity#11662

*Description of changes:*
* Added Action required component

### DIFF
--- a/src/components/Dashboard/ActionRequired.vue
+++ b/src/components/Dashboard/ActionRequired.vue
@@ -30,7 +30,7 @@ import { ContactInfo } from '@/components/common'
     ContactInfo
   }
 })
-export default class LegalObligation extends Vue {
+export default class ActionRequired extends Vue {
   public panel: number = 1;
 
   public togglePanel (): void {
--- a/src/components/Dashboard/ActionRequired.vue
+++ b/src/components/Dashboard/ActionRequired.vue
@@ -1,0 +1,56 @@
+<template>
+  <v-expansion-panels v-model="panel">
+    <v-expansion-panel class="action-required-container">
+      <v-card flat class="px-6 py-5">
+        <h3>Action Required</h3>
+        <p class="mt-1">
+          BC Registries is missing information about your business (e.g.. business start date, nature of business,
+          business address etc.). Please contact BC Registries to input any missing business information. Missing
+          information must be entered before toy can file changes or dissolve this business.
+          <v-btn class="details-btn" color="primary" text :ripple="false" @click.stop="togglePanel">
+            <v-icon left color="primary">mdi-information-outline</v-icon>
+            <span color="primary">{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
+          </v-btn>
+        </p>
+        <v-expansion-panel-content class="action-required-contact-info">
+          <p>BC Registries Contact Information:</p>
+          <ContactInfo class="mt-4" />
+        </v-expansion-panel-content>
+      </v-card>
+    </v-expansion-panel>
+  </v-expansion-panels>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+import { ContactInfo } from '@/components/common'
+
+@Component({
+  components: {
+    ContactInfo
+  }
+})
+export default class LegalObligation extends Vue {
+  public panel: number = 1;
+
+  public togglePanel (): void {
+    this.panel = this.panel === 1 ? 0 : 1
+  }
+}
+</script>
+
+<style lang="scss">
+.action-required-contact-info .v-expansion-panel-content__wrap {
+  padding: 0 0 16px !important;
+}
+</style>
+
+<style lang="scss" scoped>
+@import "@/assets/styles/theme.scss";
+
+.action-required-container {
+  display: flex;
+  justify-content: center;
+  margin-bottom: 1.5rem;
+}
+</style>
--- a/src/components/Dashboard/ActionRequired.vue
+++ b/src/components/Dashboard/ActionRequired.vue
@@ -1,6 +1,6 @@
 <template>
   <v-expansion-panels v-model="panel">
-    <v-expansion-panel class="action-required-container">
+    <v-expansion-panel class="action-required-container d-flex justify-center mb-6">
       <v-card flat class="px-6 py-5">
         <h3>Action Required</h3>
         <p class="mt-1">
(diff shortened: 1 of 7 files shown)


src/components/Dashboard/ActionRequired.vue
```diff
@@ -0,0 +1,56 @@
...
+export default class LegalObligation extends Vue {
+  public panel: number = 1;
+
+  public togglePanel (): void {
+    this.panel = this.panel === 1 ? 0 : 1
+  }
+}
+</script>
+
+<style lang="scss">
```
shabeeb-aot, line 42: Need to remove global padding v-expansion-panel-content__wrap which come in contact section

src/components/Dashboard/ActionRequired.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+  <v-expansion-panels v-model="panel">
+    <v-expansion-panel class="action-required-container">
+      <v-card flat class="px-6 py-5">
+        <h3>Action Required</h3>
+        <p class="mt-1">
+          BC Registries is missing information about your business (e.g.. business start date, nature of business,
```
severinbeauvais, line 7: Typo. Should be `e.g.,` (confirm with UI/UX).

src/components/Dashboard/ActionRequired.vue
```diff
@@ -0,0 +1,56 @@
...
+          business address etc.). Please contact BC Registries to input any missing business information. Missing
```
severinbeauvais, line 8: "to input" doesn't sound right. How about "to enter"? (confirm with UI/UX)

src/components/Dashboard/ActionRequired.vue
```diff
@@ -0,0 +1,56 @@
...
+          information must be entered before toy can file changes or dissolve this business.
```
severinbeauvais, line 9: Typo `toy` -> `you`.

src/components/Dashboard/ActionRequired.vue
shabeeb-aot: Screen Shot 2022-05-26 at 11 14 24 AM

tests/unit/ActionRequired.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    const wrapper = mount(ActionRequired, {
+      vuetify
+
+    })
+    const paragraphs1 = wrapper.findAll('p')
+    expect(paragraphs1.length).toBe(1)
+    expect(wrapper.find(ContactInfo).exists()).toBe(false)
+    // click the view button
+    wrapper.find('.details-btn').trigger('click')
+    await Vue.nextTick()
```
severinbeauvais, line 38: Check if you can await the trigger function on the previous line instead. (Cleaner code.)

src/mixins/allowable-actions-mixin.ts
```diff
@@ -32,8 +33,14 @@ export default class AllowableActionsMixin extends Vue {
...
-        return (!this.isHistorical && !this.hasBlocker && !!businessId &&
+        const isDissolveAllowed = (!this.isHistorical && !!businessId &&
           !!getFeatureFlag('supported-dissolution-entities')?.includes(this.getEntityType))
+        // if its not SP/GP , then consider hasBlocker flag (existing)
+        if (!this.isFirm) {
+          return isDissolveAllowed && !this.hasBlocker
+        }
+
+        return isDissolveAllowed
       }
```
severinbeauvais, line 44: Looks good.

Since you're now familiar with this code, I should tell you that this was intended to be an "initial implementation". This is business logic and should be in the BE instead. The idea is that the API would return, say, an array of what actions are allowed. However, this is not been fully implemented yet. If you have any thoughts on this, speak up or create a ticket or whatever. I'd still like this business logic to exist in the BE only.

PS Please update the unit tests for this mixin to verify your changes.

tests/unit/ActionRequired.spec.ts
shabeeb-aot: I made the change but it's showing some warning in my code base,
So i think its better to wait for nextTick

Screen Shot 2022-05-26 at 2 21 24 PM

src/mixins/allowable-actions-mixin.ts
shabeeb-aot: Absolutely I agree with you, business logic should be in BE

tests/unit/ActionRequired.spec.ts
severinbeauvais: Weird. I don't get that warning and my VS Code shows that `trigger` can return a promise.

image

Anyway, it's fine.

src/components/Dashboard/ActionRequired.vue
```diff
@@ -0,0 +1,56 @@
...
+</style>
+
+<style lang="scss" scoped>
+@import "@/assets/styles/theme.scss";
+
+.action-required-container {
+  display: flex;
+  justify-content: center;
+  margin-bottom: 1.5rem;
+}
```
severinbeauvais, line 55: Next time, you could use Vuetify classes to do all of this right in the template.

(It's a gray area whether it's better to specify this in the template or style section. The advantage of having it in the template is that is makes layout/whitespace changes easier.)

Example: `d-flex justify-center mb-6`

Refs:
https://vuetifyjs.com/en/styles/flex/
https://vuetifyjs.com/en/styles/spacing/

src/components/Dashboard/TodoList.vue
```diff
@@ -26,8 +26,8 @@
       @okay="resetCancelPaymentErrors()"
       attach="#todo-list"
     />
-
-    <v-expansion-panels v-if="todoItems && todoItems.length > 0" accordion  v-model="panel">
+    <ActionRequired v-if="isStaffActionRequired"/>
```
severinbeauvais, line 29: I see that you're conditionally showing this new expansion panel.

However, I believe the API is providing a "todo task" when a business is out of compliance. Could you use that instead?

You can check with Argus on the details of the todo task item.

You can also see that the Todo List component is not handling this task (it should, even if it deliberately ignores it):

image

severinbeauvais: Faiz and I discussed this and agreed to go with this "special case" implementation for now due to time constraints.

However, we should refactor this some time to fit into the existing "todo" item functionality.

Perhaps when the staff version of this is implemented, this can be refactored accordingly.

severinbeauvais (approved):
OK for now.

PR #334, opened 2022-05-27
Updating the To-Do for clients in case of missing Business Informatio…

shabeeb-aot:
…n#11662 | staff version updated and some wording changes

*Issue #:* /bcgov/entity#11662
*Issue #:* /bcgov/entity#12405

*Description of changes:*
* show action required for staff
* company word change to business 

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -698,7 +698,7 @@ export default class TodoList extends Mixins(
       }
     }
 
-    // Add todo items count +1 when SP/GP with compaince warning
+    // Add todo items count +1 when SP/GP with compliance warning
     const todoLength = this.isActionRequired ? this.todoItems.length + 1 : this.todoItems.length
     // report number of items back to parent (dashboard)
     this.$emit('todo-count', todoLength)
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -666,7 +666,7 @@ export default class TodoList extends Mixins(
     return this.requiresAlteration ? 'Alter Now' : 'Resume'
   }
 
-  /** show action required only for SP/GP with compaince warning. */
+  /** Show action required only for SP/GP with Compliance warning. */
   get isStaffActionRequired (): boolean {
     return this.isFirm && this.isNotInCompliance
   }
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -668,7 +668,7 @@ export default class TodoList extends Mixins(
 
   /** show action required only for SP/GP with compaince warning. */
   get isStaffActionRequired (): boolean {
-    return this.isFirm && this.isNotInCompliance && !this.isRoleStaff
+    return this.isFirm && this.isNotInCompliance
   }
   get showTodoPanel () {
     return this.todoItems && this.todoItems.length > 0
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -26,7 +26,7 @@
       @okay="resetCancelPaymentErrors()"
       attach="#todo-list"
     />
-    <ActionRequired v-if="isStaffActionRequired"/>
+    <ActionRequired v-if="isActionRequired"/>
     <v-expansion-panels v-if="showTodoPanel" accordion  v-model="panel">
       <v-expansion-panel
         class="align-items-top todo-item px-6 py-5"
@@ -515,7 +515,7 @@
     </v-expansion-panels>
 
     <!-- No Results Message -->
-    <v-card class="no-results" flat v-if="!showTodoPanel && !isStaffActionRequired">
+    <v-card class="no-results" flat v-if="!showTodoPanel && !isActionRequired">
       <v-card-text>
         <div class="no-results__title">You don't have anything to do yet</div>
         <div class="no-results__subtitle">Filings that require your attention will appear here</div>
@@ -667,7 +667,7 @@ export default class TodoList extends Mixins(
   }
 
   /** Show action required only for SP/GP with Compliance warning. */
-  get isStaffActionRequired (): boolean {
+  get isActionRequired (): boolean {
     return this.isFirm && this.isNotInCompliance
   }
   get showTodoPanel () {
@@ -699,7 +699,7 @@ export default class TodoList extends Mixins(
     }
 
     // Add todo items count +1 when SP/GP with compaince warning
-    const todoLength = this.isStaffActionRequired ? this.todoItems.length + 1 : this.todoItems.length
+    const todoLength = this.isActionRequired ? this.todoItems.length + 1 : this.todoItems.length
     // report number of items back to parent (dashboard)
     this.$emit('todo-count', todoLength)
 
--- a/src/components/EntityInfo.vue
+++ b/src/components/EntityInfo.vue
@@ -44,7 +44,7 @@
                   @click="promptChangeCompanyInfo()"
                 >
                   <v-icon medium>mdi-file-document-edit-outline</v-icon>
-                  <span class="font-13 ml-1">View and Change Company Information</span>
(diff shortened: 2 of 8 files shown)


src/components/EntityInfo.vue
```diff
@@ -54,34 +54,32 @@
...
                   <img src="@/assets/images/Dissolution_Header_Icon.svg" alt="" class="pa-1">
                   <span class="font-13 ml-1">Dissolve this Business</span>
-                  <v-tooltip top content-class="top-tooltip" transition="fade-transition" nudge-right="7">
-                    <template v-slot:activator="{ on }">
-                      <span class="pl-1" v-on="on">
-                        <v-icon size="1rem">mdi-information-outline</v-icon>
-                      </span>
-                    </template>
-                    Dissolving the {{ entityTitle }} will make this {{ entityTitle }} historical
+                </v-btn>
```
shabeeb-aot, line 78: Now we don't need info icon to dissolve the business, tooltip will appear on hover of text or main icon

src/store/getters.ts
```diff
@@ -36,10 +36,15 @@ export default {
   },
 
   /** Is True if there are any pending tasks or filings. */
-  hasBlocker (state: StateIF, getters): boolean {
-    let blocker = (state.hasBlockerTask || state.hasBlockerFiling || state.isCoaPending)
+  hasBlockerExceptStaffApproval (state: StateIF): boolean {
```
shabeeb-aot, line 39: created a new getter like before , and used this to avoid conversion menu disable, Let me know if we have other ways to do it

src/components/Dashboard/TodoList.vue
```diff
@@ -668,7 +668,7 @@ export default class TodoList extends Mixins(
 
   /** show action required only for SP/GP with compaince warning. */
```
severinbeauvais, line 669: Typo.
Also capitalize. (This is a JSDoc comment that is displayed when you hover on the getter.)

src/components/EntityInfo.vue
```diff
@@ -44,7 +44,7 @@
                   @click="promptChangeCompanyInfo()"
                 >
                   <v-icon medium>mdi-file-document-edit-outline</v-icon>
-                  <span class="font-13 ml-1">View and Change Company Information</span>
+                  <span class="font-13 ml-1">View and Change Business Information</span>
```
severinbeauvais, line 47: This change affects all entity types. Is it approved by UI/UX team?

src/components/EntityInfo.vue
severinbeauvais: "all types of entities"

(A corp is a type of entity. A coop is another. Firm, etc.)

severinbeauvais:
Please fix broken unit test.

Also please add some tests for the Allowable Actions mixin and other conditions that you've updated.

src/store/getters.ts
severinbeauvais: This is a special case that might get more and complicated over time.

What if you copied `(state.hasBlockerTask || state.hasBlockerFiling || state.isCoaPending)` into the allowable actions case (instead of using the global getter)? (You will also have to update the global getter.) I think this might be cleaner.

shabeeb-aot: Have to use those in two places, To avoid code duplicate , I used as a getter

src/components/EntityInfo.vue
shabeeb-aot: Yeah, UI/UX team approved it

Screen Shot 2022-05-27 at 3 01 15 PM

src/components/Dashboard/TodoList.vue
```diff
@@ -666,9 +666,9 @@ export default class TodoList extends Mixins(
     return this.requiresAlteration ? 'Alter Now' : 'Resume'
   }
 
-  /** show action required only for SP/GP with compaince warning. */
+  /** Show action required only for SP/GP with Compliance warning. */
   get isStaffActionRequired (): boolean {
-    return this.isFirm && this.isNotInCompliance && !this.isRoleStaff
+    return this.isFirm && this.isNotInCompliance
```
severinbeauvais, line 671: It's not a staff action any more :)

src/components/EntityInfo.vue
severinbeauvais: Woohoo! (I've been asking for this for months.)

tests/unit/EntityInfo.spec.ts
```diff
@@ -366,13 +366,12 @@ describe('EntityInfo - Click Tests - Dissolutions', () => {
 
   it('displays the Dissolve this Company button', async () => {
     // mount the component and wait for everything to stabilize
-    const wrapper = shallowMount(EntityInfo, {
+    const wrapper = mount(EntityInfo, {
```
severinbeauvais, line 369: This shouldn't be needed since EntityInfo doesn't need any sub-component to be rendered (stubs should be enough). What error does this fix? A Vuetify component maybe?

src/store/getters.ts
```diff
@@ -36,10 +36,15 @@ export default {
...
+    return (state.hasBlockerTask || state.hasBlockerFiling || state.isCoaPending)
+  },
+
+  /** Is True if there are any pending tasks or filings. */
+  hasBlocker (_state: StateIF, getters): boolean {
```
severinbeauvais, line 44: What if you only had this getter, and in it you check if user is staff or not?

tests/unit/EntityInfo.spec.ts
shabeeb-aot: since I moved, the button inside the tooltip, shallowMount will not render the button inside vuetify, Else test case will fail, button will not exist

src/components/EntityInfo.vue
shabeeb-aot: yes, for all entity

src/store/getters.ts
shabeeb-aot: hasBlocker method is used in many places to disable action and show html, We cannot check user is staff or not because, we need the same behaviour for staff and use except one place ```case AllowableActions.FILE_STAFF_NOTATION:```  There we need to show button

tests/unit/EntityInfo.spec.ts
severinbeauvais: Ah, makes sense.

src/store/getters.ts
severinbeauvais: Yes, you're right.

How about this?

allowable-actions-mixin.ts
```
      case AllowableActions.FILE_STAFF_NOTATION: {
        const blocker = (state.hasBlockerTask || state.hasBlockerFiling || state.isCoaPending)
        return (!this.isHistorical && !blocker && !!businessId && this.isRoleStaff)
      }
```

getters.ts
```
  /** Is True if there are any pending tasks or filings or the firm is not compliance. */
  hasBlocker (_state: StateIF, getters): boolean {
    let blocker = (state.hasBlockerTask || state.hasBlockerFiling || state.isCoaPending)
    // check for SP and GP not in compliance
    if (getters.isFirm && getters.isNotInCompliance) {
      blocker = true
    }
    return blocker
  }
```

src/components/Dashboard/TodoList.vue
```diff
@@ -699,7 +699,7 @@ export default class TodoList extends Mixins(
     }
 
     // Add todo items count +1 when SP/GP with compaince warning
```
severinbeauvais, line 701: Another typo 😋

src/store/getters.ts
shabeeb-aot: In this case, I think we have to update in both places when something need to disable

severinbeauvais: Yes but I think it's cleaner overall.

If it was me, I would do it like this and add comments in each place referring to the other place.

Another way to think of this is in terms of compartmentalization:
- the allowable mixins checks only what it cares about
- the getter checks only what it cares about
- the mixin and getter don't depend on each other (except by code comments)

severinbeauvais (approved):

PR #335, opened 2022-05-27
Updating the entity info on business registry/12370

meganawong:
*Issue #:* /bcgov/entity#12370

*Description of changes:*

* Update the entity info header for the business registry dashboard to include registration date and registration number, and remove incorporation number and name request number for SP/GP


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.1.2",
+  "version": "5.1.4",
```
severinbeauvais, line 3: You should only increase by 1.

I know: your other PR might get merged first. But if it does then you'll need to rebase anyway, so this isn't saving you any work. (Sorry.)

src/components/EntityInfo.vue
```diff
@@ -108,20 +108,34 @@
...
             <!-- Incorporation Number -->
-            <template v-if="businessId">
+            <template v-if="businessId && !isFirm">
               <dt class="mr-2">Incorporation Number:</dt>
               <dd id="entity-incorporation-number">{{ getIdentifier || 'Not Available' }}</dd>
             </template>
 
             <!-- NR Number -->
-            <template v-if="getNrNumber">
+            <template v-if="getNrNumber && !isFirm">
```
severinbeauvais, line 138: I think this code is to display the NR number when a user is registering the firm (in other words, the entity has not been filed and does not have a business ID yet; it is using a temporary registration number).

severinbeauvais: If this is not clear then I can provide some examples.

meganawong: I see, so as long as there is an NR number, it should be displayed?

package.json
meganawong: reverted back to 5.1.3

src/components/EntityInfo.vue
severinbeauvais: Actually, I confused "NR number" with "temporary registration number". Sorry!

Here, we are looking at the "NR number". I believe this should only be displayed when the user is still registering/incorporating their business (ie, while this page uses temp reg number to load data instead of the permanent entity id). Once the business is registered/incorporated then we don't care about the NR number.

And I believe the NR number is only available before this business is registered/incorporated.

Therefore, I think the old logic is fine: just display it if it's available. (And it should display for all registrations/incorporations with a NR, not just firms.) Let me know if you have more info on this.

src/components/EntityInfo.vue
```diff
@@ -108,14 +108,28 @@
...
           <dl>
+
+            <!-- Registration Date -->
+            <template v-if="businessId && isFirm">
+              <dt class="mr-2">Registration Date:</dt>
+              <dd id="entity-business-registration-date">
+                {{ this.dateToPacificDate(getEntityFoundingDate) || 'Not Available' }}</dd>
+            </template>
+
+            <!-- Registration Number -->$
```
severinbeauvais, line 119: Typo at end of line.

src/components/EntityInfo.vue
```diff
@@ -108,14 +108,28 @@
...
+            <template v-if="businessId && isFirm">
```
severinbeauvais, line 120: Is the design to show both Registration Number and Business Number?

I _think_ those might be the same number with a different name. If so then you should not rendered both (ie, add condition to line 126).

meganawong: The design includes both but I will confirm whether the numbers represent the same thing or not

meganawong: Jyoti confirmed that both the Registration Number and Business Number should be displayed

shabeeb-aot (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.1.3",
+  "version": "5.1.5",
```
severinbeauvais, line 3: I don't know why GH isn't showing the latest version (per the main branch) on the left side of the diff, but your change here is indeed correct.

After this PR is merged, please confirm that your change is reflected in main branch.

severinbeauvais (commented):
Looks good so. Just 1 comment to make future development easier.

package.json
meganawong: will do

src/components/EntityInfo.vue
```diff
@@ -106,14 +106,28 @@
...
             <template v-if="businessId">
               <dt class="mr-2">Business Number:</dt>
               <dd id="entity-business-number">{{ getBusinessNumber || 'Not Available' }}</dd>
             </template>
 
             <!-- Incorporation Number -->
-            <template v-if="businessId">
+            <template v-if="businessId && !isFirm">
               <dt class="mr-2">Incorporation Number:</dt>
               <dd id="entity-incorporation-number">{{ getIdentifier || 'Not Available' }}</dd>
```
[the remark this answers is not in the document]
meganawong: The only thing is that in the designs, Registration Number should display above Business Number whereas Incorporation Number should display below Business Number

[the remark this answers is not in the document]
severinbeauvais: OK, then this is how it must be. Resolving.

severinbeauvais (approved):

PR #336, opened 2022-05-30
UI/UX issues#12405

shabeeb-aot:
*Issue #:* /bcgov/entity#12405

*Description of changes:*
* Ui/UX issue fix
*  spacing adjusted
* disabled button and icon colour changed to blue from gray

### DIFF
--- a/src/assets/styles/overrides.scss
+++ b/src/assets/styles/overrides.scss
@@ -124,3 +124,12 @@
   border-left: 10px solid transparent !important;
   border-top: 8px solid RGBA(73,80,87,0.95) !important;
 }
+
+.theme--light.v-btn.v-btn--disabled.v-btn--text {
+  opacity: 0.4 !important;
+  color: $app-blue !important;
+
+  .v-icon {
+    color: $app-blue !important;
+  }
+}
--- a/src/components/Dashboard/ActionRequired.vue
+++ b/src/components/Dashboard/ActionRequired.vue
@@ -7,9 +7,9 @@
           BC Registries is missing information about your business (e.g., business start date, nature of business,
           business address etc.). Please contact BC Registries to input any missing business information. Missing
           information must be entered before you can file changes or dissolve this business.
-          <v-btn class="details-btn" color="primary" text :ripple="false" @click.stop="togglePanel">
-            <v-icon left color="primary">mdi-information-outline</v-icon>
-            <span color="primary">{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
+          <v-btn class="details-btn pl-1" outlined color="primary"  :ripple="false" @click.stop="togglePanel">
+            <v-icon left color="primary" class="mr-0">mdi-information-outline</v-icon>
+            <span color="primary" >{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
           </v-btn>
         </p>
         <v-expansion-panel-content class="action-required-contact-info">
@@ -43,5 +43,14 @@ export default class ActionRequired extends Vue {
 @import "@/assets/styles/theme.scss";
 .action-required-contact-info .v-expansion-panel-content__wrap {
   padding: 0 0 16px !important;
+}
+</style>
+
+<style lang="scss" scoped>
+
+.details-btn {
+  margin-bottom: 0.25rem;
+  height: inherit !important;
+  border: none;
 }
 </style>
--- a/src/components/Dashboard/ActionRequired.vue
+++ b/src/components/Dashboard/ActionRequired.vue
@@ -9,7 +9,7 @@
           information must be entered before you can file changes or dissolve this business.
           <v-btn class="details-btn pl-1" outlined color="primary"  :ripple="false" @click.stop="togglePanel">
             <v-icon left color="primary" class="mr-0">mdi-information-outline</v-icon>
-            <span color="primary" >{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
+            <span color="primary">{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
           </v-btn>
         </p>
         <v-expansion-panel-content class="action-required-contact-info">
--- a/src/components/Dashboard/ActionRequired.vue
+++ b/src/components/Dashboard/ActionRequired.vue
@@ -49,8 +49,7 @@ export default class ActionRequired extends Vue {
 <style lang="scss" scoped>
 
 .details-btn {
-  margin-bottom: 0.25rem;
-  height: inherit !important;
   border: none;
+  margin-top: -6px;
 }
 </style>
--- a/src/components/Dashboard/CustodianListSm.vue
+++ b/src/components/Dashboard/CustodianListSm.vue
@@ -88,8 +88,8 @@ $avatar-width: 2.75rem;
 
 // Complete filing required styling
 .complete-filing {
-  padding: 1.1rem;
-  color: $gray7;
+  padding: 2rem;
+  color: $gray6;
   font-size: $px-14;
   white-space: pre-wrap;
   display: inline-block;
--- a/src/components/Dashboard/CustodianListSm.vue
+++ b/src/components/Dashboard/CustodianListSm.vue
@@ -88,8 +88,8 @@ $avatar-width: 2.75rem;
 
 // Complete filing required styling
 .complete-filing {
-  padding: 2rem;
-  color: $gray6;
+  padding: 1.1rem;
+  color: $gray7;
   font-size: $px-14;
   white-space: pre-wrap;
   display: inline-block;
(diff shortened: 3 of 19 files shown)


src/components/Dashboard/CustodianListSm.vue
```diff
@@ -88,8 +88,8 @@ $avatar-width: 2.75rem;
 
 // Complete filing required styling
 .complete-filing {
-  padding: 2rem;
-  color: $gray6;
+  padding: 1.1rem;
```
severinbeauvais, line 91: Don't do this. It's going away from our standardized spacing in this UI.

src/components/Dashboard/ProprietorPartnersListSm.vue
```diff
@@ -147,11 +147,12 @@ $avatar-width: 2.75rem;
 
 // Complete filing required styling
 .complete-filing {
-  padding: 2rem;
-  color: $gray6;
+  padding: 1.2rem;
```
severinbeauvais, line 150: Don't do this. It's going away from our standardized spacing in this UI.

src/components/Dashboard/ProprietorPartnersListSm.vue
```diff
@@ -147,11 +147,12 @@ $avatar-width: 2.75rem;
...
+  color: $gray7;
   font-size: $px-14;
   white-space: pre-wrap;
   display: inline-block;
+  // padding-left: 1.1rem;
```
severinbeauvais, line 155: Delete if not being used.

src/views/Dashboard.vue
```diff
@@ -295,4 +296,11 @@ section header {
 .pending-tooltip {
   max-width: 16.5rem;
 }
+
+#dashboard {
+  & .theme--light.v-btn.v-btn--disabled.v-btn--text, .theme--light.v-btn.v-btn--disabled .v-icon {
+  opacity: 0.4 !important;
+  color: $app-blue !important;
+}
+}
```
severinbeauvais, line 305: With this, there will be 3 places where this style is declared.

Do something like this instead:
https://github.com/bcgov/business-create-ui/blob/703aed21cb5c00fc3d4db467e6094c1be5213788/src/assets/styles/overrides.scss#L18

If not, this UI will be even harder to maintain in the future.

severinbeauvais: Please clean up instead of making more special cases.

src/components/Dashboard/DirectorListSm.vue
```diff
@@ -116,8 +116,8 @@ $avatar-width: 2.75rem;
 
 // Complete filing required styling
 .complete-filing {
-  padding: 2rem;
-  color: $gray6;
+  padding: 1.1rem;
```
severinbeauvais, line 119: Don't do this. It's going away from our standardized spacing in this UI.

src/components/Dashboard/ActionRequired.vue
```diff
@@ -7,9 +7,9 @@
           BC Registries is missing information about your business (e.g., business start date, nature of business,
           business address etc.). Please contact BC Registries to input any missing business information. Missing
           information must be entered before you can file changes or dissolve this business.
-          <v-btn class="details-btn" color="primary" text :ripple="false" @click.stop="togglePanel">
-            <v-icon left color="primary">mdi-information-outline</v-icon>
-            <span color="primary">{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
+          <v-btn class="details-btn pl-1" outlined color="primary"  :ripple="false" @click.stop="togglePanel">
+            <v-icon left color="primary" class="mr-0">mdi-information-outline</v-icon>
+            <span color="primary" >{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
```
severinbeauvais, line 12: The added space here does nothing. Please remove it.

src/components/Dashboard/ProprietorPartnersListSm.vue
shabeeb-aot: I was discussing with tracey, to fix alignment, May be i can leave it as it is 
Screen Shot 2022-05-30 at 2 19 14 PM

src/components/Dashboard/ActionRequired.vue
```diff
@@ -7,9 +7,9 @@
           BC Registries is missing information about your business (e.g., business start date, nature of business,
           business address etc.). Please contact BC Registries to input any missing business information. Missing
           information must be entered before you can file changes or dissolve this business.
-          <v-btn class="details-btn" color="primary" text :ripple="false" @click.stop="togglePanel">
-            <v-icon left color="primary">mdi-information-outline</v-icon>
-            <span color="primary">{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
+          <v-btn class="details-btn pl-1" outlined color="primary"  :ripple="false" @click.stop="togglePanel">
+            <v-icon left color="primary" class="mr-0">mdi-information-outline</v-icon>
```
severinbeauvais, line 11: Why do you need to remove right margin here? Why do this icon and button have different styling than the exact same buttons in TodoList.vue?

We are already starting to see maintainability problems with duplicated code. This is why I wanted this section built in to TodoList.vue. This is not good. This increases tech debt instead of reducing it.

src/components/Dashboard/ActionRequired.vue
```diff
@@ -45,3 +45,12 @@ export default class ActionRequired extends Vue {
...
 }
 </style>
+
+<style lang="scss" scoped>
+
+.details-btn {
+  margin-bottom: 0.25rem;
+  height: inherit !important;
+  border: none;
+}
```
severinbeauvais, line 55: Why the special case for this button?

severinbeauvais: If you need to "pull this button up" so it doesn't affect the linespacing, the pull the button up with negative top margin.

Or maybe make this button have less padding overall (but it will be different from the other buttons).

Special case are not good!

severinbeauvais (changes requested):
I want to see less tech debt, not more.

src/components/Dashboard/ActionRequired.vue
shabeeb-aot: in todo list spacing is wrong i think

Screen Shot 2022-05-30 at 2 40 59 PM
I made it like this, to remove spacing between icon and word
Screen Shot 2022-05-30 at 2 41 33 PM

I can remove this change if you suggest

src/components/Dashboard/ProprietorPartnersListSm.vue
severinbeauvais: The alignmnent is off because the component expects to display an icon for the person's name. There is already an if-then-else in the code to display the icon+name or "not entered". I think it just needs to be adjusted.

src/components/Dashboard/ActionRequired.vue
shabeeb-aot: In todolist also we have the button with an outline (instead of text) and border none. 
As mentioned i removed height and did with negative top margin

src/views/Dashboard.vue
shabeeb-aot: updated

shabeeb-aot:
@severinbeauvais can you please review again

src/assets/styles/overrides.scss
```diff
@@ -124,3 +124,12 @@
...
 }
+
+.theme--light.v-btn.v-btn--disabled.v-btn--text {
+  opacity: 0.4 !important;
+  color: $app-blue !important;
+
+  .v-icon {
+    color: $app-blue !important;
+  }
+}
```
severinbeauvais, line 135: Nice! :)

src/components/Dashboard/ActionRequired.vue
severinbeauvais: This is OK for now.

I'm hoping we can simplify this (or make it common with other sub-components) when the Todo List is refactored/cleaned up.

src/views/Dashboard.vue
```diff
@@ -276,7 +276,8 @@ export default {
 }
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss" >
+@import "@/assets/styles/theme.scss";
```
severinbeauvais, line 280: Why did you unscope this? This component should be able to see all the assets already.

Also you don't need the theme import since you don't use any of the variables in it.

Also you don't need the blank line 299.

src/components/Dashboard/ActionRequired.vue
severinbeauvais: OK, thanks.

severinbeauvais (approved):
LGTM

severinbeauvais:
@shabeeb-aot I recommend that you update the app version, then you can merge this.

PR #340, opened 2022-06-03
UI Confirm Dissolution Modal Fixes/11663

meganawong:
*Issue #:* /bcgov/entity#11663

*Description of changes:*
* fix confirm dissolution modal for SP/GP 

### DIFF
--- a/src/components/dialogs/ConfirmDissolution.vue
+++ b/src/components/dialogs/ConfirmDissolution.vue
@@ -102,7 +102,7 @@ export default class ConfirmDissolution extends Vue {
 
   /** The incorporation text to display. */
   private get incorporationText (): string {
-    return this.getDissolutionConfirmationResource?.incorporationText
+    return this.getDissolutionConfirmationResource.incorporationText
   }
 
   // Pass click event to parent.
--- a/src/components/dialogs/ConfirmDissolution.vue
+++ b/src/components/dialogs/ConfirmDissolution.vue
@@ -72,7 +72,7 @@ export default class ConfirmDissolution extends Vue {
 
   /** The entity title to display. */
   get subEntityTitle (): string {
-    return this.getDissolutionConfirmationResource.subTitle
+    return this.getDissolutionConfirmationResource?.subTitle
   }
 
   /** The entity title to display. */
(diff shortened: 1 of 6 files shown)


src/resources/business-config.ts
```diff
@@ -80,7 +80,8 @@ export const ConfigJson = [
       act: 'Business Corporations',
       modalTitle: 'Voluntary Dissolution',
       confirmButtonText: 'Continue with Voluntary Dissolution',
-      dissolutionType: 'voluntarily dissolve'
+      dissolutionType: 'voluntarily dissolve',
+      incorporationText: 'an incorporated'
```
severinbeauvais, line 84: I _think_ `incorporationText` and `subTitle` are only used in one place, and always together. Why not combine them?

src/components/dialogs/ConfirmDissolution.vue
```diff
@@ -100,6 +100,11 @@ export default class ConfirmDissolution extends Vue {
     return this.getDissolutionConfirmationResource?.dissolutionType
   }
 
+  /** The incorporation text to display. */
+  private get incorporationText (): string {
```
severinbeauvais, line 104: Small cleanup: please change the local getters in this file to public (just remove private).

Also please change the prop to readonly.

severinbeauvais (commented):
Unit tests should have broken this change... but I see there are no unit tests for the various warning text scenarios.

Please add one (you only need to pass in props and verify that the warning text uses them correctly).

src/resources/business-config.ts
meganawong: combined

src/components/dialogs/ConfirmDissolution.vue
meganawong: done

src/components/dialogs/ConfirmDissolution.vue
```diff
@@ -58,45 +58,45 @@ import { Getter } from 'vuex-class'
...
-  private get entityTitle (): string {
+  get entityTitle (): string {
     return this.getDissolutionConfirmationResource?.entityTitle
   }
 
   /** The entity title to display. */
-  private get subEntityTitle (): string {
-    return this.getDissolutionConfirmationResource?.subTitle
+  get subEntityTitle (): string {
+    return this.getDissolutionConfirmationResource.subTitle
```
severinbeauvais, line 75: You removed the `?`, so if getDissolutionConfirmationResource is undefined/null then this getter will fail.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

severinbeauvais (approved):
OK to merge after optional chaining operator is restored.

PR #341, opened 2022-06-03
Create the filing for SP/GP/12092

meganawong:

*Description of changes:*
* add 'Put Back On' option for historical companies and 'Administrative Dissolution' option for active companies to 'Add Staff Filing' drop-down
* implement modal for 'Put Back On' and 'Administrative Dissolution' 

**PR for Review - NOT ready for merge**

### DIFF
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -93,8 +93,7 @@
                   <span class="app-blue">Put Back On</span>
                 </v-list-item-title>
               </v-list-item>
-              <v-list-item v-if="!isHistorical"
-              @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
+              <v-list-item v-if="!isHistorical" @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
                 <v-list-item-title>
                   <span class="app-blue">Administrative Dissolution</span>
                 </v-list-item-title>
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -87,18 +87,18 @@
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
-            <template v-if='isFirm'>
-            <v-list-item v-if="isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
-              <v-list-item-title>
-                <span class="app-blue">Put Back On</span>
-              </v-list-item-title>
-            </v-list-item>
-            <v-list-item v-if="!isHistorical"
-            @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
-              <v-list-item-title>
-                <span class="app-blue">Administrative Dissolution</span>
-              </v-list-item-title>
-            </v-list-item>
+            <template v-if="isFirm">
+              <v-list-item v-if="isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
+                <v-list-item-title>
+                  <span class="app-blue">Put Back On</span>
+                </v-list-item-title>
+              </v-list-item>
+              <v-list-item v-if="!isHistorical"
+              @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
+                <v-list-item-title>
+                  <span class="app-blue">Administrative Dissolution</span>
+                </v-list-item-title>
+              </v-list-item>
             </template>
           </v-list-item-group>
         </v-list>
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -87,17 +87,19 @@
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
-            <v-list-item v-if="isFirm && isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
+            <template v-if='isFirm'>
+            <v-list-item v-if="isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
               <v-list-item-title>
                 <span class="app-blue">Put Back On</span>
               </v-list-item-title>
             </v-list-item>
-            <v-list-item v-if="isFirm && !isHistorical"
+            <v-list-item v-if="!isHistorical"
             @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
               <v-list-item-title>
                 <span class="app-blue">Administrative Dissolution</span>
               </v-list-item-title>
             </v-list-item>
+            </template>
           </v-list-item-group>
         </v-list>
       </v-menu>
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -87,7 +87,7 @@
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
-            <v-list-item v-if="isFirm && isHistorical" @click="showPutBackOnDialog()" :disabled="disabled">
+            <v-list-item v-if="isFirm && isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
               <v-list-item-title>
                 <span class="app-blue">Put Back On</span>
               </v-list-item-title>
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -25,6 +25,22 @@
       courtOrderNumberRequired="true"
     />
 
+    <AddStaffNotationDialog
+      :dialog="isAddingPutBackOn"
+      @close="hidePutBackOnDialog($event)"
+      attach="#staff-notation"
+      displayName="Put Back On"
+      name="putBackOn"
+    />
+
+    <AddStaffNotationDialog
+      :dialog="isAddingAdministrativeDissolution"
+      @close="hideAdministrativeDissolutionDialog($event)"
+      attach="#staff-notation"
+      displayName="Adminstrative Dissolution"
+      name="administrativeDissolution"
+    />
+
     <div class="filing-item__actions">
       <v-menu offset-y left transition="slide-y-transition" v-model="expand">
         <template v-slot:activator="{ on }">
@@ -71,6 +87,17 @@
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
+            <v-list-item v-if="isFirm && isHistorical" @click="showPutBackOnDialog()" :disabled="disabled">
+              <v-list-item-title>
+                <span class="app-blue">Put Back On</span>
+              </v-list-item-title>
+            </v-list-item>
+            <v-list-item v-if="isFirm && !isHistorical"
+            @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
+              <v-list-item-title>
+                <span class="app-blue">Administrative Dissolution</span>
+              </v-list-item-title>
+            </v-list-item>
           </v-list-item-group>
         </v-list>
       </v-menu>
@@ -92,6 +119,8 @@ export default class StaffNotation extends Vue {
   private isAddingRegistrarsNotation = false
   private isAddingRegistrarsOrder = false
   private isAddingCourtOrder = false
+  private isAddingPutBackOn = false
+  private isAddingAdministrativeDissolution = false
   private expand = false
 
   /** Prop for the scrollbar offset to be added. */
@@ -102,6 +131,7 @@ export default class StaffNotation extends Vue {
 
   @Getter isFirm!: boolean
   @Getter getIdentifier: string
+  @Getter isHistorical!: boolean
 
   /** The Edit URL string. */
   get editUrl (): string {
@@ -132,6 +162,24 @@ export default class StaffNotation extends Vue {
 
   hideCourtOrderDialog (needReload: boolean): void {
     this.isAddingCourtOrder = false
+    this.close(needReload)
+  }
+
+  showPutBackOnDialog (): void {
+    this.isAddingPutBackOn = true
+  }
+
+  hidePutBackOnDialog (needReload: boolean): void {
+    this.isAddingPutBackOn = false
+    this.close(needReload)
+  }
+
+  showAdministrativeDissolutionDialog (): void {
+    this.isAddingAdministrativeDissolution = true
+  }
+
+  hideAdministrativeDissolutionDialog (needReload: boolean): void {
+    this.isAddingAdministrativeDissolution = false
     this.close(needReload)
   }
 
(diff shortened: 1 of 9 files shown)


src/components/Dashboard/StaffNotation.vue
```diff
@@ -71,6 +87,17 @@
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
+            <v-list-item v-if="isFirm && isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
```
severinbeauvais, line 90: Should this also be disabled when prop `disabled` is True?

severinbeauvais: Ok I get it now.

I recommend using a v-template for `isFirm`, and in that block put these 2 v-ifs (isHistorial and !isHistorical). I think that might more clearly bundle this functionality and make it easier to understand. What do you think?

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -1,17 +1,27 @@
 <template>
   <v-dialog v-model="dialog" width="45rem" persistent :attach="attach" content-class="add-notation-dialog">
     <v-card>
-      <v-card-title id="dialog-title">Add a {{displayName}}</v-card-title>
+      <v-card-title id="dialog-title">
+        <span v-if="administrativeDissolution" id="dialog-title"><strong>{{displayName}}</strong></span>
+        <span v-else-if="putBackOn" id="dialog-title"><strong>Correction - {{displayName}}</strong></span>
+        <span v-else id="dialog-title"><strong>Add a {{displayName}}</strong> </span>
+      </v-card-title>
```
severinbeauvais, line 8: Very nice and clean and clear 👍

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -1,17 +1,27 @@
...
       <v-card-text>
+        <div id="dialog-text" class='dialog-text'>
+          <p v-if="administrativeDissolution || putBackOn"> You are about to put
+          <strong><span class="entity-name">{{getEntityName}}</span>, {{getIdentifier}}</strong>
+          back on the register.</p>
+        </div>
         <div id="notation-text" class="mb-4 mt-2">
-          Enter a {{displayName}} that will appear on the ledger for this entity
+          Enter a {{administrativeDissolution || putBackOn? 'Detail' : displayName}}
```
severinbeauvais, line 16: For clarity, please put the compound conditional in parentheses, and add a space before the ? .

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -117,6 +129,14 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin) {
   /** Flag to enable validation in this component */
   private enableValidation = false
 
+  private get administrativeDissolution (): boolean {
+    return this.name === FilingTypes.ADMINISTRATIVE_DISSOLUTION
+  }
+
+  private get putBackOn (): boolean {
+    return this.name === FilingTypes.PUT_BACK_ON
```
severinbeauvais, line 137: Please add both of these to state-getters.ts and import them here instead.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -235,4 +255,10 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin) {
   color: $gray7;
   font-size: $px-16;
 }
+.dialog-text {
+  padding-bottom: 10px;
+}
+.entity-name {
+  text-transform: uppercase;
```
severinbeauvais, line 262: Please use built-in Vuetify class in the template instead of this.

https://vuetifyjs.com/en/styles/text-and-typography/

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -235,4 +255,10 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin) {
   color: $gray7;
   font-size: $px-16;
 }
+.dialog-text {
+  padding-bottom: 10px;
```
severinbeauvais, line 259: Most of our HTML uses top whitespace (on the next component) instead of bottom whitespace.

It's better to stick with all-top or all-bottom whitespace for layout purposes to handle conditional blocks (and also, as a de facto style, easier to update in the future).

severinbeauvais (commented):
Please remember to update the app version.

Please add some unit tests. In particular:
- you have new display scenarios in AddStaffNotationDialog depending on props and state
- you have added some filing types (see enum-mixin.spec.ts)

src/enums/filingTypes.ts
```diff
@@ -18,6 +18,8 @@ export enum FilingTypes {
   SPECIAL_RESOLUTION = 'specialResolution',
   TRANSITION = 'transition',
   VOLUNTARY_DISSOLUTION = 'voluntaryDissolution',
+  ADMINISTRATIVE_DISSOLUTION = 'administrativeDissolution',
+  PUT_BACK_ON = 'putBackOn',
```
severinbeauvais, line 22: Are there any mixins methods to update? `isTypeXXX' and `filingTypeToName`

src/components/Dashboard/StaffNotation.vue
meganawong: yes that makes sense, the change has been made

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: fixed

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: 'name' is passed as a prop from the parent so I don't think the functions can be moved to state-getters.ts and imported

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: done

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: switched to 'padding-top' for component below (notation-text)

src/enums/filingTypes.ts
meganawong: added mixins methods for new Filing Types

src/components/Dashboard/StaffNotation.vue
```diff
@@ -71,6 +87,19 @@
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
+            <template v-if='isFirm'>
```
severinbeauvais, line 90: Please follow existing code style.
1. use " for v-if statements (in template)
2. indent nested blocks

severinbeauvais:
> **PR for Review - NOT ready for merge**

In this case, click the text button in the Reviewers panel (top right) to convert this PR to a draft.

When this PR is ready for review/merge, then click the button down here somewhere to mark it "ready for review".

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -226,6 +247,9 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin) {
     margin-top: 0.5rem;
     padding-top: 0;
   }
+  #notation-text {
+    padding-top: 15px;
```
severinbeauvais, line 251: Do you need the padding as well as the margin? (The margin uses class helper "mt-2" in the template.) You may be able to use just one of these.

PS We try to use the class helpers as much as possible. In this case the closest one would be "pt-4" (16px). Is that close enough or does UI/UX want exactly 15px?

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -1,17 +1,28 @@
...
+          Enter a {{(administrativeDissolution || putBackOn) ? 'Detail' : displayName}}
+          that will appear on the ledger for this entity
         </div>
         <v-form ref="notationFormRef" v-model="notationFormValid" id="notation-form">
           <v-textarea
             v-model="notation"
             class="text-input-field mb-2"
             filled
-            :label="displayName"
+            :label="administrativeDissolution || putBackOn? 'Add Detail' : displayName"
```
severinbeauvais, line 25: You have this same logic on line 17. Would it be possible to remove from both places and add a local getter instead?

src/enums/filingNames.ts
```diff
@@ -19,4 +19,6 @@ export enum FilingNames {
   SPECIAL_RESOLUTION = 'Special Resolution',
   TRANSITION_APPLICATION = 'Transition Application',
   VOLUNTARY_DISSOLUTION = 'Voluntary Dissolution',
+  ADMINISTRATIVE_DISSOLUTION = 'Administrative Dissolution',
+  PUT_BACK_ON = 'Put Back On'
```
severinbeauvais, line 23: Please sort by key.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -117,6 +130,14 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin) {
   /** Flag to enable validation in this component */
   private enableValidation = false
 
+  private get administrativeDissolution (): boolean {
```
severinbeauvais, line 133: Since this component already imports mixins, you can import the Enum Mixin and use the new filing type helpers you added, instead of duplicating them here 😃 

(The helpers expect an object with property 'name', so you could call it like this: `isTypeAdministrativeDissolution({ name: this.name })`.)

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -2,19 +2,20 @@
   <v-dialog v-model="dialog" width="45rem" persistent :attach="attach" content-class="add-notation-dialog">
     <v-card>
       <v-card-title id="dialog-title">
-        <span v-if="administrativeDissolution" id="dialog-title"><strong>{{displayName}}</strong></span>
-        <span v-else-if="putBackOn" id="dialog-title"><strong>Correction - {{displayName}}</strong></span>
+        <span v-if="isTypeAdministrativeDissolution({ name: this.name })"
```
severinbeauvais, line 5: Oh no! My fault. This became more complicated than I thought.

Please continue to have the local getter (so that this template code is clean), and in the getter just call the global state getter.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Nice fix!

src/enums/filingNames.ts
severinbeauvais: Thanks.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: I was thinking it would be a getter for the string itself. `notationLabel` maybe?

(Lines 17 and 25 aren't consistent in what is displayed in the else case, so I leave it to you to work that out.)

In case you're stuck and there is higher-priority work to do, I think this isn't serious to block this PR. If you decide not to do this now then just add a comment to that effect here.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Or... using a JS object property shortcut, you could use this in the template instead:

`v-if="istypeAdministrativeDissolution({ name })"`

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Looks good.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Looks good.

meganawong:
When I tried using the global state getter in the local getter as (this.isTypeAdministrativeDissolution({ name: this.name }) and (this.isTypePutBackOn({ name: this.name }), both conditions always appeared to evaluate to true and would render both cases? If I pass **name** in as a parameter to both local getters, it behaves as expected but I am not sure if this is a desired implementation. Please let me know your thoughts

severinbeauvais:
> both conditions always appeared to evaluate to true and would render both cases? If I pass **name** in as a parameter to both local getters, it behaves as expected but I am not sure if this is a desired implementation. Please let me know your thoughts

The global getters are like variables, not like functions. Try without `()`.

severinbeauvais:
> > both conditions always appeared to evaluate to true and would render both cases? If I pass **name** in as a parameter to both local getters, it behaves as expected but I am not sure if this is a desired implementation. Please let me know your thoughts
> 
> The global getters are like variables, not like functions. Try without `()`.

This isn't going to work anyway, since the global getters look at the CURRENT filing name, and here you want to determine if it's a given name. Looks like your original local getter solution was the correct one. (Sorry)

severinbeauvais:
> ~This isn't going to work anyway, since the global getters look at the CURRENT filing name, and here you want to determine if it's a given name. Looks like your original local getter solution was the correct one. (Sorry)~

I am quickly making mistakes and slowly remembering where you're at.

Those global calls are indeed FUNCTIONS. They come from a mixin. Ignore my comments above 🙄 

Now I will look at your code again and try to see why it doesn't work.

severinbeauvais:
Try changing administrativeDissolution() to a local getter. Then you can also remove the parameters in the template.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -1,17 +1,29 @@
 <template>
   <v-dialog v-model="dialog" width="45rem" persistent :attach="attach" content-class="add-notation-dialog">
     <v-card>
-      <v-card-title id="dialog-title">Add a {{displayName}}</v-card-title>
+      <v-card-title id="dialog-title">
+        <span v-if="administrativeDissolution(name)"
+        id="dialog-title"><strong>{{displayName}}</strong></span>
```
severinbeauvais, line 6: Move this to previous line.

meganawong: done.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -71,6 +87,19 @@
...
               </v-list-item-title>
             </v-list-item>
+            <template v-if="isFirm">
+              <v-list-item v-if="isHistorical" @click="showPutBackOnDialog()" :disabled="!isHistorical">
+                <v-list-item-title>
+                  <span class="app-blue">Put Back On</span>
+                </v-list-item-title>
+              </v-list-item>
+              <v-list-item v-if="!isHistorical"
+              @click="showAdministrativeDissolutionDialog()" :disabled="disabled">
```
severinbeauvais, line 97: Move this to previous line (or if necessary, indent it as per existing code style).

shabeeb-aot (approved):

PR #342, opened 2022-06-06
11374 Filings UI: get account id from CURRENT_ACCOUNT

lewischenstudio:
*Issue #:* [/bcgov/entity#11374](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11374)

*Description of changes:*
- removed `accountId` in file: fetch-config.ts 
- updated `accountId` in file: navigate.ts
- updated app version in file: package.json
- updated many unit test files to accommodate `CURRENT_ACCOUNT`

### DIFF
--- a/tests/unit/fetch-config.spec.ts
+++ b/tests/unit/fetch-config.spec.ts
@@ -43,7 +43,6 @@ xdescribe('Fetch Config', () => {
     await fetchConfig()
 
     // verify data
-    expect(sessionStorage.getItem('ACCOUNT_ID')).toBe('2288')
     expect(sessionStorage.getItem('AUTH_WEB_URL')).toBe('auth web url')
     expect(sessionStorage.getItem('BUSINESSES_URL')).toBe('businesses url')
     expect(sessionStorage.getItem('BUSINESS_CREATE_URL')).toBe('business create url')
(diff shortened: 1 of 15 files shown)


tests/unit/fetch-config.spec.ts
```diff
@@ -43,7 +43,6 @@ xdescribe('Fetch Config', () => {
     await fetchConfig()
 
     // verify data
-    expect(sessionStorage.getItem('ACCOUNT_ID')).toBe('2288')
```
severinbeauvais, line 46: Somewhere, we should have a test to verify that the account id is correctly parsed and used, ie:
- in BreadcrumbResources.ts
- in navigate.ts

Maybe app.spec.ts would be a good place? I'd be OK with a new describe section in there just to keep it separate from everything else.

severinbeauvais (commented):
Looks really good.

Thanks for the other, small cleanup!

I'm hoping for another unit test but otherwise I see no needed code changes.

lambert-alex (approved):

tests/unit/fetch-config.spec.ts
severinbeauvais: I'm hoping for some tests that verify that both places this value is used, it's properly parsed from the current account object and they build the correct output string (for breadcrumb or navigate).

severinbeauvais: Ah, OK, I see now. Looks good.

vysakh-menon-aot (approved):

severinbeauvais (approved):

PR #346, opened 2022-06-10
Create filing sp gp/12092

meganawong:

*Description of changes:*
* update button text on modal for file/save
* addition of local getter to check if filing type is Administrative Dissolution/Put Back
* update data sent to create Administrative Dissolution/Put Back On filing

### DIFF
--- a/src/components/dialogs/AddStaffNotationDialog.vue
+++ b/src/components/dialogs/AddStaffNotationDialog.vue
@@ -14,7 +14,7 @@
           {{getIdentifier}}</strong> back on the register.</p>
         </div>
         <div id="notation-text" class="mb-4 mt-2 pt-4">
-          Enter a {{(administrativeDissolution || putBackOn) ? 'Detail' : displayName}}
+          Enter a {{isAdministrativeDissoltionOrPutBackOn ? 'Detail' : displayName}}
           that will appear on the ledger for this entity
         </div>
         <v-form ref="notationFormRef" v-model="notationFormValid" id="notation-form">
@@ -22,7 +22,7 @@
             v-model="notation"
             class="text-input-field mb-2"
             filled
-            :label="(administrativeDissolution || putBackOn) ? 'Add Detail' : displayName"
+            :label="isAdministrativeDissoltionOrPutBackOn ? 'Add Detail' : displayName"
             id="notation"
             rows="5"
             :no-resize="true"
@@ -49,7 +49,13 @@
       <v-card-actions class="pt-0">
         <v-spacer></v-spacer>
         <div class="form__btns">
-          <v-btn text color="primary"
+          <v-btn v-if="isAdministrativeDissoltionOrPutBackOn" text color="primary"
+            id="dialog-save-button"
+            :loading="saving"
+            @click.native="save()"
+            class="save-btn"
+          >File</v-btn>
+          <v-btn v-else text color="primary"
             id="dialog-save-button"
             :loading="saving"
             @click.native="save()"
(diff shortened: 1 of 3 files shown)


src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -49,7 +49,13 @@
       <v-card-actions class="pt-0">
         <v-spacer></v-spacer>
         <div class="form__btns">
-          <v-btn text color="primary"
+          <v-btn v-if="isAdministrativeDissolutionOrPutBackOn" text color="primary"
+            id="dialog-save-button"
+            :loading="saving"
+            @click.native="save()"
+            class="save-btn"
+          >File</v-btn>
```
severinbeauvais, line 57: What is this doing? It looks like this v-btn is the same as the one below it.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -140,6 +148,11 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin, EnumMixin)
     return this.isTypePutBackOn({ name: this.name })
   }
 
+  /** Return true if this filing is an administrative dissolution or put back on */
+  private get isAdministrativeDissolutionOrPutBackOn (): boolean {
+    return (this.isTypeAdministrativeDissolution({ name: this.name }) || this.isTypePutBackOn({ name: this.name }))
```
severinbeauvais, line 153: You already have 2 getters for these conditions (and getter values are cached), so it would be better to just use your getters here 😀

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -194,8 +207,27 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin, EnumMixin)
       this.saving = false
       return
     }
-
-    const data = {
+    const data = this.isAdministrativeDissolutionOrPutBackOn ? {
```
severinbeauvais, line 210: Could you break this up into separate consts please?
You will have to switch between the objects in the code that uses them.

severinbeauvais (commented):
Looks good so far.

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: will do, are there any specific names i should be using for the consts

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: reverted!

src/components/dialogs/AddStaffNotationDialog.vue
meganawong: fixed

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -194,7 +196,28 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin, EnumMixin)
       this.saving = false
       return
     }
-
+    const dataSpGp = {
```
shabeeb-aot, line 199: instead of making two-variable data and dataSpGp, can combine and push to object 
as discussed you can look here for example
https://github.com/bcgov/business-create-ui/blob/main/src/mixins/filing-template-mixin.ts#L495

meganawong: done @severinbeauvais please review again when you have the chance

severinbeauvais: This is Faiz's conversation to resolve.

shabeeb-aot (approved):

PR #350, opened 2022-06-22
Admin dissolution/12092

meganawong:

*Description of changes:*12092
* add frontend support for admin dissolution
* update error message to identify file type when request cannot be fulfilled 

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -46,9 +46,8 @@
 
                 <!-- is this a Staff Only filing? -->
                 <div v-if="filing.isTypeStaff" class="item-header__subtitle">
-                  <span v-if="filing.name === 'putBackOn' ||
-                  filing.name === 'administrativeDissolution'">FILED<FiledLabel :filing="filing" /></span>
-                  <span v-else><FiledLabel :filing="filing" /></span>
+                  <span v-if="filing.putBackOnOrAdminDissolution">FILED<FiledLabel :filing="filing" /></span>
+                  <span v-else><FiledLabel :filing="filing"/></span>
                 </div>
 
                 <!-- is this a FE BCOMP COA pending (not yet completed)? -->
@@ -592,6 +591,7 @@ export default class FilingHistoryList extends Mixins(
         item.isTypeStaff = true
         item.notationOrOrder = filing.data.order?.orderDetails // should not be falsy
         item.planOfArrangement = filing.data.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
+        item.putBackOnOrAdminDissolution = filing.name === 'putBackOn' || filing.displayName === 'Administrative Dissolution'
       }
 
       this.historyItems.push(item)
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -591,7 +591,8 @@ export default class FilingHistoryList extends Mixins(
         item.isTypeStaff = true
         item.notationOrOrder = filing.data.order?.orderDetails // should not be falsy
         item.planOfArrangement = filing.data.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
-        item.putBackOnOrAdminDissolution = filing.name === 'putBackOn' || filing.displayName === 'Administrative Dissolution'
+        item.putBackOnOrAdminDissolution = filing.name === 'putBackOn' ||
+        filing.displayName === 'Administrative Dissolution'
       }
 
       this.historyItems.push(item)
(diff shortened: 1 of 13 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -592,6 +591,8 @@ export default class FilingHistoryList extends Mixins(
         item.isTypeStaff = true
         item.notationOrOrder = filing.data.order?.orderDetails // should not be falsy
         item.planOfArrangement = filing.data.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
+        item.putBackOnOrAdminDissolution = filing.name === 'putBackOn' ||
+        filing.displayName === 'Administrative Dissolution'
```
severinbeauvais, line 595: String codes should not be used. At a minimum, you should use enums here.

I recommend you use the mixin methods you created a few weeks back (`isTypePutBackOn()`, etc).

src/enums/dissolutionTypesNames.ts
```diff
@@ -4,6 +4,7 @@ export enum DissolutionTypes {
   // COURT_ORDER_LIQUIDATION = 'courtOrderedLiquidation',
   // INVOLUNTARY = 'involuntary',
   VOLUNTARY = 'voluntary',
+  ADMIN = 'administrative',
```
severinbeauvais, line 7: Please use line 3 instead.

src/enums/dissolutionTypesNames.ts
```diff
@@ -29,6 +30,7 @@ export enum DissolutionNames {
   // INVOLUNTARY = 'Involuntary Dissolution',
   VOLUNTARY = 'Voluntary Dissolution',
   FIRM_DISSOLUTION = 'Dissolution',
+  ADMIN = 'Administrative Dissolution'
```
severinbeauvais, line 33: Please use line 28 instead.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -38,7 +38,8 @@
       @close="hideAdministrativeDissolutionDialog($event)"
       attach="#staff-notation"
       displayName="Adminstrative Dissolution"
-      name="administrativeDissolution"
+      name="dissolution"
+      dissolutionType="administrative"
     />
```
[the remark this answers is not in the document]
severinbeauvais: ^^ I see you found the existing code 👍 

I think you just need the commented-out enum values instead of the duplicate ones your created.

tests/unit/FilingHistoryList2.spec.ts
```diff
@@ -60,7 +62,8 @@ filings.forEach((filing: any, index: number) => {
 
       // mount the component
       const $route = { query: {} }
-      wrapper = mount(FilingHistoryList, { store, mocks: { $route }, vuetify })
+      wrapper = mount(FilingHistoryList, {
+        store, mocks: { $route }, propsData: { dissolutionType: 'administrative' }, vuetify })
```
severinbeauvais, line 66: This will make prop dissolutionType = 'administrative' for all tests in this for loop. You don't want that.

Instead, you can use `wrapper.setProps()` in your specific test blocks below.

Note that props exist for the life of the wrapper -- so other tests may see a changed prop -- so you may need to clear the prop after your test block.

src/components/Dashboard/StaffNotation.vue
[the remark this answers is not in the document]
severinbeauvais: Thanks! Great cleanup!

tests/unit/filings.json
```diff
@@ -519,6 +521,33 @@
       "status": "COMPLETED",
       "submittedDate": "Sun, 01 Oct 2017 00:02:26 GMT",
       "submitter": "Registry Staff"
+    },
+    {
+      "availableOnPaperOnly": false,
+      "businessIdentifier": "CP0000840",
```
severinbeauvais, line 527: I don't know if it makes any difference in the code, but is this a valid entity type for administrative dissolution?

Actually this is the business ID; the entity type is set elsewhere. Maybe this test suite needs to be laid out different for CP vs BC vs FM... can be future work (as needed) if the current tests aren't broken by this.

tests/unit/FilingHistoryList2.spec.ts
severinbeauvais: Good fix!

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -592,6 +591,9 @@ export default class FilingHistoryList extends Mixins(
         item.isTypeStaff = true
         item.notationOrOrder = filing.data.order?.orderDetails // should not be falsy
         item.planOfArrangement = filing.data.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
+        item.putBackOnOrAdminDissolution = this.isTypePutBackOn({ name: filing.name }) ||
+        this.isTypeAdministrativeDissolution({ name: filing.name,
```
severinbeauvais, line 595: I'm not sure why the linter didn't catch this, but it should be indented (since it's a continuation of the previous line).

PR #352, opened 2022-06-24
12199 Update Doc List: Suppress output for Coop Annual Report Correction for all Coop records

lewischenstudio:
*Issue #:* [/bcgov/entity#12199](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12199)

*Description of changes:*
- suppress `Notice of Articles` item for "Correction - Annual Report" in Recent Filing History, if 
  - the filing type is `Coop`,
  - the display name is `Correction - Annual Report`, and
  - the document link is for `noticeOfArticles`

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -885,9 +885,9 @@ export default class FilingHistoryList extends Mixins(
   /** Check whether to supprese notice of articles. */
   private suppressNOA (displayName:string, prop: string): boolean {
     // Only suppress NOA when it meets the following conditions
-    let type = this.getEntityType === CorpTypeCd.COOP
-    let name = displayName === 'Correction - Annual Report'
-    let isNOA = prop === 'noticeOfArticles'
+    const type = this.getEntityType === CorpTypeCd.COOP
+    const name = displayName === 'Correction - Annual Report'
+    const isNOA = prop === 'noticeOfArticles'
     return type && name && isNOA
   }
 
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -351,7 +351,7 @@ import { DetailsList } from '@/components/common'
 import { AddCommentDialog, DownloadErrorDialog, LoadCorrectionDialog } from '@/components/dialogs'
 
 // Enums, interfaces and mixins
-import { AllowableActions, FilingTypes, Routes } from '@/enums'
+import { AllowableActions, CorpTypeCd, FilingTypes, Routes } from '@/enums'
 import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryItemIF, LegalFilingIF }
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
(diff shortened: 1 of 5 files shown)


severinbeauvais (changes requested):
This is not the correct solution.

The list of outputs is provided by the Legal API. Hence, the Legal API should not include this output in this scenario.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -882,6 +882,15 @@ export default class FilingHistoryList extends Mixins(
     )
   }
 
+  /** Check whether to supprese notice of articles. */
+  private suppressNOA (displayName:string, prop: string): boolean {
+    // Only suppress NOA when it meets the following conditions
+    const type = this.getEntityType === CorpTypeCd.COOP
+    const name = displayName === 'Correction - Annual Report'
+    const isNOA = prop === 'noticeOfArticles'
```
severinbeauvais, line 890: If this was the right solution then you should use existing getters and enums as much as possible.

lewischenstudio: Yeah. I was looking for enums for the `displayName` and the document name, but we don't have them. They are however available in the backend.

vysakh-menon-aot:
> This is not the correct solution.
> 
> The list of outputs is provided by the Legal API. Hence, the Legal API should not include this output in this scenario.

@lewischenstudio look into this code. https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/core/filing.py#L454
i think you just need to add an or condition to it.

lewischenstudio:
> > This is not the correct solution.
> > The list of outputs is provided by the Legal API. Hence, the Legal API should not include this output in this scenario.
> 
> @lewischenstudio look into this code. https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/core/filing.py#L454 i think you just need to add an or condition to it since the ticket say's `Suppress all outputs`.

I verified this ticket with Mihai that we only need to suppress "Notice of Articles" and keep the rest of the outputs.

lewischenstudio:
> This is not the correct solution.
> 
> The list of outputs is provided by the Legal API. Hence, the Legal API should not include this output in this scenario.

There are two places to work on this ticket, and I wasn't sure if I should use the FE or BE. I guess we can use the legal-api for this ticket then.

lewischenstudio:
> > This is not the correct solution.
> > The list of outputs is provided by the Legal API. Hence, the Legal API should not include this output in this scenario.
> 
> @lewischenstudio look into this code. https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/core/filing.py#L454 i think you just need to add an or condition to it since the ticket say's `Suppress all outputs`.

For only to suppress NOA, I think I need to modify under this if condition (https://github.com/bcgov/lear/blob/e259eb1fb42c77b8e43b020081ef25558ff5ceea/legal-api/src/legal_api/core/meta/filing.py#L334-339), since it's for correction type.

vysakh-menon-aot:
> > > This is not the correct solution.
> > > The list of outputs is provided by the Legal API. Hence, the Legal API should not include this output in this scenario.
> > 
> > 
> > @lewischenstudio look into this code. https://github.com/bcgov/lear/blob/main/legal-api/src/legal_api/core/filing.py#L454 i think you just need to add an or condition to it since the ticket say's `Suppress all outputs`.
> 
> For only to suppress NOA, I think I need to modify under this if condition [334-339](https://github.com/bcgov/lear/blob/e259eb1fb42c77b8e43b020081ef25558ff5ceea/legal-api/src/legal_api/core/meta/filing.py#L334-339), since it's for correction type.

Yes

PR #353, opened 2022-06-29
Dissolution ux fixes/12629

meganawong:
*Issue #:* /bcgov/entity#12629

*Description of changes:*
Implemented Dissolution UX Fixes in entity-info header and drop down button.

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -367,17 +367,13 @@
                     <v-list class="actions__more-actions">
                       <v-list-item
                         class="dissolution-drop-down"
-                        @mouseenter="hover = true"
-                        @mouseleave="hover = false"
-                        :class="{ 'dissolution-drop-down-hover': hover }"
                         v-if="businessId"
                         id="btn-delete-draft"
                         @click="confirmDeleteDraft(item)"
                       >
                         <v-icon class="pr-1" color="primary" size="18px">mdi-delete-forever</v-icon>
                         <template v-if="isTypeDissolution(item)">
-                          <v-list-item-title
-                          >Delete {{ toDoListTitle }}</v-list-item-title>
+                          <v-list-item-title>Delete {{ toDoListTitle }}</v-list-item-title>
                         </template>
                         <v-list-item-title v-else>Delete Draft</v-list-item-title>
                       </v-list-item>
@@ -1636,12 +1632,6 @@ export default class TodoList extends Mixins(
     // cancel the check timer if it is running
     clearTimeout(this.checkTimer)
   }
-
-  data () {
-    return {
-      hover: false
-    }
-  }
 }
 </script>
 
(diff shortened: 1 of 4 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -366,13 +366,18 @@
                     </template>
                     <v-list class="actions__more-actions">
                       <v-list-item
+                        class="dissolution-drop-down"
+                        @mouseenter="hover = true"
```
shabeeb-aot, line 370: this one can be controlled by css only
like 
dissolution-drop-down:hover{
css
}

src/components/Dashboard/TodoList.vue
```diff
@@ -1736,6 +1737,10 @@ export default class TodoList extends Mixins(
     font-size: $px-14;
     color: $app-blue;
   }
+
+  .dissolution-drop-down:hover {
+    background-color: #F1F3F5;
```
shabeeb-aot, line 1742: this is $gray1 in theme.scss , so can use variable

shabeeb-aot (approved):

src/components/Dashboard/TodoList.vue
```diff
@@ -1726,7 +1727,7 @@ export default class TodoList extends Mixins(
 }
 
 .actions__more-actions {
-  padding: 0;
+  padding: 1;
```
severinbeauvais, line 1730: What units is this?

I _think_ that by default these are pixels, and if this is what you mean then please add the "px".

If, instead, you mean the equivalent of "pa-1", then note that the Vuetify spacing classes are 4px per unit.

If possible, please update this next time you're in here. Thanks.

src/components/EntityInfo.vue
```diff
@@ -243,12 +242,7 @@ export default class EntityInfo extends Mixins(AllowableActionsMixin, CommonMixi
 
   /** The business description. */
   get businessDescription (): string {
-    const corpTypeDescription = this.getCorpTypeDescription(this.getEntityType)
-
-    if (this.isSoleProp) {
-      return `${corpTypeDescription} / Doing Business As (DBA)`
```
severinbeauvais, line 249: Are you certain about removing this? It was an explicit requirement from Yui. Please confirm.

severinbeauvais: Looking at the design requirements in the ticket (12629), "DBA" should be removed once a SP has been registered/created.

This change also removed "DBA" when the business is still unregistered, which is incorrect.

image

meganawong: Is there a specific field you know off the top of your head which I can check to see if a business is registered or not?

severinbeauvais: Yes. If there is a BUSINESS_ID in session storage then this is a business. Otherwise there should be a TEMP_REG_NUMBER in session storage, indicating that this is not a business yet. These are mutually exclusive.

https://github.com/bcgov/business-filings-ui/blob/493b73328f8377515832264f8740e0419ca551c8/src/App.vue#L221
https://github.com/bcgov/business-filings-ui/blob/493b73328f8377515832264f8740e0419ca551c8/src/App.vue#L231

meganawong: perfect, thanks!

severinbeauvais: This will be fixed in https://github.com/bcgov/business-filings-ui/pull/357.

PR #354, opened 2022-06-29
remove court order validation for AD and PutBackOn

meganawong:

*Description of changes:*
* remove frontend validation for court order number (not a required field)
* update data sent for admin dissolution in court order object


ritvick (approved):

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -236,10 +246,12 @@ export default class AddStaffNotationDialog extends Mixins(DateMixin, EnumMixin)
       data.filing[this.name] = { ...data.filing[this.name],
         dissolutionType: 'administrative',
         dissolutionDate: this.getCurrentDate,
-        courtOrder: {
+        details: this.notation,
+        ...(this.courtOrderNumber && { courtOrder: {
```
shabeeb-aot, line 250: it won't be a good idea to use a lot of conditions inside spread operator better to put in the next line with the condition .

PR #356, opened 2022-07-05
12392 Filings UI - launch the filing

lewischenstudio:
*Issue #:* [/bcgov/entity#12392](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12392)

*Description of changes:*
- [x] created `FileCorrectionDialog.vue` for correction dialog modal
- [x] updated the logic to enable correction for corrected filings
- [x] updated redirect url for separate cases
- [x] updated correction interfaces to fit the [new schema](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/correction.json)
- [x] unit tests

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -22,7 +22,7 @@
     <FileCorrectionDialog
       :dialog="fileCorrectionDialog"
       @exit="fileCorrectionDialog=false"
-      @start="startCorrection($event)"
+      @redirect="correctThisFiling($event)"
       attach="#filing-history-list"
     />
 
@@ -209,7 +209,7 @@
                         </v-list-item-icon>
                         <v-list-item-title
                           class="file-correction-item"
-                          @click.stop="correctThisFiling(filing)"
+                          @click.stop="startCorrectionDialog(filing)"
                         >
                           <span class="app-blue">File a Correction</span>
                         </v-list-item-title>
@@ -365,7 +365,7 @@ import { DetailsList } from '@/components/common'
 import { AddCommentDialog, DownloadErrorDialog, FileCorrectionDialog, LoadCorrectionDialog } from '@/components/dialogs'
 
 // Enums, interfaces and mixins
-import { AllowableActions, FilingTypes, FilingNames, Routes } from '@/enums'
+import { AllowableActions, FilingTypes, Routes, CorrectionTypes } from '@/enums'
 import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryItemIF, LegalFilingIF }
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
@@ -420,7 +420,6 @@ export default class FilingHistoryList extends Mixins(
 
   // enum for template
   readonly AllowableActions = AllowableActions
-  readonly FilingNames = FilingNames
 
   /** The Edit URL string. */
   private get editUrl (): string {
@@ -647,7 +646,13 @@ export default class FilingHistoryList extends Mixins(
     if (index >= 0) this.togglePanel(index, this.historyItems[index])
   }
 
-  private async correctThisFiling (item: HistoryItemIF): Promise<void> {
+  private startCorrectionDialog (item: HistoryItemIF): void {
+    this.fileCorrectionDialog = true
+    this.correctionFiling = item
+  }
+
+  private async correctThisFiling (correctionType: CorrectionTypes): Promise<void> {
+    const item = this.correctionFiling
     const correctedFilingId = item.filingId?.toString()
 
     switch (item?.name) {
@@ -687,15 +692,52 @@ export default class FilingHistoryList extends Mixins(
       case FilingTypes.INCORPORATION_APPLICATION:
       case FilingTypes.CHANGE_OF_REGISTRATION:
       case FilingTypes.REGISTRATION:
-        this.fileCorrectionDialog = true
-        this.correctionFiling = item
+        try {
+          // show spinner since the network calls below can take a few seconds
+          this.$root.$emit('showSpinner', true)
+
+          // fetch original filing
+          const origFiling = item.filingLink && await this.fetchFiling(item.filingLink)
+          if (!origFiling) {
+            throw new Error('Invalid API response')
+          }
+
+          // build correction filing
+          let correctionFiling: CorrectionFilingIF
+          if (this.isBComp) correctionFiling = this.buildIaCorrectionFiling(origFiling)
+          if (this.isFirm) correctionFiling = this.buildFmCorrectionFiling(origFiling)
+
+          if (!correctionFiling) throw new Error('Invalid filing type')
+
+          // save draft filing
+          const draftCorrection = await this.createFiling(this.getIdentifier, correctionFiling, true)
+          const draftCorrectionId = +draftCorrection?.header?.filingId
+          if (isNaN(draftCorrectionId)) {
+            throw new Error('Invalid API response')
+          }
+
+          // navigate to Edit web app to complete this correction
+          // NB: no need to clear spinner
+          const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction/?correction-id=${draftCorrectionId}`
+          navigate(correctionUrl)
+        } catch (error) {
+          // clear spinner on error
+          this.$root.$emit('showSpinner', false)
+
+          // eslint-disable-next-line no-console
+          console.log('Error creating correction =', error)
+          this.loadCorrectionDialog = true
+        }
         break
 
       case FilingTypes.CORRECTION:
         // FUTURE: allow a correction to a correction?
         // this.$router.push({ name: Routes.CORRECTION,
         //   params: { correctedFilingId } })
-        alert('At this time, you cannot correct a correction. Please contact Ops if needed.')
+        this.$router.push({
+          name: Routes.CORRECTION,
+          params: { correctedFilingId }
+        })
         break
 
       case FilingTypes.ALTERATION:
@@ -868,7 +910,7 @@ export default class FilingHistoryList extends Mixins(
       item.isFutureEffective ||
       this.isStatusCorrected(item) ||
       this.isTypeAlteration(item) ||
-      // this.isTypeCorrection(item) ||
+      this.isTypeCorrection(item) ||
       this.isTypeTransition(item) ||
       this.isTypeIncorporationApplication(item) || // *** temporary
       // the following corrections are supported:
@@ -876,45 +918,6 @@ export default class FilingHistoryList extends Mixins(
       (this.isTypeChangeOfRegistration(item) && !this.isFirm) ||
       (this.isTypeRegistration(item) && !this.isFirm)
     )
-  }
-
-  private async startCorrection (startCorrection: boolean): Promise<void> {
-    this.fileCorrectionDialog = false
-    try {
-      // show spinner since the network calls below can take a few seconds
-      this.$root.$emit('showSpinner', true)
-
-      // fetch original filing
-      const origFiling = this.correctionFiling.filingLink && await this.fetchFiling(this.correctionFiling.filingLink)
-      if (!origFiling) {
-        throw new Error('Invalid API response')
-      }
-
-      // build correction filing
-      let correctionFiling: CorrectionFilingIF
-      if (this.isBComp) correctionFiling = this.buildIaCorrectionFiling(origFiling)
-      if (this.isFirm) correctionFiling = this.buildFmCorrectionFiling(origFiling)
-      if (!correctionFiling) throw new Error('Invalid filing type')
-
-      // save draft filing
-      const draftCorrection = await this.createFiling(this.getIdentifier, correctionFiling, true)
-      const draftCorrectionId = +draftCorrection?.header?.filingId
-      if (isNaN(draftCorrectionId)) {
-        throw new Error('Invalid API response')
-      }
-
-      // navigate to Edit web app to complete this correction
-      // NB: no need to clear spinner
-      const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction/?correction-id=${draftCorrectionId}`
-      navigate(correctionUrl)
-    } catch (error) {
-      // clear spinner on error
-      this.$root.$emit('showSpinner', false)
-
-      // eslint-disable-next-line no-console
-      console.log('Error creating correction =', error)
-      this.loadCorrectionDialog = true
-    }
   }
 
   @Watch('getFilings', { immediate: true })
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -19,6 +19,13 @@
       attach="#filing-history-list"
     />
 
+    <FileCorrectionDialog
+      :dialog="fileCorrectionDialog"
+      @exit="fileCorrectionDialog=false"
+      @start="startCorrection($event)"
+      attach="#filing-history-list"
+    />
+
     <!-- Alternate Loading Spinner -->
     <v-fade-transition>
       <div class="loading-container grayed-out" v-show="isBusy">
@@ -355,10 +362,10 @@ import PaperFiling from './FilingHistoryList/PaperFiling.vue'
 import PendingFiling from './FilingHistoryList/PendingFiling.vue'
 import StaffFiling from './FilingHistoryList/StaffFiling.vue'
 import { DetailsList } from '@/components/common'
-import { AddCommentDialog, DownloadErrorDialog, LoadCorrectionDialog } from '@/components/dialogs'
+import { AddCommentDialog, DownloadErrorDialog, FileCorrectionDialog, LoadCorrectionDialog } from '@/components/dialogs'
 
 // Enums, interfaces and mixins
-import { AllowableActions, FilingTypes, Routes } from '@/enums'
+import { AllowableActions, FilingTypes, FilingNames, Routes } from '@/enums'
 import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryItemIF, LegalFilingIF }
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin, LegalApiMixin } from '@/mixins'
@@ -382,6 +389,7 @@ import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin, LegalApiMixin
     // dialogs
     AddCommentDialog,
     DownloadErrorDialog,
+    FileCorrectionDialog,
     LoadCorrectionDialog
   }
 })
@@ -399,6 +407,7 @@ export default class FilingHistoryList extends Mixins(
   // local properties
   private addCommentDialog = false
   private downloadErrorDialog = false
+  private fileCorrectionDialog = false
   private loadCorrectionDialog = false
   private panel: number = null // currently expanded panel
   private historyItems: Array<HistoryItemIF> = []
@@ -407,9 +416,11 @@ export default class FilingHistoryList extends Mixins(
   private currentFiling: HistoryItemIF = null
   private loadingOneIndex = -1
   private isBusy = false
+  private correctionFiling: HistoryItemIF = null
 
   // enum for template
   readonly AllowableActions = AllowableActions
+  readonly FilingNames = FilingNames
 
   /** The Edit URL string. */
   private get editUrl (): string {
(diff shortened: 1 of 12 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -1075,6 +1076,8 @@ export default class TodoList extends Mixins(
         correctedFilingId: correction.correctedFilingId,
         // this is only used for external corrections (IA):
         correctedFilingType: this.filingTypeToName(correction.correctedFilingType as FilingTypes),
+        // Future: verify correction.type for either CLIENT OR STAFF
+        type: isEmpty(correction.type) ? null : correction.type,
```
severinbeauvais, line 1080: `correction.type` is an enum, not an object, so you don't need isEmpty.

In fact, why do anything special here -- can't type = correction.type ?

src/components/Dashboard/TodoList.vue
```diff
@@ -1326,8 +1329,11 @@ export default class TodoList extends Mixins(
           (item.correctedFilingType === FilingNames.CHANGE_OF_REGISTRATION) ||
           (item.correctedFilingType === FilingNames.REGISTRATION)
         ) {
-          // navigate to Edit UI to resume this correction
-          const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction/?correction-id=${item.filingId}`
+          // navigate to Edit UI to resume this correction (depending on either CLIENT OR STAFF)
+          const userType = isEmpty(item?.type) ? '' : '?u=' + item?.type.toLowerCase()
```
severinbeauvais, line 1333: I think you don't need to pass the user type as a URL parameter -- it's already in the draft correction filing that the Edit UI will load.

src/components/Dashboard/TodoList.vue
```diff
@@ -51,7 +51,7 @@
                   color="red"
                   :ripple=false
                 >
-                  <v-icon left>mdi-information-outline</v-icon>
+                  <v-icon>mdi-information-outline</v-icon>
```
severinbeauvais, line 54: Please include a screenshot of this change.

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+export default class FileCorrectionDialog extends Vue {
+  // Initialize hasChosenCorrection.
+  private hasChosenCorrection = true
+
+  /** Radio group model property. */
+  private correctionOptions = {
+    NONE: 'NONE',
+    CLIENT: 'CLIENT',
+    STAFF: 'STAFF'
+  }
```
severinbeauvais, line 56: This should be an enum and should be used in other places in this project.

Also, I see the value here is emitted as a string, is received as a string, and that string is appended to a route name or something. Instead, to make good use of typings and to help prevent issues, please keep this as a enum all the way until the route name is created. This is also the concept of keeping things where they belong; specifically, THIS code should not care what the route name will be.

src/components/Dashboard/TodoList.vue
lewischenstudio: Sure. I'll remove the url for the Edit UI. I wasn't sure how Edit UI would load the correction.type.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -629,7 +648,7 @@ export default class FilingHistoryList extends Mixins(
         //   params: { filingId: filing.filingId, isCorrection: true } })
         // FOR NOW:
         this.$router.push({
-          name: Routes.CORRECTION,
+          name: routeName,
```
severinbeauvais, line 651: At the moment, Annual Report corrections should continue to be handled using the original route. (It is a local filing, not a redirect to Edit UI.)

Ditto for COD and COA corrections below.

src/interfaces/correction-filing-interface.ts
```diff
@@ -21,6 +21,7 @@ export interface CorrectionFilingIF {
     correctedFilingId: string
     correctedFilingType: string
     correctedFilingDate: string
+    type?: string // ? for firms only
```
severinbeauvais, line 24: This should be an enum.

severinbeauvais: Is this for firms only, or is type (client/staff) prompted for all corrections now, too?

src/routes.ts
```diff
@@ -83,6 +83,38 @@ export default [
       ]
     }
   },
+  {
+    path: '/correction?u=client',
```
severinbeauvais, line 87: These are internal routes for AR/COA/COD corrections. They should not need to change.

src/views/Correction.vue
```diff
@@ -363,6 +365,8 @@ export default class Correction extends Mixins(
     // this is the id of the original filing to correct
     this.correctedFilingId = +this.$route.params.correctedFilingId // number (may be NaN)
 
+    // this is the correction type of either client or staff
+    this.correctionUserType = this.$route.params.correctionUserType
```
severinbeauvais, line 369: Do you have a clear direction that changing this internal correction is in scope?

I think changing internal correction filings should be out of scope for now, especially since these will be changed to "external" (Edit UI) filings soon.

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -388,13 +388,13 @@ describe('Filing History List - misc functionality', () => {
     expect(vm.disableCorrection({ ...item, isFutureEffective: true })).toBe(true)
 
     // only fifth condition:
-    expect(vm.disableCorrection({ ...item, status: 'CORRECTED' })).toBe(true)
+    expect(vm.disableCorrection({ ...item, status: 'CORRECTED' })).toBe(false)
```
severinbeauvais, line 391: Why is this now expected to be false?

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -700,8 +700,32 @@ describe('Filing History List - redirections', () => {
     await flushPromises()
 
     // verify redirection
+    // const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id
+    // const createUrl = 'https://edit.url/BC1234567/correction/?correction-id=110514'
+    // expect(window.location.assign).toHaveBeenCalledWith(createUrl + '&accountid=' + accountId)
```
severinbeauvais, line 705: Can this be deleted?

Or is there a reason it should remain? (if so, add comment)

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit, Watch } from 'vue-property-decorator'
+import { mapGetters } from 'vuex'
+import { ContactInfo } from '@/components/common'
+
+@Component({
+  computed: {
+    // Property definition for runtime environment.
+    ...mapGetters(['isRoleStaff'])
```
severinbeauvais, line 43: Delete this if it's not used.

src/components/Dashboard/TodoList.vue
severinbeauvais: Let's discuss any time there is uncertainty so you don't code the wrong solution :)

tests/unit/FilingHistoryList.spec.ts
lewischenstudio: I forgot to delete them. My bad 😞

src/components/Dashboard/TodoList.vue
lewischenstudio: Removed `left`:
image

Kept `left`:
image

[Design link](https://preview.uxpin.com/363a49b33bdb2a314ccac0f18daad8969e93b337#/pages/150002369/simulate/sitemap)

The gap between the icon and "View Details" is larger with `left` than without it, so I removed it.

severinbeauvais: Looks good to me!

src/views/Correction.vue
lewischenstudio: Okay 👌

src/components/Dashboard/TodoList.vue
```diff
@@ -1326,7 +1328,7 @@ export default class TodoList extends Mixins(
           (item.correctedFilingType === FilingNames.CHANGE_OF_REGISTRATION) ||
           (item.correctedFilingType === FilingNames.REGISTRATION)
         ) {
-          // navigate to Edit UI to resume this correction
+          // navigate to Edit UI to resume this correction (depending on either CLIENT OR STAFF)
```
severinbeauvais, line 1331: This navigation does not depend on CLIENT or STAFF. What are you pointing out here?

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+@Component({
+  components: { ContactInfo }
+})
+export default class FileCorrectionDialog extends Vue {
+  // enums for template
+  readonly CorrectionTypes = CorrectionTypes
+
+  // local variables
+  private hasChosenCorrection = true
+  private correctionType = ''
```
severinbeauvais, line 49: Use enum for type here, and '' is not a valid enum value. You can use null.

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+    this.hasChosenCorrection = true
+    this.correctionType = ''
+  }
+
+  /**
+   * Emits event to Edit UI for correction.
+   * Redirect to start correction.
+   */
+  @Emit('redirect')
+  private emitRedirect (correctionType: string): void { }
```
severinbeauvais, line 94: Use enum for correctionType.

src/enums/correctionTypes.ts
```diff
@@ -0,0 +1,5 @@
+export enum CorrectionTypes {
+    NONE = '',
```
severinbeauvais, line 2: NONE is not a valid enum value, according to the schema.

Use null for "no value".

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -871,9 +897,7 @@ export default class FilingHistoryList extends Mixins(
       item.availableOnPaperOnly ||
       item.isTypeStaff ||
       item.isFutureEffective ||
-      this.isStatusCorrected(item) ||
       this.isTypeAlteration(item) ||
-      this.isTypeCorrection(item) ||
       this.isTypeTransition(item) ||
       // the following corrections are supported:
```
lewischenstudio, line 902: Line 874 and 876 are removed to allow correction filings that have been corrected or are correction types. This needs to be resolved after discussion.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -700,7 +723,10 @@ export default class FilingHistoryList extends Mixins(
         // FUTURE: allow a correction to a correction?
         // this.$router.push({ name: Routes.CORRECTION,
         //   params: { correctedFilingId } })
-        alert('At this time, you cannot correct a correction. Please contact Ops if needed.')
+        this.$router.push({
+          name: Routes.CORRECTION,
+          params: { correctedFilingId }
+        })
```
severinbeauvais, line 729: Please revert original code.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -871,9 +897,7 @@ export default class FilingHistoryList extends Mixins(
       item.availableOnPaperOnly ||
       item.isTypeStaff ||
       item.isFutureEffective ||
-      this.isStatusCorrected(item) ||
       this.isTypeAlteration(item) ||
-      this.isTypeCorrection(item) ||
```
severinbeauvais, line 876: I have not heard that we are allowing corrections to corrections or already-corrected filings, so please revert this code.

src/mixins/filing-mixin.ts
```diff
@@ -142,6 +142,7 @@ export default class FilingMixin extends Mixins(DateMixin) {
         correctedFilingId: correctedFiling.header.filingId,
         correctedFilingType: FilingTypes.REGISTRATION,
         correctedFilingDate: correctedFiling.header.date,
+        type: correctedFiling.correction.type,
```
severinbeauvais, line 145: You can only de-reference this if the corrected filing has a correction object, which is only true is the corrected filing is a correction, which is only true if we allow corrections to corrections, which we haven't agreed to (yet).

Regardless, I think this is indeed the correct place to set type. And maybe also in buildIaCorrectionFiling() if we allow CLIENT ve STAFF for those. Just keep this in mind for now since we haven't decided this either.

src/views/Correction.vue
```diff
@@ -692,14 +695,14 @@ export default class Correction extends Mixins(
         correctedFilingId: this.correctedFilingId,
         correctedFilingType: this.origFiling.header.name,
         correctedFilingDate: this.dateToYyyyMmDd(this.apiToDate(this.origFiling.header.date)),
-        comment: `${this.defaultComment}\n${this.detailComment}`
+        comment: `${this.defaultComment}\n${this.detailComment}`,
+        type: this.correctionType
```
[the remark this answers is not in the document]
lewischenstudio: Sure. I'll correct it once we have the decision.

src/mixins/filing-mixin.ts
lewischenstudio: Sure.

tests/unit/FilingHistoryList.spec.ts
lewischenstudio: I misunderstood it. I've reverted the two lines.

src/components/Dashboard/FilingHistoryList.vue
lewischenstudio: It's revereted.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -672,7 +688,14 @@ export default class FilingHistoryList extends Mixins(
           // build correction filing
           let correctionFiling: CorrectionFilingIF
           if (this.isBComp) correctionFiling = this.buildIaCorrectionFiling(origFiling)
-          if (this.isFirm) correctionFiling = this.buildFmCorrectionFiling(origFiling)
+          if (this.isFirm) {
+            correctionFiling = this.buildFmCorrectionFiling(origFiling)
+            if (correctionFiling.correction.type === CorrectionTypes.NONE) {
+              correctionFiling = { ...correctionFiling,
+                correction: { ...correctionFiling.correction, type: correctionType } }
+            }
```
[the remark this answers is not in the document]
lewischenstudio: I just realized that `type` will be available for `correctionFiling = this.buildFmCorrectionFiling(origFiling)`, so I removed these lines.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -700,7 +717,10 @@ export default class FilingHistoryList extends Mixins(
         // FUTURE: allow a correction to a correction?
         // this.$router.push({ name: Routes.CORRECTION,
         //   params: { correctedFilingId } })
-        alert('At this time, you cannot correct a correction. Please contact Ops if needed.')
+        this.$router.push({
+          name: Routes.CORRECTION,
+          params: { correctedFilingId }
+        })
```
severinbeauvais, line 723: This is still not reverted.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -399,6 +408,7 @@ export default class FilingHistoryList extends Mixins(
   private currentFiling: HistoryItemIF = null
   private loadingOneIndex = -1
   private isBusy = false
+  private correctionFiling: HistoryItemIF = null
```
severinbeauvais, line 411: Any chance you could reuse `currentFiling` (line 408) instead of declaring a new variable?

I am OK with a new variable except that it's named exactly the same as the correction filing on line 689, which may lead to confusion/errors.

src/components/Dashboard/FilingHistoryList.vue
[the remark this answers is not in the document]
severinbeauvais: I don't see how buildFmCorrectionFiling() gets the correction type, but maybe you have more code to commit...

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+        </template>
+      </v-card-text>
+
+      <v-card-actions class="py-10">
+        <v-btn outlined large
+          class="mr-2"
+          color="primary"
+          id="dialog-exit-button"
+          @click="exit()"
+        > Cancel </v-btn>
```
severinbeauvais, line 24: You shouldn't have spaces around the text.

With no spaces, you can "expect the text toBe('Cancel')" (otherwise you have to test with the spaces, or else use "toContain()" which is not exact).

If you did this to get extra whitespace around the text, use padding (or min-width) instead.

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit, Watch } from 'vue-property-decorator'
+import { ContactInfo } from '@/components/common'
+import { CorrectionTypes } from '@/enums'
+
+@Component({
+  components: { ContactInfo }
```
severinbeauvais, line 41: Not used. Please remove.

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+  private hasChosenCorrection = true
+  protected correctionType = null as CorrectionTypes
+
+  // Prop to display the dialog.
+  @Prop() readonly dialog: boolean
+
+  // Prop to provide attachment selector.
+  @Prop() readonly attach: string
+
+  private checkToStart () {
```
severinbeauvais, line 57: Since this is called from the template, it should protected not private.

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+        this.correctionType = CorrectionTypes.STAFF
+        this.hasChosenCorrection = true
+        break
+      default:
+        break
+    }
+  }
+
+  // Pass click event to parent.
+  @Emit() private exit () {
```
severinbeauvais, line 84: Since this is called from the template, it should protected not private.

src/components/dialogs/FileCorrectionDialog.vue
```diff
@@ -0,0 +1,116 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import "@/assets/styles/theme.scss";
+
+::v-deep .v-dialog {
+  .v-card {
+    .v-card__title {
+      color: black;
+      background: #ffff;
```
severinbeauvais, line 105: Use "white" instead of hex colour.

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -699,7 +699,26 @@ describe('Filing History List - redirections', () => {
     fileCorrectionItem.trigger('click')
     await flushPromises()
 
+    // verify to display the correction dialog modal
+    const fileCorrectionDialog = wrapper.find('#correction-filing-dialog')
+    expect(fileCorrectionDialog).toBeDefined()
+
+    // verify radio buttons for client and staff
```
severinbeauvais, line 706: This only verifies the client radio button.

tests/unit/FilingHistoryList.spec.ts
```diff
@@ -699,7 +699,26 @@ describe('Filing History List - redirections', () => {
...
+    const clientRadioBtn = wrapper.find('#correct-client-radio')
+    expect(clientRadioBtn).toBeDefined()
+
+    // click the staff radio button
+    const staffRadioBtn = wrapper.find('#correct-staff-radio')
+    expect(staffRadioBtn).toBeDefined()
+    staffRadioBtn.trigger('click')
+    await flushPromises()
+
     // verify redirection
```
severinbeauvais, line 716: If possible, please also test the type field in the draft correction filing. (You might be able to check if the build method was called and verify its return value.)

src/components/Dashboard/TodoList.vue
```diff
@@ -1075,6 +1075,7 @@ export default class TodoList extends Mixins(
         correctedFilingId: correction.correctedFilingId,
         // this is only used for external corrections (IA):
         correctedFilingType: this.filingTypeToName(correction.correctedFilingType as FilingTypes),
+        type: correction.type,
```
severinbeauvais, line 1078: What is this change for?

If this change is needed, please add this property to the Todo Item interface.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -636,7 +646,13 @@ export default class FilingHistoryList extends Mixins(
     if (index >= 0) this.togglePanel(index, this.historyItems[index])
   }
 
-  private async correctThisFiling (item: HistoryItemIF): Promise<void> {
+  private startCorrectionDialog (item: HistoryItemIF): void {
+    this.fileCorrectionDialog = true
+    this.correctionFiling = item
```
severinbeauvais, line 651: I recommend swapping these 2 lines so that the filing is set before the dialog is displayed.

(JS is single-threaded and run-to-completion, and these statements are atomic, so it probably works fine as-is but conceptually it makes more sense the other way IMO.)

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -636,7 +646,13 @@ export default class FilingHistoryList extends Mixins(
...
+  }
+
+  private async correctThisFiling (correctionType: CorrectionTypes): Promise<void> {
```
severinbeauvais, line 654: It looks like you don't do anything with correctionType. It should be saved in the draft correction filing.

src/mixins/filing-mixin.ts
severinbeauvais: If firm corrections always have a "correction" object then this code is fine and you can resolve this conversation.

PR #361, opened 2022-07-14
12872 Fixed incorrect correctedFilingType

severinbeauvais:
*Issue #:* /bcgov/entity#12872

*Description of changes:*
- fixed incorrect correctedFilingType
- app version = 5.2.13

### DIFF
--- a/src/mixins/filing-mixin.ts
+++ b/src/mixins/filing-mixin.ts
@@ -120,8 +120,8 @@ export default class FilingMixin extends Mixins(DateMixin) {
   }
 
   /**
-   * Builds a Firm Correction filing body from a Change of Registration filing
-   * or Registration filing. Used when creating a Firm Correction.
+   * Builds a Firm Correction filing body from a Registration or Change of Registration filing.
+   * Used when creating a Firm Correction.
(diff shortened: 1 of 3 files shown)


src/mixins/filing-mixin.ts
```diff
@@ -140,7 +140,7 @@ export default class FilingMixin extends Mixins(DateMixin) {
       },
       correction: {
         correctedFilingId: correctedFiling.header.filingId,
-        correctedFilingType: FilingTypes.REGISTRATION,
+        correctedFilingType: correctedFiling.header.name,
```
severinbeauvais, line 143: This might be `registration` or `changeOfRegistration`.

This is OK to be hard-coded on line 112 because atm only BEN IA corrections are supported, but when we make all BEN corrections go through this flow then that will need to be updated as well.

lambert-alex (approved):

lewischenstudio (approved):

src/mixins/filing-mixin.ts
severinbeauvais: Hopefully we'll remember when we rename that method from buildIaCorrectionFiling() to buildBenCorrectionFiling() and also change the code that calls it! 😁

PR #367, opened 2022-08-02
13197 Don't save Certified By in a draft correction

severinbeauvais:
*Issue #:* /bcgov/entity#13197

*Description of changes:*
- don't save Certified By in BEN or FM draft correction
- app version = 5.2.18

### DIFF
--- a/src/mixins/filing-mixin.ts
+++ b/src/mixins/filing-mixin.ts
@@ -99,7 +99,6 @@ export default class FilingMixin extends Mixins(DateMixin) {
     const correctionFiling: CorrectionFilingIF = {
       header: {
         name: FilingTypes.CORRECTION,
-        certifiedBy: correctedFiling.header.certifiedBy,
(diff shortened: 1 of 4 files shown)


lewischenstudio (approved):

src/mixins/filing-mixin.ts
```diff
@@ -129,7 +128,6 @@ export default class FilingMixin extends Mixins(DateMixin) {
     const correctionFiling: CorrectionFilingIF = {
       header: {
         name: FilingTypes.CORRECTION,
-        certifiedBy: correctedFiling.header.certifiedBy,
```
severinbeauvais, line 132: In both cases here, it's incorrect to set the Certified By as the original filing's certifier (ie, the correction may be certified by a different person, or no person if this is a staff-error correction).

The Edit UI should (and does) handle this case correctly.

PR #393, opened 2022-09-21
12671 show business has court order

lambert-alex:
*Issue #:* /bcgov/entity#12671

*Description of changes:*
 - version 5.6.2
 - add header showing that a business has a court order
 - set hasCourtOrders from the legal-api business response


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -26,6 +26,14 @@
       attach="#filing-history-list"
     />
 
+    <!-- Display for if a court order(s) have been filed -->
+    <v-card v-if="hasCourtOrders" class="my-6 pa-6" elevation="0">
```
severinbeauvais, line 30: Do you need to set elevation?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -26,6 +26,14 @@
...
+      <v-icon id="has-court-order-icon">mdi-gavel</v-icon>
```
severinbeauvais, line 31: We haven't been setting an id on v-icons. If you need this for testing, please find a different way to select it.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -26,6 +26,14 @@
...
+      <span>
+        Court order(s) have been filed on this company. Review the filing history for impacts to business informaiton.
+      </span>
```
severinbeauvais, line 34: Spans are supposed to be "inline elements" (not block elements), so the content "should be" on the same line if possible. Sometimes it matters for display, sometimes it doesn't.

One thing, though: in a unit test, the span's text will contain the line breaks here, so you may have to "expect to contain XXX" instead of the more specific "expect to be XXX".

Ideal:
`<span>blah blah blah</span>`

OK:
```
<span>this is the start of this sentence
  and this is the end</span>
```

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -26,6 +26,14 @@
       attach="#filing-history-list"
     />
 
+    <!-- Display for if a court order(s) have been filed -->
+    <v-card v-if="hasCourtOrders" class="my-6 pa-6" elevation="0">
+      <v-icon id="has-court-order-icon">mdi-gavel</v-icon>
+      <span>
+        Court order(s) have been filed on this company. Review the filing history for impacts to business informaiton.
```
severinbeauvais, line 33: Typo.

src/App.vue
```diff
@@ -558,6 +558,7 @@ export default {
       this.setLastDirectorChangeDate(business.lastDirectorChangeDate) // may be empty
       this.setWarnings(Array.isArray(business.warnings) ? business.warnings : [])
       this.setGoodStanding(business.goodStanding)
+      this.setHasCourtOrders(business.hasCourtOrders)
```
severinbeauvais, line 561: Is this property always available in the business object? If not, would it make sense to fall back to False here?

src/components/Dashboard/FilingHistoryList.vue
lambert-alex: Okay I will change this

src/components/Dashboard/FilingHistoryList.vue
lambert-alex: I was adding the id for this css because the color prop for v-icon wasn't using the custom $gray9 color.

lewischenstudio (approved):
LGTM. Please resolve the conversions before merging. It'll be great if you can throw in a test case for displaying the court order.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: OK! I don't remember using this before but I guess you need it to match the UI design.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I still think you should target this v-icon differently... or, add the gray9 colour to the Vuetify scheme (in main.ts).

src/App.vue
severinbeauvais: Great, then this is fine.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Maybe `.v-icon.mdi-gavel` ?

severinbeauvais (approved):
LGTM

Go ahead and merge when ready.

lambert-alex:
Confirmed with Cameron that there is no conflict with his PR

PR #395, opened 2022-09-22
13674 Upgrade to Vue 2.7 + misc fixes and cleanup

severinbeauvais:
*Issue #:* bcgov/entity#13674

*Description of changes:*

Commit 1:
- changed ::v-deep to :deep()
- added ! to props
- removed private from getters
- changed some private to protected
- removed some redundant type declarations
- removed private from readonlys
- changed { } to {}
- added flex class to SBC Header to fix height

Commit 2:
- updated misc dependencies (pre Vue 2.7)
- now builds from scratch!

Commit 3:
- updated vue to 2.7.10
- deleted obsolete vue-template-compiler
- updated http-status-codes
- updated vue-router
- updated vue2-filters
- updated vuetify
- updated eslint-plugin-vue
- updated @types/jest
- updated @typescript-eslint/eslint-plugin
- updated @typescript-eslint/parser
- updated eslint rules
- updated sinon
- cleaned up Add Staff Filing button
- fixed some interfaces
- changed slot="label" -> v-slot:label
- imported Vue from vue not vue-property-decorator
- fixed or disabled lint issues
- deleted some unused variables
- changed Array<Function> -> Array<(v) => boolean | string>
- deleted obsolete declaration files
- fixed unit test errors
- app version = 5.7.0


src/App.vue
```diff
@@ -69,7 +69,7 @@
       </div>
     </v-fade-transition>
 
-    <SbcHeader />
+    <SbcHeader class="d-flex" />
```
severinbeauvais, line 72: The main Sbc Header div had height 48px whereas its content had height 70px. This caused the breadcrumbs to be partly covered over by the Sbc Header. Changing the display to flex fixes this.

src/components/AnnualReport/AGMDate.vue
```diff
@@ -128,24 +128,19 @@ export default class AgmDate extends Mixins(DateMixin) {
   }
 
   /** New AGM Date (from a resumed draft). */
-  @Prop({ default: null })
-  readonly newAgmDate: string
+  @Prop({ default: null }) readonly newAgmDate!: string
```
severinbeauvais, line 131: The `!` was required in one of the files to fix a lint error.

Ref: https://blog.logrocket.com/understanding-exclamation-mark-typescript

I applied it to all props for consistency.

src/components/AnnualReport/AGMDate.vue
```diff
@@ -364,11 +359,11 @@ export default class AgmDate extends Mixins(DateMixin) {
 }
 
 // reduce date input height when there are no error messages
-::v-deep .v-text-field:not(.error--text) {
+:deep(.v-text-field:not(.error--text)) {
```
severinbeauvais, line 362: This is the new format (in Vue 2.7 anyway).

src/components/EntityInfo.vue
```diff
@@ -360,20 +360,6 @@ export default class EntityInfo extends Mixins(AllowableActionsMixin, CommonMixi
...
-
-::v-deep {
-  .v-breadcrumbs a {
-    color: $gray8 !important;
-  }
-
-  .v-breadcrumbs a:hover {
-    color: $BCgovABlue3 !important;
-  }
-}
```
severinbeauvais, line 375: The breadcrumbs are no longer in this component. (They are their own shared component, rendered in App.vue.)

thorwolpert (approved):
i like and appreciated the comments when ever a change came in. easy to follow why and the link was a good read.

package.json
```diff
@@ -43,40 +43,40 @@
...
     "typescript": "^3.9.7",
-    "vue-class-component": "^7.2.5",
-    "vue-cli-plugin-vuetify": "^2.0.7",
+    "vue-class-component": "^7.2.6",
+    "vue-cli-plugin-vuetify": "^2.5.8",
     "vue-debounce-decorator": "^1.0.1",
     "vue-plugin-helper-decorator": "^0.0.12",
-    "vue-property-decorator": "^8.5.1",
+    "vue-property-decorator": "^9.1.2",
     "vue-template-compiler": "2.6.12",
```
lambert-alex, line 77: I don't think the vue-template-compiler is needed for vue 2.7.0
> You can also remove vue-template-compiler from the dependencies - it is no longer needed in 2.7.
https://blog.vuejs.org/posts/vue-2-7-naruto.html

lambert-alex (approved):
Looks awesome!

package.json
severinbeauvais: You're right. At the moment this is still Vue 2.6.x... but it installs and builds and tests cleanly.

I am doing a phased upgrade in this PR. My next update will require code changes (esp unit tests)!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.6.1",
+  "version": "6.0.0",
```
seeker25, line 3: Maybe 5.7.0? Otherwise we'll have to upgrade another major for vue3?

severinbeauvais: I'm considering it, but given the number of package changes, I'm thinking this will require a ton of regression testing... so I'm still thinking it's a major version update.

What do other devs think?

src/App.vue
severinbeauvais: This went away with the latest commit (to Vue 2.7).

src/components/AnnualReport/AGMDate.vue
severinbeauvais: ^^ This doesn't work in Vue 2.6.x but I tested it in Vue 2.7 and it's good.

package.json
lambert-alex: These changes break backward compatibility, so I would say following the semantic versioning [practices](https://semver.org/) this is a major release

severinbeauvais: Excellent point, Alex.

src/components/AnnualReport/AGMDate.vue
```diff
@@ -162,19 +158,19 @@ export default class AgmDate extends Mixins(DateMixin) {
   private backupDate = '' // for toggling No AGM
 
   /** The array of validations rule(s) for the AGM Date text field. */
-  private get agmDateRules (): Array<Function> {
+  get agmDateRules (): Array<(v) => boolean | string> {
```
severinbeauvais, line 161: Can no longer use Function.

Future: declare a type for this.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div id="staff-notation" class="pr-6" :class="{ 'add-scrollbar-offset': addScrollbarOffset }">
```
severinbeauvais, line 2: Small fix per Yui: don't offset this, so that it's in line with the buttons in the Filing History List.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -43,29 +43,16 @@
...
-            <v-btn
-              text
-              v-on="on"
-              class="menu-btn"
-              :class="{active: expand}"
-              @click="expand = !expand"
-            >
-              <v-icon>mdi-menu-down</v-icon>
-            </v-btn>
-          </span>
```
severinbeauvais, line 67: Approved by Yui.

cameron-freshworks (approved):
Looks good, nice work!

src/components/DigitalCredentials/CredentialsLanding.vue
```diff
@@ -1,5 +1,4 @@
 <template>
-  <!-- Digital Credentials Landing -->
```
severinbeauvais, line 2: This caused a parsing issue (probably because only 1 top level element is allowed).

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -14,13 +14,13 @@
         :headers="credentialsTableHeaders"
         no-data-text="All digital credentials associated with this business will appear here."
       >
-        <template v-slot:item.credentialType="{ item }">
+        <template v-slot:[`item.credentialType`]="{ item }">
```
severinbeauvais, line 17: Lint / parsing fix.

tests/unit/Certify.spec.ts
```diff
@@ -62,6 +63,7 @@ function createComponent (
   store.state.currentDate = currentDate
 
   return mount(Certify, {
+    vuetify,
```
severinbeauvais, line 66: This caused all manners of problems with updated imports. I have no idea why this worked before (since the tests expect Vuetify components to be rendered not stubbed).

package.json
seeker25: I don't think it breaks backwards compatibility.  

If you were to revert the changes from this PR, 5.6.2 and "6.0.0" the application would still work using the same API.

This isn't a library like BCRS shared components or SBC Common components, so I think it should be treated differently.

severinbeauvais: There are code changes in here (in the templates) that are not backward compatible.

That said, everyone now knows what I've done in here, so we don't _need_ to increase the major version number for visibility purposes. I'd be ok with going to 5.7.0 if the other reviewers agree.

seeker25: From a library perspective or an application perspective?

severinbeauvais: I'll think about it and make a decision on Monday 😀

seeker25 (approved):

PR #397, opened 2022-09-23
Digital Credential Api integration + Completion

cameron-freshworks:
*Issue #:* /bcgov/entity#13458

*Description of changes:*
* Renamed and updated steps to adhere to update flow requirements
* Implements API integration for Digital Credential Endpoints


src/components/DigitalCredentials/CredentialsLanding.vue
```diff
@@ -42,17 +42,20 @@
...
+          />
         </v-col>
       </v-row>
     </article>
   </section>
 </template>
 
 <script lang="ts">
-import { Component, Emit, Vue } from 'vue-property-decorator'
+import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
```
severinbeauvais, line 55: Please `import Vue from 'vue'` instead (to avoid conflicts with my PR).

src/components/DigitalCredentials/CredentialsLanding.vue
```diff
@@ -61,6 +64,9 @@ import { ImportantMessage } from '@/components/common'
   }
 })
 export default class CredentialsLanding extends Vue {
+  @Prop({ default: [] })
+  readonly issuedCredentials: Array<DigitalCredentialsIF>
```
severinbeauvais, line 68: Please make this 1 line and add the Typescript null assertion character.

src/assets/styles/layout.scss
```diff
@@ -81,6 +81,11 @@
   }
 }
 
+// Data table column helper class
+.column-lg {
+  width: 15rem;
+}
```
severinbeauvais, line 87: Please move this to the (single) component where it's used.

If multiple components declare the same class style then I'm in favour of moving it to base.scss or layout.scss.

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
         fixed
         disable-sort
         hide-default-footer
-        :items="credentialRecords"
-        :headers="!!credentialRecords ? credentialsTableHeaders : []"
+        :items="issuedCredentials"
+        :headers="!!issuedCredentials ? credentialsTableHeaders : []"
```
severinbeauvais, line 14: I don't think you can check for empty arrays like this. You may have to check length instead.

Please test and lmk what you intend here and if it works as you expect.

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
...
+import { Component, Mixins, Prop } from 'vue-property-decorator'
+import { DigitalCredentialsIF, TableHeaderIF } from '@/interfaces'
+import { DigitalCredentialTypes } from '@/enums'
+import { DateMixin } from '@/mixins'
 
 @Component({})
-export default class CredentialsTable extends Vue {
+export default class CredentialsTable extends Mixins(DateMixin) {
+  @Prop({ default: [] })
+  readonly issuedCredentials: Array<DigitalCredentialsIF>
```
severinbeauvais, line 40: Please make this 1 line and add the Typescript null assertion character.

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
...
+
   get credentialsTableHeaders (): Array<TableHeaderIF> {
     // Do not display headers if there is table data
-    if (this.credentialRecords.length === 0) return []
+    if (this.issuedCredentials.length === 0) return []
```
severinbeauvais, line 44: Ah-ha! You check this here. Prob don't need code on line 14 then.

src/assets/styles/layout.scss
severinbeauvais: Wait -- old code used `column-md`... where did that come from? Vuetify?

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
...
-      },
-      {
-        name: '08123123 B.C. LTD',
-        type: 'Business Relationship Credential',
-        status: 'Issued',
-        date: 'August 15, 2022',
-        credId: '1234-5678-9809'
-      }
-    ]
+  formatCredentialType (credentialType: DigitalCredentialTypes): string {
```
severinbeauvais, line 69: Protected.

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
...
+    return credentialType.charAt(0).toUpperCase() + credentialType.slice(1) + ' Credential'
```
severinbeauvais, line 70: Is `credentialType` guaranteed to not be null?

Does this code work if `credentialType` is empty string?

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
...
+  }
+
+  formatStatus (isIssued: boolean): string {
```
severinbeauvais, line 73: Protected.

src/components/DigitalCredentials/CredentialsTable.vue
```diff
@@ -10,72 +10,68 @@
...
+    return isIssued ? 'Issued' : 'Pending'
```
severinbeauvais, line 74: This is so simple, and used only in 1 place. Why not move it there and do away with this function?

src/components/DigitalCredentials/IssueCredentials.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+  <section id="issue-credentials" class="py-8">
```
severinbeauvais, line 2: For maintaining layouts, it's better if the top element (here, section) doesn't have layout CSS -- move it to the component that uses this instead.

src/components/DigitalCredentials/CredentialsLanding.vue
cameron-freshworks: Updated!

cameron-freshworks: Edit* for all added components

src/components/DigitalCredentials/IssueCredentials.vue
```diff
@@ -0,0 +1,130 @@
...
+  readonly issuedCredentials: Array<DigitalCredentialsIF>
+
+  readonly DigitalCredentialTypes = DigitalCredentialTypes
+  private isSendingOffer = false
+
+  get isPendingAcceptedOffer (): boolean {
+    return !this.issuedCredentials[0]?.isIssued
+  }
+
+  @Emit() back (): void {}
```
severinbeauvais, line 102: Please change this to `@Emit() protected back (): void {}`.
(Consistent with the other changes in my PR.)

src/components/DigitalCredentials/IssueCredentials.vue
```diff
@@ -0,0 +1,130 @@
...
+
+  @Emit('sendCredentialOffer')
+  private sendCredentialOffer (credentialType: DigitalCredentialTypes) {
```
severinbeauvais, line 105: Protected.

src/components/DigitalCredentials/CredentialsLanding.vue
cameron-freshworks: Updated!

src/assets/styles/layout.scss
cameron-freshworks: It had no definition in this project yet, not a Vuetify class. 
These are defined in Assets for width consistencies in the many tables etc
The intention was to make it global for the same purpose but totally fine with putting it local to the Credentials table.
Updated!

src/components/DigitalCredentials/IssueCredentials.vue
```diff
@@ -0,0 +1,130 @@
...
+                  <v-btn
+                    v-if="issuedCredentials.length === 0"
+                    id="next-btn"
+                    large color="primary"
+                    class="font-weight-bold px-6 float-right"
+                    :loading="isSendingOffer"
+                    @click="sendCredentialOffer(DigitalCredentialTypes.BUSINESS)"
+                  >Send Credential Offer</v-btn>
+
+                  <span v-else-if="isPendingAcceptedOffer" class="font-weight-bold alert-text">
```
severinbeauvais, line 46: Should this be a div with the text on line 48 inside a span?

Same for line 51.

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: removed! As you noted below, was already doing a check for this :)

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Good catch! Removed the template logic

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Updated!

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Updated

src/components/DigitalCredentials/IssueCredentials.vue
severinbeauvais: Same for line 70.

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Nothing is a guarantee i suppose! 
That would only happen in the event of a an api error/bad data etc. 

In the scenario of an empty string, it wouldn't result in a type error however. Would return 'Credential'

src/components/DigitalCredentials/RegisterWallet.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <section id="register-wallet" class="py-8">
```
severinbeauvais, line 2: Move layout class to parent.

src/components/DigitalCredentials/RegisterWallet.vue
```diff
@@ -0,0 +1,79 @@
...
+          </v-row>
+        </v-card-text>
+      </v-card>
+    </article>
+
+  </section>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator'
```
severinbeauvais, line 55: `import Vue from 'vue'`

src/components/DigitalCredentials/RegisterWallet.vue
```diff
@@ -0,0 +1,79 @@
...
+import QrcodeVue from 'qrcode.vue'
+
+@Component({
+  components: {
+    QrcodeVue
+  }
+})
+export default class RegisterWallet extends Vue {
+  @Prop({ default: false })
+  readonly credentialInvitationUrl: boolean
```
severinbeauvais, line 65: Single line + !

Also on line 67.

src/components/DigitalCredentials/IssueCredentials.vue
```diff
@@ -0,0 +1,130 @@
...
+  p, span, i {
+    color: $gray9;
+  }
+  #no-credentials-msg {
+    &:hover {
+      cursor: pointer;
+    }
+  }
+  .alert-text {
+    color: $BCgovGold9
```
severinbeauvais, line 124: Add ; to end.

Also line 127.

src/components/DigitalCredentials/RegisterWallet.vue
```diff
@@ -0,0 +1,79 @@
...
+  @Prop({ default: false })
+  readonly hasRegisteredWallet: boolean
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+#register-wallet {
+  p, span {
+    color: $gray9
```
severinbeauvais, line 76: Add ; to end.

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Updated!

src/enums/index.ts
```diff
@@ -16,6 +16,7 @@ export * from './paymentMethod'
 export * from './roles'
 export * from './routes'
 export * from './saveErrorReasons'
+export * from './digitalCredentialTypes' // Digital credentials pilot enum
```
severinbeauvais, line 19: Don't need comment.

src/enums/routes.ts
```diff
@@ -10,6 +10,6 @@ export enum Routes {
   // Digital Credential Routes for Pilot
```
severinbeauvais, line 10: Don't need comment.

src/mixins/legal-api-mixin.ts
```diff
@@ -254,4 +254,59 @@ export default class LegalApiMixin extends Mixins(CommonMixin) {
...
+   * @param businessId The business identifier (aka entity inc no)
+   * @param credentialType The credential offer type
+   * @returns a promise to return the credentials connection data
+   */
+  async issueCredentialOffer (businessId: string, credentialType: DigitalCredentialTypes): Promise<any> {
+    const url = `businesses/${businessId}/digitalCredentials/${credentialType}`
+    return axios.post(url)
+      .catch(error => {
+        console.log(error.message)
+      })
```
severinbeauvais, line 310: These all look good.

Do the callers handle null/undefined return values in case the network calls fails? (I ask because the catch blocks absorb the error and don't return anything.)

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Totally, will do. 
Since this is already within a slot a little extra logic in the template is fine.

src/views/DigitalCredentials.vue
```diff
@@ -139,6 +149,49 @@ export default class DigitalCredentials extends Vue {
   cancel (): void {
     this.$router.push({ path: `/${Routes.DIGITAL_CREDENTIALS}` })
   }
+
+  async getCredentials (): Promise<void> {
```
severinbeauvais, line 153: Functions called from the template should be protected. Please check all in this file.

src/views/DigitalCredentials.vue
```diff
@@ -139,6 +149,49 @@ export default class DigitalCredentials extends Vue {
...
+    const { data } = await this.fetchCredentials(this.getIdentifier)
+    if (data && data.issuedCredentials) {
```
severinbeauvais, line 155: Might as well use conditional chaining as on line 162 :)

src/views/DigitalCredentials.vue
```diff
@@ -139,6 +149,49 @@ export default class DigitalCredentials extends Vue {
...
+    const credentialIssued = await this.issueCredentialOffer(this.getIdentifier, credentialType)
+    if (credentialIssued) {
+      await this.getCredentials()
+    }
+  }
+
+  // Keep credential data in sync when navigating between routes.
+  // This is required due to the nature of external interactions between mobile device and api
+  @Watch('$route', { immediate: true })
+  async syncCredentials (): Promise<void> {
```
severinbeauvais, line 184: Should be private.

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: Good call, removed all step components and applied to router-view

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: Sounds good, updated!

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: updated

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: It could be for sure, i guess the styling here only applies to the text anyways, so make sense

src/components/DigitalCredentials/RegisterWallet.vue
cameron-freshworks: updated

src/components/DigitalCredentials/RegisterWallet.vue
cameron-freshworks: updated

src/components/DigitalCredentials/RegisterWallet.vue
cameron-freshworks: updated

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: Fixed

src/components/DigitalCredentials/RegisterWallet.vue
cameron-freshworks: fixed

src/enums/index.ts
cameron-freshworks: K

src/enums/routes.ts
cameron-freshworks: k, removed

src/mixins/legal-api-mixin.ts
cameron-freshworks: Definitely intended, due to the nature of these requests. 
The callers do handle these but there isn't any prompts with regards to errors because 404s etc are expected. 

The jist, the UI tries to keep in sync with whats going on between api and users mobile wallet by looking up data between steps.

src/views/DigitalCredentials.vue
cameron-freshworks: Updated!

src/views/DigitalCredentials.vue
cameron-freshworks: Doh fixed

src/views/DigitalCredentials.vue
cameron-freshworks: Updated!

src/components/DigitalCredentials/CredentialsTable.vue
severinbeauvais: Do you think conditional chaining would make this safe?

Or, you could return 'Unknown' if credentialType is falsey. (Like many safety checks we do.)

src/components/DigitalCredentials/IssueCredentials.vue
severinbeauvais: My rationale: Divs are block elements while spans are inline elements. Spans inside divs are typical. Spans inside spans is weird :) I think it's better to use divs where possible. 

Ref: https://www.w3schools.com/html/html_blocks.asp

src/components/DigitalCredentials/CredentialsTable.vue
cameron-freshworks: Totally! Updated

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: Sounds good to me, makes perfect sense! Updated

severinbeauvais: If this messed up your layout, since the class is now in the class, you can use template elements for the v-ifs etc.

vysakh-menon-aot (approved):

severinbeauvais (approved):
Nice a clean. I love it!

GTG

src/components/DigitalCredentials/IssueCredentials.vue
cameron-freshworks: All good! The div is conditional, so all the same.

cameron-freshworks:
> Nice a clean. I love it!
> 
> GTG

Thank you for the thorough review!

PR #403, opened 2022-10-06
Options API to Composition API

lambert-alex:
*Issue #:* /bcgov/entity#13816

*Description of changes:*
- example of a composables instead of mixins
- simple and more complex example of script setup
- some incomplete changes to router, store and auth,


src/components/AnnualReport/AGMDate.vue
```diff
@@ -112,193 +112,148 @@
...
-import { Component, Mixins, Prop, Watch, Emit } from 'vue-property-decorator'
-import { State, Getter } from 'vuex-class'
-import { DateMixin } from '@/mixins'
+<script setup lang="ts">
+import { reactive } from 'vue'
+import { AllowableActionsComposable, DateComposable } from '@/composables'
 import { FormIF } from '@/interfaces'
-
-@Component({})
-export default class AgmDate extends Mixins(DateMixin) {
```
lambert-alex, line 123: The composition API allows hacky versions of implementing inheritance functionality, initially switching to the composition api will lead to excess code   ```@Getter getCurrentDate!: string``` becomes ```const { getCurrentDate } = DateComposable``` or  ```const { getCurrentDate } = FilingComposable()``` Both work and lead to code having to be written in in DateComposable and FilingComposable which could be refactored to reduce redundant code after the inital switch

lambert-alex, line 123: script setup does not export in the composition api and instead the values are used directly

package.json
```diff
@@ -41,11 +41,12 @@
     "v-mask": "^2.3.0",
     "vue": "^2.7.10",
     "vue-affix": "^0.5.2",
-    "vue-router": "^3.6.5",
+    "vue-router": "^4.1.5",
+    "vue-template-compiler": "^2.7.10",
```
severinbeauvais, line 45: This is a dev dependency.

src/components/AnnualReport/ARDate.vue
```diff
@@ -9,20 +9,12 @@
...
-
-@Component({})
-export default class ArDate extends Mixins(DateMixin) {
-  @State nextARDate!: string
-  @Getter isBComp!: boolean
-  @Getter getCurrentDate!: string
-}
+const { formatYyyyMmDd } = DateComposable()
+const { isBComp } = AllowableActionsComposable() // probably can pull from state getters
+const { getCurrentDate } = FilingComposable()
```
severinbeauvais, line 17: Should  come from Vuex getters.

src/components/AnnualReport/ARDate.vue
```diff
@@ -9,20 +9,12 @@
...
-import { State, Getter } from 'vuex-class'
+<script setup lang="ts">
+import { DateComposable, AllowableActionsComposable, FilingComposable } from '@/composables'
 
-// Mixins
-import { DateMixin } from '@/mixins'
-
-@Component({})
-export default class ArDate extends Mixins(DateMixin) {
-  @State nextARDate!: string
```
severinbeauvais, line 22: Where does this come from?

src/components/common/ContactInfo.vue
```diff
@@ -18,12 +18,7 @@
   </ul>
 </template>
 
-<script lang="ts">
-import Vue from 'vue'
-import { Component } from 'vue-property-decorator'
-
-@Component({})
-export default class ContactInfo extends Vue {}
```
severinbeauvais, line 26: How does this component get named?

src/composables/allowable-actions-composable.ts
```diff
@@ -0,0 +1,122 @@
...
+  const isBcCompany = computed(() => store.getters['isBcCompany'] as boolean)
+  const isUlc = computed(() => store.getters['isUlc'] as boolean)
+  const isHistorical = computed(() => store.getters['isHistorical'] as boolean)
+  const isRoleStaff = computed(() => store.getters['isRoleStaff'] as boolean)
+  const isSoleProp = computed(() => store.getters['isSoleProp'] as boolean)
+  const isPartnership = computed(() => store.getters['isPartnership'] as boolean)
+  const isFirm = computed(() => store.getters['isFirm'] as boolean)
+  const isCoop = computed(() => store.getters['isCoop'] as boolean)
+  const hasBlockerExceptStaffApproval = computed(() => store.getters['hasBlockerExceptStaffApproval'] as boolean)
+  const isGoodStanding = computed(() => store.getters[''] as boolean)
```
severinbeauvais, line 22: This is really repetitive. Is there a way to combine these and just import/declare a single thing?

severinbeauvais: Or we could write a helper?
Or is there a decorator for this somewhere?

src/composables/common-composable.ts
```diff
@@ -0,0 +1,41 @@
...
+  /** True if Jest is running the code. */
+  const isJestRunning = computed(() => process.env.JEST_WORKER_ID !== undefined as boolean)
+
+  /**
+   * Removes the specified properties from nested objects.
+   * @param baseObj the base object
+   * @param keys    the nested object keys which to omit properties from
+   * @param prop    the properties to be removed
+   */
+  const omitProps = (baseObj: any, keys: Array<string>, prop: Array<string>): any => {
```
severinbeauvais, line 17: Please set type of baseObj to `object`.

src/composables/date-composable.ts
```diff
@@ -0,0 +1,379 @@
...
+export const DateComposable = () => {
+  const getCurrentJsDate = computed(() => store.getters['getCurrentJsDate'] as Date)
+
+  /**
+   * Fetches and returns the web server's current date (in UTC).
+   * Used to bypass the user's local clock/timezone.
+   * Ref: https://www.npmjs.com/package/serverdate
+   * @returns a promise to return a Date object
+   */
+  const getServerDate = async (): Promise<Date> => {
```
severinbeauvais, line 19: Do we have to use this function format or can we use this?
```
async function getServerDate (): Promise<Date> {
 ...
}
```
I think this ^^^ is easier to read. What do you think?

src/composables/director-composable.ts
```diff
@@ -0,0 +1,18 @@
...
+   */
+  const fieldSorter = (fields: Array<string>) => {
+    return (a, b) => fields.map(o => {
+      return a['officer'][o] > b['officer'][o] ? 1 : a['officer'][o] < b['officer'][o] ? -1 : 0
+    }).reduce((p, n) => p || n, 0)
+  }
+  return {
+    fieldSorter
+  }
+}
```
severinbeauvais, line 18: Do you think this is better as a composable or would it be better as a standalone util function like these?

https://github.com/bcgov/business-edit-ui/tree/main/src/utils

src/router.ts
```diff
@@ -1,27 +1,23 @@
...
+  const router = createRouter({
+    history: createWebHistory(base),
     // set base URL for Vue Router
-    base,
     routes,
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     scrollBehavior (to, from, savedPosition) {
       // see https://router.vuejs.org/guide/advanced/scroll-behavior.html
-      return { x: 0, y: 0 }
+      return { left: 0, top: 0 }
```
severinbeauvais, line 20: Wow, you've done a ton of R&D to figure all this out. Nice work! 👍

severinbeauvais (commented):
Nice work!

I'm looking forward to what the next steps should be in our eventual move to Vue3 -- keeping in mind what we _must_ do versus what we _could_ do.

src/components/common/DetailComment.vue
```diff
@@ -16,67 +16,61 @@
   </v-card>
 </template>
 
-<script lang="ts">
-import Vue from 'vue'
-import { Component, Prop, Watch, Emit } from 'vue-property-decorator'
-import { Debounce } from 'vue-debounce-decorator'
```
severinbeauvais, line 22: I created [13853](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/13853) to investigate what we can do about this.

This is pretty useful to limit the amount of validation/events that need to be processed.

severinbeauvais: Oh wait -- the lodash method works?

src/components/common/DetailComment.vue
```diff
@@ -16,67 +16,61 @@
...
+<script setup lang="ts">
+import _ from 'lodash'
+import { computed, ref, watch, defineEmits } from 'vue'
```
severinbeauvais, line 21: Does `defineProps` need to be imported here as well?

src/components/AnnualReport/AGMDate.vue
severinbeauvais: Can we use a mix of Composition API and Options API?

Can we use a mix of Composables and Mixins?

lambert-alex: From what I understand the options api uses the composition api under the hood, so yes you can use both together. However that can lead to issues initially because all of the composables and components  will be initialized first. Even before the beforeCreate() in the options api.

src/components/common/DetailComment.vue
lambert-alex: Ya it looks like the lodash method works for this.

src/components/common/DetailComment.vue
lambert-alex: No and defineEmits doesn't need to be there either. Both are part of script setup

src/composables/allowable-actions-composable.ts
lambert-alex: Yes this is way to repetitive.  This is what I mean when I think we would need to do a major refactor. As code used anywhere can be reactive this could be replaced with 

const { getEntityType, hasBlcker, ect.... } = getters where all the computed methods are stored

src/composables/date-composable.ts
lambert-alex: The function format is better. I agree.

src/composables/director-composable.ts
lambert-alex: I am not sure. It very well might be. Right now what I have converted over is the composition api with a structure of an options api app. Which I don't think uses the composition api to its full potential

src/components/common/ContactInfo.vue
lambert-alex: The component is named as a variable when it is imported.

i.e. 
```import ContactInfo from './ContactInfo.vue
<ContactInfo />```  works for this component but 

``` import SomeRandomName from './ContactInfo.vue'
<SomeRandomName />``` works exactly the same

src/components/AnnualReport/ARDate.vue
lambert-alex: Yes it does, I only used the getCurrentDate from FilingComposable because I hadn't implemented the getCurrentDate in the Vuex getters and I wanted a complete component

src/components/common/ContactInfo.vue
severinbeauvais: So then we can't do `import * from <whatever>` ?

src/components/AnnualReport/ARDate.vue
lambert-alex: I didn't implement this. I am not sure of the best way to implement state it seems like one of the advantages of the composition api is the ability to share states between different components. So I am not really sure what the best way is to handle ```@state```

src/components/AnnualReport/ARDate.vue
severinbeauvais: Ok.

It might be good to be concise moving forward.

src/components/AnnualReport/ARDate.vue
severinbeauvais: Accessing state directly was a cheat, left over from the initial implementation of this UI. It's much better to use getters.

src/components/common/ContactInfo.vue
lambert-alex: I think you do would do ```import * as All from ./components <All.Foo /> <All.Bar />``` There may be a more elegant solution but I haven't found one yet

src/components/AnnualReport/ARDate.vue
lambert-alex: Yes I agree

src/components/AnnualReport/ARDate.vue
lambert-alex: Oh okay! This could possibly be a good refactor to do before changing API's I am making some tickets for that. I can add this.

src/components/common/ContactInfo.vue
severinbeauvais: Ok

src/components/AnnualReport/AGMDate.vue
severinbeauvais: OK, thanks.

What about using Composables and Mixins at the same time... possible?

lambert-alex: I don't see any reason why you couldn't.

PR #424, opened 2022-11-29
14403 Added invalid NR messages logging

severinbeauvais:
*Issue #:* bcgov/entity#14403

*Description of changes:*
- changed isNrValid -> isNrInvalid and now returns error message
- updated unit tests
- app version = 5.7.24

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -674,10 +674,11 @@ export default {
     },
 
     storeNrData (nr: NameRequestIF, ia: any): void {
-      // check if NR is valid
-      if (!this.isNrValid(nr)) {
+      // verify that NR is valid
+      const error = this.isNrInvalid(nr)
+      if (error) {
         this.nameRequestInvalidDialog = true
-        throw new Error('Invalid NR data')
+        throw new Error(error)
       }
 
       // verify that NR type matches entity type from application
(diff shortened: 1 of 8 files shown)


src/mixins/name-request-mixin.ts
```diff
@@ -9,20 +9,19 @@ import { NameRequestIF } from '@/interfaces'
 @Component({})
```
severinbeauvais, line 9: This file is now exactly the same as in Create UI... in preparation for moving it to shared components library.

src/App.vue
```diff
@@ -674,10 +674,11 @@ export default {
...
 
     storeNrData (nr: NameRequestIF, ia: any): void {
-      // check if NR is valid
-      if (!this.isNrValid(nr)) {
+      // verify that NR is valid
+      const error = this.isNrInvalid(nr)
+      if (error) {
         this.nameRequestInvalidDialog = true
-        throw new Error('Invalid NR data')
+        throw new Error(error)
```
ChrisGabel23, line 681: Won't this just log "true" as an error? I'm not sure what Error() does, but if its old argument was a string containing info on the error, would logging "true" be beneficial?

ChrisGabel23 (approved):

src/App.vue
severinbeauvais, line 681: This won't throw "true" because error is actually a string. The condition on line 679 passes if the string is truthy or fails if error is empty/null/undefined.

lewischenstudio (approved):

PR #427, opened 2022-12-02
14350 Add alteration type to ToDo

lambert-alex:
*Issue #:* /bcgov/entity#14350

*Description of changes:*
 - Add missing alteration type to ToDo
 - update app version -> 5.7.26

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -189,7 +189,7 @@
                     :disabled="!item.enabled"
                     @click.native.stop="doResumeFiling(item)"
                   >
-                    <span>Resume</span>
+                    <span>Resume+++</span>
                   </v-btn>
 
                   <!-- dropdown menu -->
@@ -1357,6 +1357,7 @@ export default class TodoList extends Vue {
 
   /** Resumes a draft filing. */
   protected doResumeFiling (item: TodoItemIF): void {
+    console.log(item)
     switch (item.name) {
       case FilingTypes.ANNUAL_REPORT:
         // resume this Annual Report locally
@@ -1385,10 +1386,12 @@ export default class TodoList extends Vue {
         switch (item.correctedFilingType) {
           case FilingTypes.ALTERATION:
           case FilingNames.INCORPORATION_APPLICATION:
+          case FilingNames.CHANGE_OF_COMPANY_INFO:
           case FilingNames.CHANGE_OF_REGISTRATION:
           case FilingNames.CORRECTION:
           case FilingNames.REGISTRATION:
             // resume correction via Edit UI
+            console.log('in edit ui redirect.')
             const correctionUrl = `${this.editUrl}${this.getIdentifier}/correction/?correction-id=${item.filingId}`
             navigate(correctionUrl)
             break
@@ -1398,6 +1401,7 @@ export default class TodoList extends Vue {
           case FilingTypes.CHANGE_OF_DIRECTORS:
           case FilingTypes.CONVERSION:
           default:
+            console.log('defaulting', item.correctedFilingType)
             // resume local correction for all other filings
             this.setCurrentFilingStatus(FilingStatus.DRAFT)
             this.$router.push({
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -1383,6 +1383,7 @@ export default class TodoList extends Vue {
       case FilingTypes.CORRECTION:
         // see also FilingHistoryList.vue:correctThisFiling()
         switch (item.correctedFilingType) {
+          case FilingTypes.ALTERATION:
           case FilingNames.INCORPORATION_APPLICATION:
           case FilingNames.CHANGE_OF_REGISTRATION:
           case FilingNames.CORRECTION:
@@ -1393,7 +1394,6 @@ export default class TodoList extends Vue {
             break
 
           case FilingTypes.ANNUAL_REPORT:
-          case FilingTypes.ALTERATION:
           case FilingTypes.CHANGE_OF_ADDRESS:
           case FilingTypes.CHANGE_OF_DIRECTORS:
           case FilingTypes.CONVERSION:
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -189,7 +189,7 @@
                     :disabled="!item.enabled"
                     @click.native.stop="doResumeFiling(item)"
                   >
-                    <span>Resume+++</span>
+                    <span>Resume</span>
                   </v-btn>
 
                   <!-- dropdown menu -->
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-filings-ui",
-  "version": "5.7.24",
+  "version": "5.7.25",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "business-filings-ui",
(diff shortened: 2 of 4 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -1385,10 +1386,12 @@ export default class TodoList extends Vue {
         switch (item.correctedFilingType) {
           case FilingTypes.ALTERATION:
           case FilingNames.INCORPORATION_APPLICATION:
+          case FilingNames.CHANGE_OF_COMPANY_INFO:
```
lambert-alex, line 1389: I hadn't realized CHANGE_OF_COMPANY_INFO was a type of alteration. Could it be worth changing the name to ALTERATION_OF_COMPANY_INFO to prevent this error in the future?

src/components/Dashboard/TodoList.vue
```diff
@@ -1357,6 +1357,7 @@ export default class TodoList extends Vue {
 
   /** Resumes a draft filing. */
   protected doResumeFiling (item: TodoItemIF): void {
+    console.log(item)
```
severinbeauvais, line 1360: Don't commit this.

severinbeauvais: Or the other 2 below.

src/components/Dashboard/TodoList.vue
severinbeauvais: First, I'm not sure what kind of filing this is, nor how we initialize it (ie, not like existing alteration).

Second, that's a question for UI team, since that enum currently maps to "Alteration". I know the business is particular about filing names, but they may be open to change in order to improve the user experience.

severinbeauvais: Third, if Change Of Company Information is a valid filing that we want to correct, then you'll need to update some code in FilingHistoryList.vue.

Note that the BE isn't implemented to support this type of correct at the moment: https://github.com/bcgov/business-schemas/blob/675fe5de5a4d7584a88eb1e8ddfef9889c5da780/src/registry_schemas/schemas/correction.json#L29

I'm thinking that handling this type of filing, and its correction, is future work. Check with Mihai.

severinbeauvais: Note that the BE isn't implemented to support this type of correct at the moment: https://github.com/bcgov/business-schemas/blob/675fe5de5a4d7584a88eb1e8ddfef9889c5da780/src/registry_schemas/schemas/correction.json#L29

I'm thinking that handling this type of filing, and its correction, is future work. Check with Mihai.

lambert-alex: So the issue I was trying to correct, is that when a correction is saved it is actually saved as a CHANGE_OF_COMAPNY_INFO file type and not as an alteration so when the filing is reloaded from a save it isn't considered a valid filing for a edit ui correction. I drilled down a little more and I think it might actually be an issue with how the filing is saved. I'll look into this more.

severinbeauvais: A correction should be saved as a "correction".

src/components/Dashboard/TodoList.vue
lambert-alex: Remove all accidental console.logs

src/components/Dashboard/TodoList.vue
```diff
@@ -1385,8 +1386,9 @@ export default class TodoList extends Vue {
       case FilingTypes.CORRECTION:
         // see also FilingHistoryList.vue:correctThisFiling()
         switch (item.correctedFilingType) {
-          case FilingTypes.ALTERATION:
```
lambert-alex, line 1388: This was the only case using FilingTypes instead of FilingNames leading to an bug where and 'alteration' would be checked against an 'Alteration'

src/components/Dashboard/TodoList.vue
```diff
@@ -1385,8 +1386,9 @@ export default class TodoList extends Vue {
...
+          case FilingNames.ALTERATION:
           case FilingNames.INCORPORATION_APPLICATION:
+          case FilingNames.CHANGE_OF_COMPANY_INFO:
```
severinbeauvais, line 1391: You shouldn't include this kind of filing here, right?

severinbeauvais:
Oops, your rebase got rid of the app version updates.

src/components/Dashboard/TodoList.vue
```diff
@@ -1081,9 +1081,10 @@ export default class TodoList extends Vue {
         // this is only used for internal corrections (not IA):
         correctedFilingId: correction.correctedFilingId,
         // this is only used for external corrections (IA):
-        correctedFilingType: this.filingTypeToName(correction.correctedFilingType as FilingTypes),
+        // Corrections on Alterations always have a correctedFilingType of Alteration -> alterationRequired = true
+        correctedFilingType: this.filingTypeToName(correction.correctedFilingType as FilingTypes, null, true),
```
lambert-alex, line 1085: So [`filingTypeToName`](https://github.com/bcgov/business-filings-ui/blob/288cf9465fa3fa6a03e4583b306bc153fe22f1b7/src/mixins/enum-mixin.ts#L229-L230) would set the correctedFilingType as a `CHANGE_OF_COMPANY_INFO` as `alterationRequried` defaults to false. Then [`doResumeFiling`](https://github.com/bcgov/business-filings-ui/blob/288cf9465fa3fa6a03e4583b306bc153fe22f1b7/src/components/Dashboard/TodoList.vue#L1385-L1408) would resume the fling based on the `correctedFilingType` and do a default redirect to a filings UI correction.

tests/unit/TodoList1.spec.ts
```diff
@@ -2321,52 +2321,115 @@ describe('TodoList - Click Tests - IA Corrections', () => {
...
-            header: {
-              name: 'correction',
-              status: 'DRAFT',
-              filingId: 123,
-              comments: []
-            },
-            business: {},
-            correction: {
-              correctedFilingType: 'incorporationApplication'
+  const editTests = [
```
lambert-alex, line 2324: Updated the test to check against all filing types that redirect to the edit ui

tests/unit/TodoList1.spec.ts
```diff
@@ -2321,52 +2321,115 @@ describe('TodoList - Click Tests - IA Corrections', () => {
...
+      // verify redirection
+      const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id
+      const editUrl = `https://edit.url/${test.businessId}/correction/?correction-id=123`
+      expect(window.location.assign).toHaveBeenCalledWith(editUrl + '&accountid=' + accountId)
+
+      wrapper.destroy()
+    })
+  }
+
+  const localTest = ['annualReport', 'changeOfAddress', 'changeOfDirectors', 'conversion']
```
lambert-alex, line 2381: Added tests to check all the correction types that stay in the filings ui

src/components/Dashboard/TodoList.vue
lambert-alex: The description below should resolve this issue.

severinbeauvais (approved):
Looks good.

tests/unit/TodoList1.spec.ts
lambert-alex: Do the reviewers think that there is any value added to testing the default redirection?

severinbeauvais: I think so. That way, a test breaks if we do something different with these filing types.

PR #429, opened 2022-12-06
14516 Fixed pending correct status + Cancel Payment style

severinbeauvais:
*Issue #:* bcgov/entity#14516

*Description of changes:*
- added check for new pending status
- added logic to display "processing..." while checking status
- reduced retry count from 10 to 5 to finish faster
- updated list item button hover colour
- added new filing status enum value
- added new  mixin method + unit test
- app version = 5.8.1

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -130,9 +130,10 @@
                 </div>
 
                 <!-- pending filing - correction or alteration, or payment incomplete -->
-                <div v-else-if="isStatusPending(item)" class="todo-subtitle">
+                <div v-else-if="(isStatusPending(item) || isStatusPendingCorrection(item))" class="todo-subtitle">
                   <template v-if="isTypeCorrection(item) || isTypeAlteration(item)">
-                    <span>FILING PENDING</span>
+                    <span v-if="inProcessFiling === item.filingId">PROCESSING...</span>
+                    <span v-else>FILING PENDING</span>
                   </template>
 
                   <template v-else>
@@ -357,7 +358,7 @@
                     <v-list class="actions__more-actions">
                       <v-list-item
                         id="btn-cancel-payment"
-                        @click.native.stop="confirmCancelPayment(item)"
+                        @click.stop="confirmCancelPayment(item)"
                       >
                         <v-list-item-title>Cancel Payment</v-list-item-title>
                       </v-list-item>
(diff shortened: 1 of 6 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -130,9 +130,10 @@
                 </div>
 
                 <!-- pending filing - correction or alteration, or payment incomplete -->
-                <div v-else-if="isStatusPending(item)" class="todo-subtitle">
+                <div v-else-if="(isStatusPending(item) || isStatusPendingCorrection(item))" class="todo-subtitle">
```
severinbeauvais, line 133: We didn't know about this status originally. Now we handle it correctly.

src/components/Dashboard/TodoList.vue
```diff
@@ -130,9 +130,10 @@
...
                   <template v-if="isTypeCorrection(item) || isTypeAlteration(item)">
-                    <span>FILING PENDING</span>
+                    <span v-if="inProcessFiling === item.filingId">PROCESSING...</span>
+                    <span v-else>FILING PENDING</span>
```
severinbeauvais, line 136: Like the other instances of filings where we check their status a few times a second apart, this one should say "processing..." while we check and "filing pending" after.

src/components/Dashboard/TodoList.vue
```diff
@@ -728,11 +730,11 @@ export default class TodoList extends Vue {
...
   /**
    * Checks whether the subject filing is now Completed.
-   * Retries after 1 second for up to 10 iterations.
+   * Retries after 1 second for up to 5 iterations.
    */
   private checkIfCompleted (id: number, count: number): void {
-    // stop this cycle after 10 iterations
-    if (++count > 10) {
+    // stop this cycle after 5 iterations
+    if (++count > 5) {
```
severinbeauvais, line 737: Ten iterations should be about 10 seconds... except when the API takes 5 seconds to return from each "GET filing" call, so this mitigates that.

src/components/Dashboard/TodoList.vue
```diff
@@ -1756,7 +1758,9 @@ export default class TodoList extends Vue {
     color: $app-blue;
   }
 
-  #btn-draft-delete:hover {
+  #btn-draft-delete:hover,
+  #btn-delete-application:hover,
+  #btn-cancel-payment:hover {
     background-color: $gray1;
```
severinbeauvais, line 1764: Hover styling was missing altogether from the Cancel Payment button.

The Delete Application button had default Vuetify styling, which is a very light gray. Now it's darker gray and consistent.

lewischenstudio (approved):

lambert-alex (commented):
LGTM

PR #433, opened 2022-12-14
14405 Changed API mixins to services before further work

severinbeauvais:
*Issue #:* bcgov/entity#14405

*Description of changes:*
- refactored AuthApiMixin -> AuthServices
- refactored LegalApiMixin -> LegalServices
- refactored PayApiMixin -> PayServices
- updated unit tests
- app version = 5.8.5


src/services/auth-services.ts
```diff
@@ -0,0 +1,37 @@
+// Libraries
+import axios from '@/axios-auth'
+
+/**
+ * Class that provides integration with the Auth API.
+ */
+export default class AuthServices {
```
severinbeauvais, line 7: Changing these from mixins to classes:
- is cleaner since the methods have their own namespace instead of polluting `this`
- is easier to test
- is consistent with our newer UIs

ChrisGabel23 (approved):

src/services/auth-services.ts
ketaki-deodhar: Looks much cleaner!

ketaki-deodhar (approved):

src/services/auth-services.ts
lambert-alex: As part of these future changes should we start trying to make the mixins as loosely coupled as possible with the Components that use them to prepare to change to the Compositon API?

lambert-alex (approved):
LGTM

src/services/auth-services.ts
severinbeauvais: Yes, this change is heading in that direction, right?

lambert-alex: I think so, anything the creates greater distinction between the service/mixing and the component helps.

PR #434, opened 2022-12-14
14733 Set default business standing to true so IAs don't show as not good standing

lambert-alex:
*Issue #:* /bcgov/entity#14733

*Description of changes:*
 - Default state for business in good standing is now true
 - App version 5.8.5 -> 5.8.6
 - Add unit testing for dashboard alert
 - The back end issue is being fixed as part of https://github.com/bcgov/lear/pull/1860


lambert-alex:
IA do not show the alert now

Historical businesses still show as not in good standing as that error is in the BE

Businesses that require alerts still show the alert display

src/store/state.ts
```diff
@@ -20,7 +20,7 @@ export const state: StateIF = {
   entityState: null,
   entityStatus: null,
   entityFoundingDate: null,
-  goodStanding: null,
+  goodStanding: true,
```
lambert-alex, line 23: the default state would be loaded for IAs as there was no business and null evaluated to false in the check for inGoodStanding

tests/unit/Dashboard.spec.ts
```diff
@@ -34,7 +34,23 @@ describe('Dashboard - UI', () => {
...
+      data: () => ({
+        isInGoodStanding: true
+      }),
+      computed: {
+        isGoodStanding: {
+          get (): boolean {
+            return this.$data.isInGoodStanding
+          },
+          set (val: boolean) {
+            this.$data.isInGoodStanding = val
```
lambert-alex, line 49: Setting local data in the wrapper allows us to mock getters in jest. isGoodStanding in Dashboard is overridden and and the wrapper will rely on the local computed version. This allows us to change the getter for tests as done below.

src/store/state.ts
severinbeauvais: I think the store should be "empty" initially (or this can have future side-effects when something is assumed incorrectly).

How about setting Good Standing here (with a suitable comment) instead?
https://github.com/bcgov/business-filings-ui/blob/d68faa5c447162731dd9f8abfb8b763d2dd4d8fa/src/App.vue#L600

lambert-alex: Done, good catch @ChrisGabel23

src/App.vue
```diff
@@ -597,6 +597,9 @@ export default {
       this.setIdentifier(this.tempRegNumber)
       this.setEntityType(legalType)
 
+      // NameRequests are always in good standing
```
severinbeauvais, line 600: This isn't a Name Request. It's a draft application.

severinbeauvais (approved):

src/store/state.ts
severinbeauvais: LGTM now.

src/App.vue
lambert-alex: Oops fixed!

ChrisGabel23 (approved):

PR #439, opened 2023-01-07
14637 Show draft restoration in Todo list

severinbeauvais:
*Issue #:* bcgov/entity#14637

**This is only the first part of the ticket, ie, incremental PRs.**

*Description of changes:*
- added code to display restoration task (staff and non staff)
- added code to resume draft restoration
- added a Filing Name to enum
- added a Filing Type to enum
- added Restoration Types enum
- added Restoration IF
- added isTypeRestoration to mixin
- imported latest shared enums, etc
- app version = 5.9.2


package.json
```diff
@@ -16,14 +16,14 @@
...
-    "@bcrs-shared-components/court-order-poa": "2.1.2",
-    "@bcrs-shared-components/enums": "1.0.25",
-    "@bcrs-shared-components/interfaces": "1.0.44",
-    "@bcrs-shared-components/staff-comments": "1.3.2",
-    "@bcrs-shared-components/staff-payment": "2.1.2",
+    "@bcrs-shared-components/court-order-poa": "2.1.4",
+    "@bcrs-shared-components/enums": "1.0.26",
+    "@bcrs-shared-components/interfaces": "1.0.47",
+    "@bcrs-shared-components/staff-comments": "1.3.4",
+    "@bcrs-shared-components/staff-payment": "2.1.4",
```
severinbeauvais, line 26: I needed the latest enums and the others has no changes (only version update).

src/interfaces/restoration-interface.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * A filing's restoration object from the API. See:
+ * https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/restoration.json
+ */
+export interface RestorationIF {
+  date: string // *** TODO: format?
+  type: RestorationTypes
+  expiry?: string // *** TODO: format?
+}
```
severinbeauvais, line 11: This will need to be updated by future tickets (that implement restoration components/data) and/or schema changes.

severinbeauvais: Also I'm not sure about the date formats yet -- probably `YYYY-MM-DDThh:mm:ss+00:00` (API format) but maybe just `YYYY-MM-DD`. To be determined by future UI/API tickets.

src/components/Dashboard/TodoList.vue
```diff
@@ -1269,6 +1272,45 @@ export default class TodoList extends Vue {
...
+
+    if (header && restoration) {
+      // don't allow resuming a draft if not in good standing
+      if (this.isStatusDraft(header) && !this.isGoodStanding) {
+        task.enabled = false
+      }
+
+      const corpTypeDescription = this.getCorpTypeDescription(business.legalType)
+
+      const title = this.filingTypeToName(FilingTypes.RESTORATION, null, restoration.type)
```
severinbeauvais, line 1289: Filing name depends on restoration type (full or limited).

severinbeauvais:
Sample restoration draft as staff user:

image

Sample restoration draft as non-staff user:

image

src/interfaces/restoration-interface.ts
ChrisGabel23: Unused imports (from TodoList.vue) and TODOs are flagged as Code Smells by SonarCloud. Are we okay with adding three new smells to the code, or is there a better way to do this? (like saying "future use" instead of "TODO")

severinbeauvais: Good catch. It's better to not have code smells.

As for TODO vs FUTURE, my style is to use TODO for things I want to complete in the current PR, and FUTURE for things that should be completed in future work. YMMV. This feature is still under development so I'm not too worried about this right now.

I've update the code. Please have a look at the latest commit.

ChrisGabel23: Sounds good!

ChrisGabel23 (approved):

jonathan-longe (approved):
@severinbeauvais thanks for the opportunity to get familiar with this code.  LGTM

PR #448, opened 2023-01-26
14808 - show full and limited restorations

jonathan-longe:
*Issue #:* /bcgov/entity#14808

Add limited restoration component to the dashboard's filing history.  Confirm that full restoration display correctly.

image

image

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -86,8 +86,9 @@
                 <!-- or a completed Alteration? -->
                 <!-- or a completed Dissolution? -->
                 <!-- or a completed Registration? -->
+                <!-- or a limited Restoration? -->
                 <div v-else-if="filing.isCompletedIa || filing.isCompletedAlteration || filing.isCompletedDissolution ||
-                  filing.isCompletedRegistration" class="item-header__subtitle"
+                  filing.isCompletedRegistration || filing.isTypeLimitedRestoration" class="item-header__subtitle"
                 >
                   <span>FILED AND PAID <FiledLabel :filing="filing" /></span>
                   <v-btn
@@ -300,6 +301,11 @@
               <FutureEffective :filing=filing class="mt-6" />
             </template>
 
+            <template v-else-if="filing.isTypeLimitedRestoration"
+            >
+              <LimitedRestorationFiling :filing="filing" class="mt-6" />
+            </template>
+
             <!-- is this a generic paid (not yet completed) filing? -->
             <template v-else-if="isStatusPaid(filing)">
               <PendingFiling :filing=filing class="mt-6" />
@@ -381,9 +387,11 @@ import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryIt
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin } from '@/mixins'
 import { LegalServices } from '@/services/'
+import LimitedRestorationFiling from '@/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue'
 
 @Component({
   components: {
+    LimitedRestorationFiling,
     // sub-components
     CompletedAlteration,
     CompletedDissolution,
@@ -650,6 +658,13 @@ export default class FilingHistoryList extends Vue {
         item.putBackOnOrAdminDissolution = this.isTypePutBackOn({ name: filing.name }) ||
           this.isTypeAdministrativeDissolution({ name: filing.name,
             dissolutionType: filing?.data?.dissolution?.dissolutionType })
+      }
+
+      // add properties for limited restorations
+      if (this.isTypeLimitedRestoration(filing)) {
+        item.isTypeLimitedRestoration = true
+        item.legalName = filing?.data?.restoration?.legalName
+        item.expiryDate = filing?.data?.restoration?.expiryDate
       }
 
       this.historyItems.push(item)
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -664,7 +664,7 @@ export default class FilingHistoryList extends Vue {
       if (this.isTypeLimitedRestoration(filing)) {
         item.isTypeLimitedRestoration = true
         item.legalName = filing.data.restoration?.legalName
-        item.expiryDate = filing.data.restoration?.expiryDate
+        item.expiry = filing.data.restoration?.expiry
       }
 
       this.historyItems.push(item)
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -378,6 +378,7 @@ import DocumentsList from './FilingHistoryList/DocumentsList.vue'
 import FiledLabel from './FilingHistoryList/FiledLabel.vue'
 import FutureEffective from './FilingHistoryList/FutureEffective.vue'
 import FutureEffectivePending from './FilingHistoryList/FutureEffectivePending.vue'
+import LimitedRestorationFiling from '@/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue'
 import PaperFiling from './FilingHistoryList/PaperFiling.vue'
 import PendingFiling from './FilingHistoryList/PendingFiling.vue'
 import StaffFiling from './FilingHistoryList/StaffFiling.vue'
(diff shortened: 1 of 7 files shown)


ChrisGabel23 (approved):
Don't forget to up the version number.

jonathan-longe:
> Don't forget to up the version number.

Thanks for the reminder @ChrisGabel23  I'm on it!

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -86,8 +86,9 @@
                 <!-- or a completed Alteration? -->
                 <!-- or a completed Dissolution? -->
                 <!-- or a completed Registration? -->
+                <!-- or a limited Restoration? -->
                 <div v-else-if="filing.isCompletedIa || filing.isCompletedAlteration || filing.isCompletedDissolution ||
-                  filing.isCompletedRegistration" class="item-header__subtitle"
+                  filing.isCompletedRegistration || filing.isTypeLimitedRestoration" class="item-header__subtitle"
```
severinbeauvais, line 91: Why wouldn't this automatically use the "else this must be a completed filing" section on line 167?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -300,6 +301,11 @@
               <FutureEffective :filing=filing class="mt-6" />
             </template>
 
+            <template v-else-if="filing.isTypeLimitedRestoration"
+            >
+              <LimitedRestorationFiling :filing="filing" class="mt-6" />
+            </template>
```
severinbeauvais, line 307: This looks good.

The rule for where to put this (since order matters in this long if-then-else) is "as low as possible" so it doesn't pre-empt something it shouldn't but gets pre-empted by something more important.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -381,9 +387,11 @@ import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryIt
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin } from '@/mixins'
 import { LegalServices } from '@/services/'
+import LimitedRestorationFiling from '@/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue'
```
severinbeauvais, line 390: I recommend you move this up where the other sub-components are imported (and you can follow the same file path spec).

Also, please move line 394 down to keep the list alphabetized.

JazzarKarim (approved):

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -652,6 +660,13 @@ export default class FilingHistoryList extends Vue {
             dissolutionType: filing?.data?.dissolution?.dissolutionType })
       }
 
+      // add properties for limited restorations
+      if (this.isTypeLimitedRestoration(filing)) {
+        item.isTypeLimitedRestoration = true
+        item.legalName = filing?.data?.restoration?.legalName
+        item.expiryDate = filing?.data?.restoration?.expiryDate
+      }
```
severinbeauvais, line 668: The Typescript [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator is a double-edged sword -- it can mask errors that you should be handling (in previous code).

Here, you don't have to check that filing is non-null/undefined because that was checked previously, so it's guaranteed to be OK.

I looked at other code in here, and it doesn't use ? on data either, though maybe it should.

Maybe a better approach here, for this type of filing, is to verify that filing.data.restoration is non-null/undefined (as that would be a real error) and throw an exception otherwise (which would get caught by line 671).

src/mixins/enum-mixin.ts
```diff
@@ -176,6 +176,11 @@ export default class EnumMixin extends Vue {
     return (item.name === FilingTypes.SPECIAL_RESOLUTION)
   }
 
+  /** Returns True if filing is a Limited Restoration. */
+  isTypeLimitedRestoration (item: any): boolean {
+    return (item.name === FilingTypes.RESTORATION && item.data?.restoration?.type === 'limitedRestoration')
```
severinbeauvais, line 181: I agree with the optional chaining here, since this might not be a restoration item.

You should use an enum for the type -- see RestorationTypes.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+  <div v-if="filing">
+    <h4>Limited Restoration Period</h4>
+
+    <p class="limited-restoration-period">
+      The Company <strong>{{ filing.legalName }}</strong> was successfully
+      restored and active <strong>until {{ expiryDateFriendly }}</strong>
```
severinbeauvais, line 7: Should be "is active" ?

Should have period at end of sentence?

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.limited-restoration-period {
+  color: dimgrey;
```
severinbeauvais, line 44: I believe we don't use this colour. Check theme.scss for a suitable grayX colour variable (same as other sub-component in FilingHistoryList).

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
```diff
@@ -0,0 +1,47 @@
...
+  font-size: $px-15;
```
severinbeauvais, line 45: We're not super-consistent with this but in some of our UIs we have a class that sets this.

Search for `.font-15`.

severinbeauvais (commented):
Looks really good. Thanks for the unit tests. Just a couple minor comments.

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Unfortunately, we can't use the "else" block because we need the "View / Hide details" button on lines 94 to 103.

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Thanks!

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Ah, good call.  I'll fix it.

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Good call.  I've changed it as you suggest.

src/mixins/enum-mixin.ts
jonathan-longe: Good catch.  I've fixed it.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
jonathan-longe: Good catch on the missing period.  I didn't add "is" because that's not what was requested, but I'm happy to add it anytime.
image

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
jonathan-longe: Thanks for the info.  I've fixed it.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
jonathan-longe: Hmm, I couldn't find another FilingHistory sub-component using `$gray#`, but `$gray7` seems like a close match to what the UI requested.  Let me know if you'd prefer something different.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: OK, perfect.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
severinbeauvais: Double-check with Ethan. To me, it doesn't read right otherwise.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
severinbeauvais: Sounds good to me.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
jonathan-longe: @ethantspitt can you weigh in on this thread?  I'm happy to change it however you wish.

ethantspitt: Thanks for checking! let's use "is active"

severinbeauvais (approved):
Looks great!

Feel free to merge after last UI change.

src/components/Dashboard/FilingHistoryList/LimitedRestorationFiling.vue
jonathan-longe: Thanks Ethan.  Will do.

ketaki-deodhar (approved):

PR #450, opened 2023-02-02
14979 add items to staff filing menu

jonathan-longe:
*Issue #:* /bcgov/entity#14979

Added the following two menu items to the staff menu with `@click` functionality to build appropriate restoration filing, save to API and redirect to edit-ui:
- Extend Limited Restoration
- Convert to Full Restoration

This PR also undertakes some refactoring to place the API responses from state filings in the store and use getters to reference the values stored there.


src/App.vue
```diff
@@ -522,46 +522,15 @@ export default {
       this.setGoodStanding(business.goodStanding)
       this.setHasCourtOrders(business.hasCourtOrders)
 
-      if (business.state === EntityState.HISTORICAL) {
-        await this.parseStateFiling(business.stateFiling) // will throw on error
+      // retrieve stateFiling if the URL is included in the business payload
+      if (business.stateFiling) {
```
jonathan-longe, line 526: Retrieve the state filing if it exists instead of only in the case of a historical entity

src/App.vue
```diff
@@ -522,46 +522,15 @@ export default {
...
-        name = this.filingTypeToName(filingType)
-        const effectiveDate = this.apiToDate(filing.header?.effectiveDate)
-        if (!effectiveDate) throw new Error('Invalid effective date')
-        date = this.dateToPacificDateTime(effectiveDate)
-      }
-
-      const enDash = '–' // ALT + 0150
-      this.setReasonText(`${name} ${enDash} ${date}`)
-    },
-
```
jonathan-longe, line 564: Clean up App.vue by moving this logic to a getter

src/components/Dashboard/StaffNotation.vue
```diff
@@ -245,20 +271,27 @@ export default class StaffNotation extends Vue {
     this.close(needReload)
   }
 
-  async goToRestorationFiling (): Promise<void> {
+  async goToRestorationFiling (applicationName: ApplicationTypes, restorationType: RestorationTypes): Promise<void> {
+    let url: string
```
jonathan-longe, line 275: Extending the use of this method by passing in parameters

src/interfaces/state-interface.ts
```diff
@@ -55,5 +65,12 @@ export interface StateIF {
...
   hasBlockerFiling: boolean
   isCoaPending: boolean
-  coaEffectiveDate: Date
+  coaEffectiveDate: Date,
+  stateFiling: {
+    business: any,
+    header: ApiHeaderIF,
+    dissolution?: any,
+    restoration?: any
+    // any others?
```
jonathan-longe, line 74: @severinbeauvais I'm not sure I'm aware of all the stateFiling types there could be

src/mixins/enum-mixin.ts
```diff
@@ -239,68 +227,35 @@ export default class EnumMixin extends Vue {
...
-    switch (type) {
-      case DissolutionTypes.VOLUNTARY:
-        return this.isFirm ? DissolutionNames.FIRM_DISSOLUTION : DissolutionNames.VOLUNTARY
-      case DissolutionTypes.ADMINISTRATIVE:
-        return DissolutionNames.ADMINISTRATIVE
-    }
-    // fallback for unknown filings
-    return this.camelCaseToWords(type)
+    return EnumUtilities.dissolutionTypeToName(this.isFirm, type)
   }
```
jonathan-longe, line 260: As we discussed with @severinbeauvais, mixins are not supported in Vue3 and, more importantly, I can't call them from getters.  I've moved the logic to a new class using all static methods and converted this mixin to an adapter.

src/interfaces/state-interface.ts
severinbeauvais: Put Back On.

I think that's it, but double check with Mihai or Argus.

src/App.vue
```diff
@@ -92,7 +92,7 @@
 </template>
 
 <script lang="ts">
-import { mapActions, mapGetters } from 'vuex'
+import { mapActions, mapGetters, mapMutations } from 'vuex'
```
severinbeauvais, line 95: I think you don't need `mapMutations`.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -154,15 +174,26 @@
...
+    ...mapGetters([
+      'getIdentifier',
+      'isAdminFreeze',
+      'isBenBcCccUlc',
+      'isCoop',
+      'isEntityInLimitedRestoration',
+      'isFirm',
+      'isHistorical'
+    ])
+  },
```
severinbeauvais, line 196: I'm OK with this format (instead of using `@Getter` decorators. I think this is more in line with [Vue3 Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) anyway.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -118,7 +118,7 @@
             <template v-if="isBenBcCccUlc && isHistorical">
               <v-list-item
                 data-type="restoration"
-                @click="goToRestorationFiling()"
+                @click="goToRestorationFiling(ApplicationTypes.CREATE_UI, RestorationTypes.FULL)"
```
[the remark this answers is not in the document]
jonathan-longe: That's true.  The behavior of `goToRestorationFilling()` hasn't changed for create-ui.  I've simply extended the method for use by edit-ui too.

src/App.vue
jonathan-longe: Good call, I'll remove it.

src/interfaces/restoration-interfaces.ts
```diff
@@ -5,9 +5,9 @@ import { RestorationTypes } from '@/enums'
  * https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/restoration.json
  */
 export interface RestorationIF {
-  date: string // FUTURE: describe format here
   type: RestorationTypes
   expiry?: string // FUTURE: describe format here
+  legalName?: string
```
severinbeauvais, line 10: What is this for? I don't think it's in the schema.

src/interfaces/index.ts
```diff
@@ -43,3 +43,6 @@ export * from './dissolution-confirmation-interface'
 export * from './todo-list-interface'
 export * from './steps-interface'
 export * from './table-header-interface'
+
+export interface StateFilingIF {
+}
```
severinbeauvais, line 48: What is this for?

src/mixins/date-mixin.ts
```diff
@@ -1,360 +1,206 @@
...
-const MS_IN_A_DAY = (1000 * 60 * 60 * 24)
+import DateUtilities from '@/resources/date-utilities'
 
 /** Mixin that provides some useful date utilities. */
 @Component({})
 export default class DateMixin extends Vue {
   @Getter getCurrentJsDate!: Date
 
   /**
+   * DEPRECATED - call resources/date-utilities instead
```
severinbeauvais, line 13: Please create a Tech Debt ticket so we can go in and update the code that uses these deprecated methods.

Also, FYI, this Date Mixin is also used in Create UI and Edit UI, and there's even a version in BCRS Shared Components. They're all a bit different, though I'd like to have a single version (ideally, shared) that all can simply import... or whatever that looks like using Date Utilities instead.

src/interfaces/restoration-interfaces.ts
jonathan-longe: That's true, it's not in the schema, but the business has requested the legalName as at the time of restoration.  Let's discuss this in a call.

src/mixins/filing-mixin.ts
```diff
@@ -160,8 +165,9 @@ export default class FilingMixin extends DateMixin {
    * Builds a Restoration filing body.
    * @returns the filing body
    */
-  buildRestorationFiling (): RestorationFilingIF {
-    const restorationFiling: RestorationFilingIF = {
+  buildRestorationFiling (restorationType: RestorationTypes = RestorationTypes.FULL): RestorationFilingIF {
```
severinbeauvais, line 168: Some linters (or maybe our linter with specific options) will complain about "redundant typing" ... it can infer that `restorationType` is of type RestorationTypes by the default value you're assigning.

Similarly, this can cause warnings:
`const myBool: boolean = true`

So, as a general rule, I don't specific the type when it can be inferred.

(Background: a while back I experimented with bringing this code base into Vue3, which brought along extra linting features. Fixing such things now should make less work in the future.)

jonathan-longe: Ah, good point.  I'll make sure I watch for that moving forward.

src/resources/date-utilities.ts
```diff
@@ -0,0 +1,353 @@
...
+const MS_IN_A_DAY = (1000 * 60 * 60 * 24)
+
+export default class DateUtilities {
+  /**
+   * Fetches and returns the web server's current date (in UTC).
+   * Used to bypass the user's local clock/timezone.
+   * Ref: https://www.npmjs.com/package/serverdate
+   * @returns a promise to return a Date object
+   */
+  static async getServerDate (): Promise<Date> {
```
severinbeauvais, line 13: FYI, I want to eventually replace this with an external library, or our web worker, or whatever, to:
- reduce the number of network calls
- simplify the app code

All the while having "date and time synced to a known good reference" available in this UI. The reason is that we can't (shouldn't) trust the browser's date-time. For some things, we need time accuracy down to about a minute.

src/resources/date-utilities.ts
```diff
@@ -0,0 +1,353 @@
+import { isDate } from 'lodash'
```
severinbeauvais, line 1: I'd rather have this in a new folder. We've been using "resources" for resource files.

src/mixins/filing-mixin.ts
```diff
@@ -160,8 +165,9 @@ export default class FilingMixin extends DateMixin {
...
+    // For security reasons, restoration.expiry and restoration.legalName should be populated by the API
```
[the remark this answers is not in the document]
jonathan-longe: Thanks @severinbeauvais.  I'll remove the comment.  For the purposes of creating a draft, do we even need to submit a payload to the API?  It seems to me most (all?) the values we're submitting should be overridden by the API.

src/resources/date-utilities.ts
severinbeauvais: Would it make sense to call these files "services" and put them in the services folder?

jonathan-longe: Yes, I wondered about the correct directory to place these files.  Any objection if I create a new directory called `utilities`?

src/App.vue
```diff
@@ -522,46 +522,15 @@ export default {
       this.setGoodStanding(business.goodStanding)
       this.setHasCourtOrders(business.hasCourtOrders)
 
-      if (business.state === EntityState.HISTORICAL) {
-        await this.parseStateFiling(business.stateFiling) // will throw on error
+      // retrieve stateFiling if the URL is included in the business payload
+      if (business.stateFiling) {
+        await this.retrieveStateFiling(business.stateFiling)
```
severinbeauvais, line 527: I see errors are caught by the grand-parent of the current method. Is that the plan? Is it tested?

src/store/actions.ts
```diff
@@ -125,5 +124,19 @@ export default {
...
   setCoaEffectiveDate ({ commit }, coaEffectiveDate: Date) {
     commit('coaEffectiveDate', coaEffectiveDate)
+  },
+  /** Fetches stateFiling from the API and, if successful, triggers mutation. */
+  async retrieveStateFiling ({ commit }, stateFilingUrl: string): Promise<void> {
+    const filing = stateFilingUrl && await LegalServices.fetchFiling(stateFilingUrl)
+    const filingType = filing?.header?.name as FilingTypes
+    return new Promise((resolve, reject) => {
+      if (!filing || !filingType) {
+        console.log(filing, filingType)
```
severinbeauvais, line 134: I'm not sure if this console.log is required (since the rejection has a decent message already) though it may be useful to log the bad value.

If this console.log should remain then please add useful text in it so (a) user has a message of what went wrong, and (b) we can globally search the code for this message :)

Also, you should include `eslint-disable-line no-console` for log messages to avoid linter warnings/errors.

src/resources/date-utilities.ts
jonathan-longe: I'm not sure I understand the reason for getting the dateTime from the server.  If there's a filing that must be submitted before midnight.  We make the deadline clear on the form and then, on submission, validate the submission time as part of the business logic.  If the form is submitted past the deadline, we reject the submission and return a specific error message from the API.  I think I'm missing some aspect of this.  Does the complexity lie when handling submissions from parts of the province on mountain time rather than pacific?

src/store/getters.ts
```diff
@@ -283,5 +285,36 @@ export default {
...
+      date = DateUtilities.dateToPacificDateTime(effectiveDate)
+    }
+
+    const enDash = '–' // ALT + 0150
+    return `${name} ${enDash} ${date}`
+  },
+
+  /** */
+  isEntityInLimitedRestoration (state: StateIF): boolean {
+    return (state.stateFiling?.business?.state === 'ACTIVE' && state.stateFiling.hasOwnProperty('restoration'))
```
severinbeauvais, line 318: It may be the purview of a new ticket, but I'm thinking we should ideally/eventually keep "pure getters" (which return store data) with these "computed getters". Thoughts?

tests/unit/actions.spec.ts
```diff
@@ -0,0 +1,118 @@
+import LegalServices from '@/services/legal-services'
```
severinbeauvais, line 1: Please capitalize this file's name (consistent with class it's testing).

severinbeauvais: Wait -- I see this is testing store actions, not Actions.vue.

How about renaming it "store-actions" ?

src/store/getters.ts
jonathan-longe: Hmm, let me think about that.  My gut reaction is that I don't see the benefit of segregating getters in this way.  For example, if I'm calling a getter called `isDisplaySomeRandomComponent()`, I shouldn't have wonder if it's a pure getter or a computed getter.  How it works should be entirely hidden from me.

tests/unit/StaffNotation.spec.ts
```diff
@@ -45,48 +46,194 @@ describe('StaffNotation', () => {
     wrapper.destroy()
   })
 
-  it('renders drop-down menu correctly - active', async () => {
+  it('renders drop-down menu correctly - active and not limited restoration', async () => {
     // set store specifically for this test
-    store.state.entityType = 'SP'
-    store.state.entityState = 'ACTIVE'
+    store.state.entityType = CorpTypeCd.COOP
+    store.state.entityState = EntityState.ACTIVE
```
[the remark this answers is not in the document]
jonathan-longe: Good catch.  I saw your comment to Karim yesterday.  I'll fix it.

tests/unit/actions.spec.ts
jonathan-longe: Will do!

src/interfaces/restoration-interfaces.ts
severinbeauvais: OK. Need to chat with Vysakh as well (as he worked on the schema most recently).

src/mixins/filing-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Do you have an example of a draft in which we don't need to save info? What do you have in mind?

src/mixins/date-mixin.ts
severinbeauvais: The reason we have 3 repos is because they have different architectures, and to be able to work on them and deploy independently.

I am open to new ideas! How do you envision keeping the UIs separate in one repo?

src/resources/date-utilities.ts
severinbeauvais: First, all times are Pacific (Daylight or Standard)

Some examples:
- appointment / cease date for directors
- future effective date (and time)

The above need to know accurate "now" time and date.

src/resources/date-utilities.ts
severinbeauvais: Go for it.

Might want to rename "utils" folder though...?

src/store/getters.ts
severinbeauvais: Would the "store" just be one thing that we import/use getters from, regardless of where they're declared?

src/interfaces/index.ts
jonathan-longe: Agreed, it's not needed.  I'll remove it.

src/resources/date-utilities.ts
jonathan-longe: I like your idea of putting them in the existing "services" folder.

src/store/actions.ts
jonathan-longe: Good thinking. I've updated the line with your suggestions.

src/store/getters.ts
jonathan-longe: I like your idea of having a number of stores organized by the resource.  For example, filingHistory's getters, mutators and actions would be segregated from the ToDo's getters, setters and mutators.

severinbeauvais: I thought that was your idea? :)

Todo List vs FHL might not be a good example since they have few setters/actions, but yes to the general idea.

I'm still thinking along the lines of "pure" getters/actions (to access state) vs "complex/computed/rich" getters -- which might not need to be in the store at all? Dunno, depends on what they need to do, and I'm keeping my eyes open to what you're updating and proposing.

src/mixins/date-mixin.ts
jonathan-longe: There's definitely some complexity to this change, but I'd start by having a routes for edit-ui, create-ui and filings-ui.  Then simplify `App.vue` to call only `<router-view></router-view>`.  The "kick off logic" that currently resides in `App.vue` should be moved to actions and the component that the router file initially calls should trigger the appropriate action.

severinbeauvais: So you're thinking of combining our 3 UIs into 1 (huge) UI? I see more negatives than positives...
- positive: easier to share some common code
- positive: routing is much faster than redirecting
- negative: huge app to load (unless we can load parts on-demand)
- negative: increased risk/incidence of code conflicts due to several developers working on it
- negative: can't release 3 parts independently

I've seen chatter in StackOverflow about multiple UIs in the same repo (specifically so they could share code).

And there's always the shared repo that we could move more code to.

src/mixins/filing-mixin.ts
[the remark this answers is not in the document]
jonathan-longe: Sure, for example, it looks to me like most of the restoration create payload could be generated by the API since the `identifier` is a URL parameter and all other values (except the `restoration` object) could be derived from the identifier.
- header:
  - date: <don't trust what the client tells us>
  - name: <always restoration>
- business:
  - foundingDate: <this can be looked up -- right?>
  - identifier: <get from the url parameter>
  - legalName: <this can be looked up -- right?>
  - legalType: <this can be looked up -- right?>
- restoration:
  - type: <this is the only value that the front-end needs to submit>
Sending only a payload with:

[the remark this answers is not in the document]
severinbeauvais: You're mostly right and I agree with you.

(The header contains some fields from the UI: who certified, staff payment props, etc.)

src/mixins/date-mixin.ts
jonathan-longe: Thanks @severinbeauvais.  Perhaps I'm being too optimistic, but I see more positives than negatives.  By moving the kick-off-logic to actions that are triggered by specific components we can keep the amount of data that's loaded at startup to no more than present.  By separating concerns and being careful that tickets don't overlap we can avoid merge conflicts.  Let's keep the discussion going.  I SO appreciate you being open to discussing alternative ideas.

jonathan-longe: As requested, ticket created:  bcgov/entity#15234

src/store/getters.ts
jonathan-longe: Excellent, I'll continue to make small changes so we can discuss tangible examples rather than hypotheticals.  Again, I SO appreciate you being open to these discussions.

src/mixins/filing-mixin.ts
```diff
@@ -160,8 +165,9 @@ export default class FilingMixin extends DateMixin {
    * Builds a Restoration filing body.
    * @returns the filing body
    */
-  buildRestorationFiling (): RestorationFilingIF {
-    const restorationFiling: RestorationFilingIF = {
+  buildRestorationFiling (restorationType: RestorationTypes.FULL): RestorationFilingIF {
```
severinbeauvais, line 168: This should be an equal sign.

src/mixins/date-mixin.ts
severinbeauvais: Yes, it would be nice to have a code architecture where everything isn't happening at once :)

src/mixins/filing-mixin.ts
```diff
@@ -173,13 +179,8 @@ export default class FilingMixin extends DateMixin {
         legalType: this.getEntityType
       },
       restoration: {
-        // FUTURE: update as needed
-        date: null,
-        type: RestorationTypes.FULL,
-        expiry: null
+        type: restorationType
       }
     }
```
severinbeauvais, line 184: You may want to cast this anonymous object "as RestorationFilingIF" so any missing or mis-typed properties are identified.

src/store/getters.ts
```diff
@@ -1,8 +1,10 @@
-import { CorpTypeCd, EntityState, EntityStatus, FilingStatus, FilingTypes } from '@/enums'
+import {CorpTypeCd, DissolutionTypes, EntityState, EntityStatus, FilingStatus, FilingTypes} from '@/enums'
 import {
   ApiFilingIF, ApiTaskIF, DissolutionConfirmationResourceIF, OfficeAddressIF, PartyIF,
   StateIF, TodoListResourceIF, IsoDatePacific, BusinessWarningIF
 } from '@/interfaces'
+import DateUtilities from '@/services/date-utilities'
+import EnumUtilities from '@/services/enum-utilities'
```
severinbeauvais, line 7: Sometimes we have an index file in the folder, so you can import multiple things from the same source.

src/mixins/filing-mixin.ts
jonathan-longe: Hmm, the method already has a return type of RestorationFilingIF.  I'm not clear what you're asking.  Sorry, it's Friday afternoon, I may not be thinking clearly.
```
buildRestorationFiling (restorationType = RestorationTypes.FULL): RestorationFilingIF {
    return {
      header: {
        date: this.getCurrentDate,
        name: FilingTypes.RESTORATION
      },
      business: {
        foundingDate: this.dateToApi(this.getEntityFoundingDate),
        identifier: this.getIdentifier,
        legalName: this.entityName,
        legalType: this.getEntityType
      },
      restoration: {
        type: restorationType
      }
    }
  }
```

src/store/getters.ts
jonathan-longe: Ah, good thinking, I've fixed it.

src/mixins/filing-mixin.ts
severinbeauvais: The method is fine, but VS Code doesn't show errors in the anonymous object you're creating (as long as it's upward-compatible with the method return value). So it won't show missing fields. Try it.

severinbeauvais: I believe this doesn't complain:
```
  private formatAddress (address: any): AddressIF {
    return {
      severin: true,
      addressCity: address.addressCity || '',
      addressCountry: address.addressCountry || '',
      addressRegion: address.addressRegion || '',
      deliveryInstructions: address.deliveryInstructions || '',
      postalCode: address.postalCode || '',
      streetAddress: address.streetAddress || '',
      streetAddressAdditional: address.streetAddressAdditional || ''
    }
  }
```

But this does:
```
  private formatAddress (address: any): AddressIF {
    return {
      severin: true,
      addressCity: address.addressCity || '',
      addressCountry: address.addressCountry || '',
      addressRegion: address.addressRegion || '',
      deliveryInstructions: address.deliveryInstructions || '',
      postalCode: address.postalCode || '',
      streetAddress: address.streetAddress || '',
      streetAddressAdditional: address.streetAddressAdditional || ''
    } as AddressIF
  }
```

src/components/Dashboard/StaffNotation.vue
```diff
@@ -154,16 +174,38 @@
...
+    ...mapGetters([
+      'getIdentifier',
+      'isAdminFreeze',
+      'isBenBcCccUlc',
+      'isCoop',
+      'isEntityInLimitedRestoration',
+      'isFirm',
+      'isHistorical'
+    ])
+  },
```
severinbeauvais, line 208: Duplicate code.

src/mixins/filing-mixin.ts
jonathan-longe: Ah, got it.  Pushing update shortly

severinbeauvais: Yes, that works too :)

src/components/Dashboard/StaffNotation.vue
jonathan-longe: Removed

jonathan-longe: Not sure why Github isn't marking this as outdated, but it's been fixed.

severinbeauvais (approved):
Approved but please wait for a couple more reviews. You might need to poke people as they should know about the architectural changes you've made here.

JazzarKarim (approved):

src/App.vue
jonathan-longe: You are correct, the retrieveStateFiling() method sits inside a try / catch block.  If the method is unsuccessful, it throws an error and parent executes the catch block code.  I've been careful not to change the behavior.

PR #451, opened 2023-02-03
15042 Added Consent to Continuation Out to dashboard

severinbeauvais:
*Issue #:* bcgov/entity#15042

*Description of changes:*
    - added new route
    - added new enum mixin method
    - added item to staff notation list + route to new page
    - updated Filing History List to support new filing
    - updated Todo List to support new filing (staff and non-staff)
    - removed unused "haveData" from some filings
    - added filing code
    - added filing name
    - added filing type
    - added route name
    - added initial Consent Continuation Out page
    - deleted obsolete "changeOfName" object
    - added "consentContinuationOut" object
    - added properties to HistoryItemIF
    - cleaned up staff filing code
    - added handling for Consent to Continuation Out filings
    - misc cleanup
    - updated unit tests
    - app version = 5.10.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.9.11",
+  "version": "5.10.0",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -318,7 +318,9 @@
 
             <!-- is this a consent to continuation out filing? -->
             <template v-else-if="isTypeConsentContinuationOut(filing)">
-              <p class="mt-4">This consent is valid <strong>until {{filing.expiry}}</strong>.</p>
+              <p class="mt-4">
+                This consent is valid <strong>until {{filing.expiry}} at 12:01 am Pacific time</strong>.
+              </p>
               <p class="mt-4">{{filing.comment}}</p>
             </template>
 
@@ -652,8 +654,9 @@ export default class FilingHistoryList extends Vue {
 
       // add properties for Consent to Continuation Outs
       if (this.isTypeConsentContinuationOut(filing)) {
-        item.expiry = this.apiToPacificDateTime(filing.data.consentContinuationOut?.expiry)
-        item.comment = filing.data.consentContinuationOut?.comment
+        // FUTURE: expiry date may come from business object
+        item.expiry = filing.data.consentContinuationOut?.expiry
+        item.details = filing.data.consentContinuationOut?.orderDetails
       }
 
       // add properties for staff filings
@@ -662,9 +665,9 @@ export default class FilingHistoryList extends Vue {
         if (!this.isTypeCourtOrder(filing)) {
           item.documents = [] // no documents
         }
-        item.fileNumber = filing.data.order?.fileNumber || '' // may be falsy
+        item.fileNumber = filing.data.order?.fileNumber || '' // may be absent
         item.isTypeStaff = true
-        item.notationOrOrder = filing.data.order?.orderDetails // should not be falsy
+        item.details = filing.data.order?.orderDetails // should always be present
         item.planOfArrangement = filing.data.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
         item.putBackOnOrAdminDissolution = this.isTypePutBackOn({ name: filing.name }) ||
           this.isTypeAdministrativeDissolution({ name: filing.name,
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -55,7 +55,7 @@
               <!-- the filing label (left side) -->
               <div class="item-header__label">
                 <h3 class="item-header__title">
-                  <v-icon v-if="filing.displayName==FilingNames.COURT_ORDER" class="pr-1">mdi-gavel</v-icon>
+                  <v-icon v-if="isTypeCourtOrder(filing)" class="pr-1">mdi-gavel</v-icon>
                   <span>{{filing.displayName}}</span>
                 </h3>
 
@@ -253,6 +253,7 @@
               <StaffFiling
                 class="mt-6"
                 :filing="filing"
+                :showDocumentsList="isTypeCourtOrder(filing)"
                 @downloadOne="downloadOne(...arguments)"
                 @downloadAll="downloadAll($event)"
               />
@@ -301,8 +302,7 @@
               <FutureEffective :filing=filing class="mt-6" />
             </template>
 
-            <template v-else-if="filing.isTypeLimitedRestoration"
-            >
+            <template v-else-if="filing.isTypeLimitedRestoration">
               <LimitedRestorationFiling :filing="filing" class="mt-6" />
             </template>
 
@@ -314,6 +314,12 @@
             <!-- is this a paper filing? -->
             <template v-else-if="filing.availableOnPaperOnly">
               <PaperFiling class="mt-6" />
+            </template>
+
+            <!-- is this a consent to continuation out filing? -->
+            <template v-else-if="isTypeConsentContinuationOut(filing)">
+              <p class="mt-4">This consent is valid <strong>until {{filing.expiry}}</strong>.</p>
+              <p class="mt-4">{{filing.comment}}</p>
             </template>
 
             <!-- else this must be a completed filing -->
@@ -383,7 +389,7 @@ import PaperFiling from './FilingHistoryList/PaperFiling.vue'
 import PendingFiling from './FilingHistoryList/PendingFiling.vue'
 import StaffFiling from './FilingHistoryList/StaffFiling.vue'
 import { AddCommentDialog, DownloadErrorDialog, FileCorrectionDialog, LoadCorrectionDialog } from '@/components/dialogs'
-import { AllowableActions, FilingTypes, Routes, CorrectionTypes, FilingNames } from '@/enums'
+import { AllowableActions, FilingTypes, Routes, CorrectionTypes } from '@/enums'
 import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryItemIF, LegalFilingIF }
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, EnumMixin, FilingMixin } from '@/mixins'
@@ -445,7 +451,6 @@ export default class FilingHistoryList extends Vue {
 
   // enum for template
   readonly AllowableActions = AllowableActions
-  readonly FilingNames = FilingNames
 
   /** The Edit URL string. */
   get editUrl (): string {
@@ -643,6 +648,12 @@ export default class FilingHistoryList extends Vue {
           !!this.tempRegNumber &&
           this.isStatusCompleted(filing)
         )
+      }
+
+      // add properties for Consent to Continuation Outs
+      if (this.isTypeConsentContinuationOut(filing)) {
+        item.expiry = this.apiToPacificDateTime(filing.data.consentContinuationOut?.expiry)
+        item.comment = filing.data.consentContinuationOut?.comment
       }
 
       // add properties for staff filings
--- a/src/components/Dashboard/FilingHistoryList/StaffFiling.vue
+++ b/src/components/Dashboard/FilingHistoryList/StaffFiling.vue
@@ -1,21 +1,24 @@
 <template>
   <div v-if="filing" class="staff-filing-details body-2">
-    <p>{{filing.notationOrOrder}}</p>
+    <p v-if="filing.notationOrOrder" class="mt-4">
+      {{filing.notationOrOrder}}
+    </p>
 
-    <template v-if="filing.documents && filing.documents.length > 0">
-      <DocumentsList
-        :filing=filing
-        :loadingOne=loadingOne
-        :loadingAll=loadingAll
-        :loadingOneIndex=loadingOneIndex
-        @downloadOne="downloadOne(...arguments)"
-        @downloadAll="downloadAll($event)"
-      />
-   </template>
+    <!-- Documents list may be displayed here or in Filing History List. -->
+    <DocumentsList
+      v-if="showDocumentsList && filing.documents && filing.documents.length > 0"
+      :filing=filing
+      @downloadOne="downloadOne(...arguments)"
+      @downloadAll="downloadAll($event)"
+    />
 
-    <p class="mb-0" v-if="filing.fileNumber">Court Order Number: {{filing.fileNumber}}</p>
+    <p v-if="filing.fileNumber" class="mt-4 mb-0">
+      Court Order Number: {{filing.fileNumber}}
+    </p>
 
-    <p class="mt-0" v-if="filing.planOfArrangement">{{filing.planOfArrangement}}</p>
+    <p v-if="filing.planOfArrangement" class="mt-0">
+      {{filing.planOfArrangement}}
+    </p>
   </div>
 </template>
 
@@ -23,7 +26,6 @@
 import Vue from 'vue'
 import { Component, Emit, Prop } from 'vue-property-decorator'
 import { HistoryItemIF } from '@/interfaces'
-import { Getter } from 'vuex-class'
 import DocumentsList from '@/components/Dashboard/FilingHistoryList/DocumentsList.vue'
 
 @Component({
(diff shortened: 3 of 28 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.9.10",
+  "version": "5.10.0",
```
severinbeauvais, line 3: I'm adding something significant, so while the code is backward-compatible, it's a milestone (and resets the patch version so it doesn't grow too large).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -645,6 +645,13 @@ export default class FilingHistoryList extends Vue {
         )
       }
 
+      // add properties for Consent to Continuation Outs
+      if (this.isTypeConsentContinuationOut(filing)) {
+        // *** TODO: implement
+        // *** TODO: also implement special panel in template
+        item.isTypeStaff = true
```
severinbeauvais, line 652: I'm still working on this :)

src/components/Dashboard/StaffNotation.vue
```diff
@@ -145,6 +146,18 @@
...
+              <v-list-item
+                data-type="consent-continue-out"
+                @click="goToConsentContinuationOutFiling()"
+                :disabled="disabled || isAdminFreeze"
+              >
+                <v-list-item-title>
+                  <span class="app-blue">Consent to Continuation Out</span>
+                </v-list-item-title>
+              </v-list-item>
+            </template>
```
severinbeauvais, line 160: Per Ethan, move this item to the bottom of the list, and yes, it's "consent to" not "consent for".

src/components/Dashboard/TodoList.vue
```diff
@@ -1176,9 +1191,12 @@ export default class TodoList extends Vue {
...
       const payErrorObj = paymentStatusCode && await PayServices.getPayErrorObj(paymentStatusCode)
 
-      const draft = changeOfRegistration
       // FUTURE: verify that this checks for all possible data:
-      const haveData = Boolean(draft?.offices || draft?.contactPoint || draft?.parties)
+      const haveData = Boolean(
+        changeOfRegistration.offices ||
+        changeOfRegistration.contactPoint ||
+        changeOfRegistration.parties
+      )
```
severinbeauvais, line 1199: No need for optional chaining since `changeOfRegistration` was already checked on line 1178.

Ditto with `conversion` below.

src/components/Dashboard/TodoList.vue
```diff
@@ -1236,6 +1254,40 @@ export default class TodoList extends Vue {
...
+        // FUTURE: ideally, this would come from the filing:
+        warnings: this.getBusinessWarnings.map(warning => warning.message)
+      }
+      this.todoItems.push(item)
+    } else {
+      // eslint-disable-next-line no-console
+      console.log('ERROR - invalid header or business in filing =', filing)
+    }
+  }
+
```
jonathan-longe, line 1290: Hey, I know we're in a rush to get these changes implemented, but when we have time let's refactor this so that each filing type has it's own TODO component.  I think we'll find this simplifies the code.

src/enums/filingCodes.ts
```diff
@@ -1,8 +1,10 @@
+// FUTURE: add these to shared enum and import that instead
```
severinbeauvais, line 1: I might do this as part of this ticket.

jonathan-longe (approved):

src/components/Dashboard/TodoList.vue
severinbeauvais: Good idea!

src/components/Dashboard/TodoList.vue
severinbeauvais: Outdated conversation.

package.json
ChrisGabel23: Is there anything bad with having a high patch version?

src/components/Dashboard/TodoList.vue
```diff
@@ -1176,10 +1191,6 @@ export default class TodoList extends Vue {
       const paymentStatusCode = header.paymentStatusCode
       const payErrorObj = paymentStatusCode && await PayServices.getPayErrorObj(paymentStatusCode)
 
-      const draft = changeOfRegistration
-      // FUTURE: verify that this checks for all possible data:
-      const haveData = Boolean(draft?.offices || draft?.contactPoint || draft?.parties)
```
severinbeauvais, line 1181: This and code below doesn't apply to Change of Registration, etc.

`isEmptyFiling` only applies to Incorporation Applications and Registrations (which are used to bootstrap a business) to vary some functionality depending on whether the user has entered any data.

src/views/AnnualReport.vue
```diff
@@ -990,7 +990,7 @@ export default class AnnualReport extends Vue {
 
     if (this.isCoop) {
       annualReport = {
-        annualReport: {
+        [FilingTypes.ANNUAL_REPORT]: {
```
severinbeauvais, line 993: "Better" code since it's more generic, even though the old code worked. But I think it does emphasize the point that this data object is named as per the filing type.

tests/unit/FilingHistoryList1.spec.ts
```diff
@@ -430,7 +430,7 @@ describe('Filing History List - misc functionality', () => {
     expect(wrapper.find('.limited-restoration-period').text())
       .toContain('The Company BC1234567 LTD. was successfully')
     expect(wrapper.find('.limited-restoration-period').text())
-      .toContain('restored and active until Apr 1, 2021')
+      .toContain('restored and is active until Apr 1, 2021')
```
severinbeauvais, line 433: Since Travis' PR is not yet merged.

ChrisGabel23 (approved):

src/views/ConsentContinuationOut.vue
```diff
@@ -0,0 +1,822 @@
+<template>
```
severinbeauvais, line 1: This is an initial version only, copied from local corrections. There's lots that has to be updated in here, but it does save and load :)

There's another ticket to finish implementing this.

package.json
severinbeauvais: Not really.

But we had one UI go to patch version 39 or something and that just begs the question of whether we did anything significant in there or was it all bugfixes 😁

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -316,6 +316,12 @@
               <PaperFiling class="mt-6" />
             </template>
 
+            <!-- is this a consent to continuation out filing? -->
+            <template v-else-if="isTypeConsentContinuationOut(filing)">
+              <p class="mt-4">This consent is valid <strong>until {{filing.expiry}}</strong>.</p>
+              <p class="mt-4">{{filing.comment}}</p>
+            </template>
```
severinbeauvais, line 323: It looks like this (except there will be a receipt document and the "Download All" button will automatically be displayed). Ethan has approved this.

image

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -645,6 +650,12 @@ export default class FilingHistoryList extends Vue {
         )
       }
 
+      // add properties for Consent to Continuation Outs
+      if (this.isTypeConsentContinuationOut(filing)) {
+        item.expiry = this.apiToPacificDateTime(filing.data.consentContinuationOut?.expiry)
```
severinbeauvais, line 655: I'm assuming that `expiry` will be an API date (eg, "2023-12-31T08:01:00.000+00:00") but maybe it will instead be "2023-12-31" and the UI will append the "12:01 AM".

This can be updated easily later when we get deeper into this new filing type.

src/components/Dashboard/FilingHistoryList/StaffFiling.vue
```diff
@@ -35,13 +37,7 @@ export default class StaffFiling extends Vue {
   /** The subject filing. */
   @Prop({ required: true }) readonly filing!: HistoryItemIF
 
-  @Prop({ default: false }) readonly loadingOne!: boolean
-
-  @Prop({ default: false }) readonly loadingAll!: boolean
-
-  @Prop({ default: -1 }) readonly loadingOneIndex!: number
```
severinbeauvais, line 42: These 3 props were not needed so I removed them.

src/components/Dashboard/FilingHistoryList/StaffFiling.vue
```diff
@@ -1,29 +1,31 @@
 <template>
```
severinbeauvais, line 1: I updated this component because I was going to reuse it for the Consent to Continuation Out ledger item. Then I decided not to reuse it, but I kept the cleanup/updates.

Sample court order (looks the same as before):

image

src/interfaces/api-filing-interface.ts
```diff
@@ -53,7 +53,8 @@ export interface ApiFilingIF {
     // COD filings only
     changeOfDirectors?: any // some object
 
-    changeOfName?: ChangeOfNameIF
```
severinbeauvais, line 56: I did a global search and it looks like this wasn't used anywhere, so probably a left-over from a previous design.

tests/unit/filings.json
```diff
@@ -54,7 +54,6 @@
       "commentsCount": 5,
       "commentsLink": "https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v2/businesses/CP0000840/filings/112758/comments",
       "data": {
-        "applicationDate": "2019-09-01",
```
severinbeauvais, line 57: Application Date only applies to incorporation applications and registrations.

severinbeauvais:
Please ignore the failed checks above. The first complains about duplicate code, which we can't easily avoid at the moment. The second complains about a small reduction in code coverage (in the test), which we will increase again when we finish implementing the new filing page.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: @vysakh-menon-aot We forgot to decide whether expiry ^^ will be a full API UTC string, or simply YYYY-MM-DD. What do you think?

What is done currently (in Lear) for corp COA filings that are effective the next day at 12:01?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -645,15 +650,22 @@ export default class FilingHistoryList extends Vue {
...
       if (this.isTypeStaff(filing)) {
         // court orders may have documents so leave property as null (they load on toggle)
         if (!this.isTypeCourtOrder(filing)) {
           item.documents = [] // no documents
         }
-        item.fileNumber = filing.data.order?.fileNumber || '' // may be falsy
+        item.fileNumber = filing.data.order?.fileNumber || '' // may be absent
         item.isTypeStaff = true
-        item.notationOrOrder = filing.data.order?.orderDetails // should not be falsy
+        item.details = filing.data.order?.orderDetails // should always be present
```
severinbeauvais, line 668: We don't need 3 completely different names for the same thing. (It's called `details` when we create a staff notation filing. It's called `orderDetails` in the API response. And it _was_ called `notationOrOrder` here.)

src/components/Dashboard/FilingHistoryList.vue
vysakh-menon-aot: for limited restoration (filing.data.restoration) we decide to go with YYYY-MM-DD since that's how the user provides. i could do the same here

severinbeauvais: In this case, it's not the user providing this value -- the API will calculate it (today plus 6 months, I think).

image

Let's go with YYYY-MM-DD for now.

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Ha, good catch. Naming things is SO difficult.

src/components/Dashboard/TodoList.vue
```diff
@@ -1379,6 +1415,12 @@ export default class TodoList extends Vue {
         this.$router.push({ name: Routes.STANDALONE_ADDRESSES, params: { filingId: item.filingId.toString() } })
         break
 
+      case FilingTypes.CONSENT_CONTINUATION_OUT:
+        // resume this Consent to Continuation Out locally
+        this.setCurrentFilingStatus(FilingStatus.DRAFT)
+        this.$router.push({ name: Routes.CONSENT_CONTINUATION_OUT, params: { filingId: item.filingId.toString() } })
+        break
+
```
jonathan-longe, line 1423: Similar comment as above.  We're extending an existing pattern and this is not the appropriate time to refactor, but I'd suggest the TodoList component is strong candidate for refactoring.  Perhaps it's just me, but multiple `if-then-else` or `case` statements based on the same criteria (in this case FilingType) is a hint that there may be a cleaner way to structure this code.

src/enums/filingCodes.ts
```diff
@@ -1,8 +1,10 @@
...
 export enum FilingCodes {
   ANNUAL_REPORT_BC = 'BCANN', // BCOMP - Annual Report
   ANNUAL_REPORT_OT = 'OTANN', // Others - Annual Report
   ADDRESS_CHANGE_BC = 'BCADD', // BCOMP - Change of Address
   ADDRESS_CHANGE_OT = 'OTADD', // Others - Change of Address
+  CONSENT_CONTINUATION_OUT = 'CONTO', // All entity types
```
jonathan-longe, line 7: Should filing codes be a common component?  Are they used in the other repos too?

src/views/ConsentContinuationOut.vue
```diff
@@ -0,0 +1,822 @@
...
+    />
+
+    <PaymentErrorDialog
+      attach="#consent-continuation-out"
+      filingName="Correction"
+      :dialog="paymentErrorDialog"
+      :errors="saveErrors"
+      :warnings="saveWarnings"
+      @exit="onPaymentErrorDialogExit()"
+    />
```
jonathan-longe, line 15: Are dialogs like these common in many components?  Could we simplify the code by listing them in a common place?

jonathan-longe (approved):
Thanks for the opportunity to review.  @severinbeauvais you added this new functionality very quickly!

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -301,8 +302,7 @@
               <FutureEffective :filing=filing class="mt-6" />
             </template>
 
-            <template v-else-if="filing.isTypeLimitedRestoration"
-            >
+            <template v-else-if="filing.isTypeLimitedRestoration">
```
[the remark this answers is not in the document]
severinbeauvais: That's a no-win situation atm. But I'll buy you a beer any time 🍻

src/enums/filingCodes.ts
severinbeauvais: Yes, filing codes are used in other UIs.

src/views/ConsentContinuationOut.vue
severinbeauvais: The local filings (AR, COA, COD, Correction, and now CCO) have dialogs like this.

I'd like to see a clean way to reduce duplication, but this is another pattern I've seen in [Namerequest](https://github.com/bcgov/namerequest/blob/2eedf417b8f8554869fd275c6344b125b1bd091f/src/App.vue#L48) that I don't like (in Namerequest anyway) -- why declare them where they're not needed?

Also, we plan to move AR, COA, COD and Correction to Edit UI ("multi-filigns"), so maybe it's not worth a ton of effort today to refactor those.

src/components/Dashboard/TodoList.vue
severinbeauvais: Agreed -- the Todo List is a strong candidate for refactoring. I thought a bit about it a while back and wanted to do "something" to make it easier to have unit tests (esp. to prevent regressions). I started some unit test changes along those lines but didn't have time to finish it.

severinbeauvais:
Merging now to unblock @jonathan-longe . I will have a few more changes for this ticket but I can do them after.

src/components/Dashboard/TodoList.vue
```diff
@@ -1228,7 +1235,36 @@ export default class TodoList extends Vue {
...
+        name: FilingTypes.CONSENT_CONTINUATION_OUT,
+        filingId: header.filingId,
+        title: FilingNames.CONSENT_CONTINUATION_OUT,
+        draftTitle: FilingNames.CONSENT_CONTINUATION_OUT,
+        status: header.status,
+        enabled: task.enabled,
+        order: task.order,
+        paymentMethod: header.paymentMethod || null,
+        paymentToken: header.paymentToken || null,
+        payErrorObj,
```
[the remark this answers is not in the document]
severinbeauvais: The decision, then, is whether to keep this logic all in this component, or move it to (separate?) sub-components. I'm undecided.

I would like, however, to perhaps combine the object list coming from the API with the "enhanced" object list we're building here... in a future ticket, of course, but food for thought.

PR #452, opened 2023-02-06
15042 New badge + misc fixes

severinbeauvais:
*Issue #:* bcgov/entity#15042

*Description of changes:*
- added display fields to CCO filing
- added v-chip to Entity Info
- fixed reasonText bug
- deleted obsolete "reasonText"
- added safety check to getReasonText()
- fixed apiToDate bug
- updated "is limited restoration"
- added "is authorized continue out"
- added/fixed unit tests
- app version = 5.10.1

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -322,6 +322,8 @@
                 This consent is valid <strong>until {{filing.expiry}} at 12:01 am Pacific time</strong>.
               </p>
               <p class="mt-4">{{filing.comment}}</p>
+              <p v-if="filing.fileNumber" class="mt-4 mb-0">Court Order Number: {{filing.fileNumber}}</p>
+              <p v-if="filing.planOfArrangement" class="mt-0">{{filing.planOfArrangement}}</p>
             </template>
 
             <!-- else this must be a completed filing -->
@@ -657,6 +659,9 @@ export default class FilingHistoryList extends Vue {
         // FUTURE: expiry date may come from business object
         item.expiry = filing.data.consentContinuationOut?.expiry
         item.details = filing.data.consentContinuationOut?.orderDetails
+        // court order properties
+        item.fileNumber = filing.data.order?.fileNumber || '' // may be absent
+        item.planOfArrangement = filing.data.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : ''
       }
 
       // add properties for staff filings
--- a/src/components/EntityInfo.vue
+++ b/src/components/EntityInfo.vue
@@ -13,7 +13,16 @@
             </div>
 
             <!-- Description -->
-            <div v-if="businessId" id="business-subtitle">{{ businessDescription }}</div>
+            <div v-if="businessId">
+              <span id="business-subtitle">{{ businessDescription }}</span>
+              <v-chip
+                v-if="isAuthorizedToContinueOut"
+                id="authorized-to-continue-out-chip"
+                class="primary mt-n1 ml-3 pointer-events-none" small label text-color="white"
+              >
+                <strong>AUTHORIZED TO CONTINUE OUT</strong>
+              </v-chip>
+            </div>
             <div v-if="tempRegNumber" id="ia-reg-subtitle">{{ iaRegDescription }}</div>
           </header>
 
@@ -28,10 +37,13 @@
             </span>
 
             <span v-if="isHistorical">
-              <v-chip class="primary mt-n1 ml-4 pointer-events-none" small label text-color="white">
+              <v-chip
+                id="historical-chip"
+                class="primary mt-n1 ml-4 pointer-events-none" small label text-color="white"
+              >
                 <strong>HISTORICAL</strong>
               </v-chip>
-              <span class="font-14 mx-3">{{reasonText || 'Unknown Reason'}}</span>
+              <span class="font-14 mx-3">{{getReasonText || 'Unknown Reason'}}</span>
             </span>
 
             <template v-else>
@@ -217,15 +229,17 @@ export default class EntityInfo extends Vue {
   @State businessEmail!: string
   @State businessPhone!: string
   @State businessPhoneExtension!: string
-  @State reasonText!: string
 
   @Getter getBusinessNumber!: string
-  @Getter getIdentifier!: number
-  @Getter getEntityName!: string
-  @Getter getNameRequestNumber!: string
-  @Getter isAdminFreeze!: boolean
-  @Getter isPendingDissolution!: boolean
   @Getter getEntityFoundingDate!: Date
+  @Getter getEntityName!: string
+  @Getter getIdentifier!: number
+  @Getter getNameRequestNumber!: string
+  @Getter getReasonText!: string
+  @Getter isAdminFreeze!: boolean
+  @Getter isAuthorizedToContinueOut!: boolean
+  @Getter isHistorical!: boolean
+  @Getter isPendingDissolution!: boolean
 
   // enums for template
   readonly axios = axios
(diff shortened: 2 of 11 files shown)


severinbeauvais:
@jonathan-longe If you get this, refresh your RC and let's chat.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -322,6 +322,8 @@
                 This consent is valid <strong>until {{filing.expiry}} at 12:01 am Pacific time</strong>.
               </p>
               <p class="mt-4">{{filing.comment}}</p>
+              <p v-if="filing.fileNumber" class="mt-4 mb-0">Court Order Number: {{filing.fileNumber}}</p>
+              <p v-if="filing.planOfArrangement" class="mt-0">{{filing.planOfArrangement}}</p>
```
severinbeauvais, line 326: These might not be displayed correctly but the uxpin has not been updated. This can be fixed later.

src/components/EntityInfo.vue
```diff
@@ -28,10 +37,13 @@
...
             <span v-if="isHistorical">
-              <v-chip class="primary mt-n1 ml-4 pointer-events-none" small label text-color="white">
+              <v-chip
+                id="historical-chip"
+                class="primary mt-n1 ml-4 pointer-events-none" small label text-color="white"
+              >
                 <strong>HISTORICAL</strong>
               </v-chip>
-              <span class="font-14 mx-3">{{reasonText || 'Unknown Reason'}}</span>
+              <span class="font-14 mx-3">{{getReasonText || 'Unknown Reason'}}</span>
```
severinbeauvais, line 46: This is a fix.

src/components/EntityInfo.vue
```diff
@@ -13,7 +13,16 @@
...
+            <div v-if="businessId">
+              <span id="business-subtitle">{{ businessDescription }}</span>
+              <v-chip
+                v-if="isAuthorizedToContinueOut"
+                id="authorized-to-continue-out-chip"
+                class="primary mt-n1 ml-3 pointer-events-none" small label text-color="white"
+              >
+                <strong>AUTHORIZED TO CONTINUE OUT</strong>
+              </v-chip>
+            </div>
```
severinbeauvais, line 25: Looks like this:

image

src/interfaces/state-interface.ts
```diff
@@ -37,7 +37,6 @@ export interface StateIF {
   entityStatus: EntityStatus
   entityType: CorpTypeCd
   goodStanding: boolean
-  reasonText: string
```
severinbeauvais, line 40: Obsolete (now using a getter).

src/store/getters.ts
```diff
@@ -286,9 +286,14 @@ export default {
...
-  /** A formatted concatenation of the name and the effectiveDate of the filing  */
-  getReasonText (state: StateIF, getters): string {
-    const filingType: FilingTypes = state.stateFiling.header.name
+  /**
+   * A formatted concatenation of the name and the effective date of the filing.
+   * Only used when entity is historical.
+   */
+  getReasonText (state: StateIF, getters: any): string {
+    const filingType = state.stateFiling?.header?.name
+    if (!filingType) return null // safety check
```
severinbeauvais, line 295: Also I added conditional chaining (?) when getting filing type above.

src/store/getters.ts
```diff
@@ -303,7 +308,7 @@ export default {
       date = DateUtilities.dateToPacificDate(dissolutionDate, true)
     } else {
       name = EnumUtilities.filingTypeToName(filingType)
-      const effectiveDate = this.apiToDate(state.stateFiling.header?.effectiveDate)
+      const effectiveDate = DateUtilities.apiToDate(state.stateFiling.header.effectiveDate)
```
severinbeauvais, line 311: This is a fix.

src/store/getters.ts
```diff
@@ -312,8 +317,19 @@ export default {
     return `${name} ${enDash} ${date}`
   },
 
-  /** */
+  /**
+   * Is True if business is in Limited Restoration state, ie
+   * the last filing to change state was a Restoration filing.
+   */
   isEntityInLimitedRestoration (state: StateIF): boolean {
-    return (state.stateFiling?.business?.state === 'ACTIVE' && state.stateFiling.hasOwnProperty('restoration'))
```
severinbeauvais, line 317: We don't care about the business state in the state filing (though I think it's the current business state anyway, not the state at the time of the filing).

src/store/getters.ts
```diff
@@ -312,8 +317,19 @@ export default {
...
+    return (state.stateFiling?.hasOwnProperty(FilingTypes.RESTORATION))
+  },
```
severinbeauvais, line 326: It's redundant to check if the business is currently active -- it must be if the last state filing was a restoration. (If a limited restoration expires then there will be a new state filing to re-dissolve the business.)

You could argue that checking current business state is a safe thing to do, but I think we're more likely to catch state issues if we don't add that check than if we do (ie, we might notice a state mismatch but we might not notice a missing badge or menu item).

tests/unit/store-actions.spec.ts
```diff
@@ -115,4 +115,58 @@ describe('testing actions', () => {
...
+      }
+    )
+    await store.dispatch('retrieveStateFiling', 'http://localhost/')
+      .then(() => {
+        expect(LegalServices.fetchFiling).toHaveBeenCalled()
+        expect(store.state.stateFiling).toHaveProperty('business')
+        expect(store.state.stateFiling).toHaveProperty('consentContinuationOut')
+        expect(store.state.stateFiling).toHaveProperty('header')
+      })
+  })
```
severinbeauvais, line 171: This doesn't do much but can be extended. (The current ticket is kind of a framework, not filing implementation, ticket.)

ChrisGabel23 (approved):

JazzarKarim (approved):

ketaki-deodhar (approved):

src/components/EntityInfo.vue
```diff
@@ -13,7 +13,16 @@
...
 
             <!-- Description -->
-            <div v-if="businessId" id="business-subtitle">{{ businessDescription }}</div>
+            <div v-if="businessId">
+              <span id="business-subtitle">{{ businessDescription }}</span>
+              <v-chip
+                v-if="isAuthorizedToContinueOut"
+                id="authorized-to-continue-out-chip"
+                class="primary mt-n1 ml-3 pointer-events-none" small label text-color="white"
+              >
```
jonathan-longe, line 22: EntityInfo.vue includes a large number of `v-if` conditionals.  Instead of adding another `v-if` what do you think about refactoring this component?   For example, it looks like the elements visible for entities with a businessId could be separated from those with a temporary registration number.

jonathan-longe (approved):
Thanks for the opportunity to review.

src/components/EntityInfo.vue
severinbeauvais: 1. I have another PR going in for this ticket so I can address some of these comments there.

2. I am wary of conditional chaining because it can hide/mask some issues that we should be handling a better way. In this case, however, since the data came from an API, I think it's OK to use CC, especially since we check the result on the next line. There may be another way to handle this as well (eg, when the data is fetched, validate that the received object is complete). I don't think it's always an anti-pattern. Thoughts? Ideas on improving this in general?

3. The conditional display is interspersed in the template. Probably we'd get lots of duplication if we split this apart, and it might cause future issues if we update the business code without updating the other code (or vice-versa). Thoughts?

severinbeauvais: > 3. The conditional display is interspersed in the template. Probably we'd get lots of duplication if we split this apart, and it might cause future issues if we update the business code without updating the other code (or vice-versa). Thoughts?

I don't see a clean way to separate out the conditional sections in EntityInfo without complicating things. I'm thinking of leaving this one alone.

PR #454, opened 2023-02-09
14783 Fixed voluntary dissolution code

severinbeauvais:
*Issue #:* bcgov/entity#14783

*Description of changes:*
- deleted isTypeDissolution()
- used isTypeVoluntaryDissolution() instead
- replaced enum-mixin with enum-utilities in Filing History List
- misc cleanup
- app version = 5.10.3


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -625,8 +625,8 @@ export default class FilingHistoryList extends Vue {
         }
       }
 
-      // add properties for Dissolutions
-      if (this.isTypeDissolution(filing)) {
+      // add properties for Voluntary Dissolutions
+      if (this.isTypeVoluntaryDissolution(filing)) {
```
severinbeauvais, line 629: I'm calling this a "fix" because, if this was an administrative dissolution, then the old code would execute this block as well as the block at line 668 below.

THIS block is specific to voluntary dissolutions only.

src/services/enum-utilities.ts
```diff
@@ -251,7 +251,6 @@ export default class EnumUtilities {
       case FilingTypes.DISSOLUTION: return FilingNames.DISSOLUTION
       case FilingTypes.DISSOLVED: return FilingNames.DISSOLVED
       case FilingTypes.INCORPORATION_APPLICATION: return FilingNames.INCORPORATION_APPLICATION
-      case FilingTypes.INVOLUNTARY_DISSOLUTION: return FilingNames.INVOLUNTARY_DISSOLUTION
```
severinbeauvais, line 254: There is no such thing as a filing type of "involuntary dissolution". Instead, the type is "dissolution" and the subtype is "involuntary".

Ditto for voluntary dissolution below.

src/mixins/enum-mixin.ts
```diff
@@ -159,6 +154,16 @@ export default class EnumMixin extends Vue {
     return EnumUtilities.isTypeAdministrativeDissolution(item)
   }
 
+  /** DEPRECATED Returns True if filing is an Involuntary Dissolution. */
```
ChrisGabel23, line 157: I don't understand why you're adding deprecated code.

severinbeauvais: To be consistent in how these methods are laid out. This will be all updated in ticket 15234.

The alternative is a "special case" for this method, which probably makes things a bit more complicated than being consistent.

src/services/enum-utilities.ts
severinbeauvais: Ref: https://github.com/bcgov/business-schemas/blob/f70aef8ff39ac6ef2cd6194362aa10ac14260197/src/registry_schemas/schemas/filing.json#L44

ChrisGabel23 (approved):

src/mixins/enum-mixin.ts
jonathan-longe: What would you think about replacing all calls to the `enum-mixin.ts` in `FilingHistory.vue` with calls to `enum-utilities.ts`?  I know it's slightly more work, but they're both pure functions and should be drop-in replacements for each other.  Once complete, we'd be one step closer to retiring the deprecated code in the `enum-mixin.ts`

severinbeauvais: OK.

jonathan-longe (approved):
Nice work @severinbeauvais

JazzarKarim (approved):

src/mixins/enum-mixin.ts
severinbeauvais: Done. Please resolve this conversation if OK.

PR #455, opened 2023-02-09
15051 add current account store module

jonathan-longe:
*Issue #:* /bcgov/entity15051

This PR adds a "CurrentAccount" store module and illustrates a new application architecture.  Rather than loading the current account details from App.vue, the logic is triggered on demand by dispatching an action like this:
```
await this.loadCurrentAccount().catch(() => {throw error})
```
The advantage of this approach is that it allows us to load the resources for the component as needed.  In this particular case, there's no external API triggered by the action so the benefits are minor, but it's a pattern I'd suggest we consider adopting more widely.  Any and all comments are appreciated!


severinbeauvais:
> ```
> await this.loadCurrentAccount().catch(() => {throw error})
> ```

The catch block is redundant here.

src/enums/accountTypes.ts
```diff
@@ -0,0 +1,6 @@
+export enum AccountTypes {
+  PREMIUM = 'PREMIUM',
+  BASIC = 'BASIC',
+  STAFF = 'STAFF',
+  SBC_STAFF = 'SBC_STAFF'
+}
```
severinbeauvais, line 6: I suggest these be moved to the shared repo.

src/store/CurrentAccount/actions.ts
```diff
@@ -0,0 +1,41 @@
+import {CurrentAccountInterface} from "@/interfaces";
```
severinbeauvais, line 1: Lint will ask for spaces after { and before }.

severinbeauvais: Also I think we don't need the semi-colons.

src/store/CurrentAccount/actions.ts
```diff
@@ -0,0 +1,41 @@
...
+import {SessionStorageKeys} from "sbc-common-components/src/util/constants";
+import {sleep} from "@/utils";
+import currentAccount from "@/store/CurrentAccount/index";
```
severinbeauvais, line 4: It feels weird to import from an index file in the current folder.

src/store/CurrentAccount/actions.ts
```diff
@@ -0,0 +1,41 @@
...
+          context.commit('setAccountInformation', accountInfo)
+          resolve()
+        })
+        .catch(() => {
+          reject()
+        })
+    })
+  },
+
+  async waitForCurrentAccount (): Promise<any> {
```
severinbeauvais, line 26: Please copy the comments here as well (or write your own). Mainly, why are we trying 50 times?

src/store/CurrentAccount/actions.ts
```diff
@@ -0,0 +1,41 @@
...
+        const currentAccount = sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+        account = JSON.parse(currentAccount)
+        if (account) break
+          sleep(100)
+      }
+      if (account) {
+        return resolve(account)
+      }
+      return reject()
+    })
```
severinbeauvais, line 39: I _think_ resolving or rejecting a returning promise is equivalent to returning when done or throwing an error. It might make for simpler code?

src/store/CurrentAccount/getters.ts
```diff
@@ -0,0 +1,16 @@
...
+  /** Whether the current account is a premium account. */
+  isPremiumAccount (state: CurrentAccountInterface): boolean {
+    return (state.accountType === AccountTypes.PREMIUM)
+  },
+
+  /** Whether the user is ServiceBC Staff (which is not the same as Staff). */
+  isSbcStaff (state: CurrentAccountInterface): boolean {
+    return (state.accountType === AccountTypes.SBC_STAFF)
+  }
+}
```
severinbeauvais, line 16: Any idea if we can add `isStaff` here (instead of check the auth roles in the KC token as we do now)? Maybe double-check with Travis. Maybe we never need to parse the KC token.

src/store/CurrentAccount/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { state } from './state'
+
+/**
+ * Configures and returns Vuex Store.
+ */
+export default {
+    actions,
+    getters,
+    mutations,
+    state
```
severinbeauvais, line 13: Indent should be 2 spaces.

src/enums/accountTypes.ts
jonathan-longe: Good call.  I might need a quick refresher on the steps involved.   I'll reach out if I need help.

src/store/CurrentAccount/actions.ts
jonathan-longe: Good catch, I'll fix it.

severinbeauvais (commented):
Looks good to me. I'm curious how well other store modules will integrate into our current code -- which way is simpler?

src/store/CurrentAccount/actions.ts
jonathan-longe: Good call, that line isn't needed.  I'll remove it.

src/store/CurrentAccount/actions.ts
jonathan-longe: Good idea, I'll do it.

src/store/CurrentAccount/actions.ts
severinbeauvais: (Async methods automatically have a return type of `<Promise<any>>`.

jonathan-longe: Hmm, I don't think that's true.  The beauty of using promises is that caller can handle a promise-resolve differently from a promise-reject.  For example, I could imagine wanting to do something like this in the future.
```
await this.fetchSomeResourceFromApi()
  .catch((error) => {
    commit('addApiErrorMessage', error)
  })
```

src/store/CurrentAccount/getters.ts
jonathan-longe: No, unfortunately not, the Keycloak resource is separate from the resource that determine Premium vs SBC_STAFF accounts, but I do think we should create a store for the Keycloak resource.

jonathan-longe:
I forgot to mention that I've deliberately not updated the app version number because currently this code is not invoked.  Please let me know if it would be preferable to update the version number anyway.

src/store/CurrentAccount/actions.ts
severinbeauvais: A called can use `.then()` for a resolved promise and `.catch()` for a rejected promise.

Or just await the async method if you don't care about errors.

src/store/CurrentAccount/getters.ts
severinbeauvais: Travis confirmed that `CURRENT_ACCOUNT.accountType` (eg, "STAFF") is not in sync with the Keycloak (jwt) roles array?

No matter if not. It would have been a nice cleanup.

severinbeauvais:
> I forgot to mention that I've deliberately not updated the app version number because currently this code is not invoked. Please let me know if it would be preferable to update the version number anyway.

Aren't you also update the old code to use the new store module?

Either way, please update the app version so we can track these changes through the various environments.

JazzarKarim (approved):
Interesting. Correct if I'm wrong here, whenever you want to use the current user's details, you would call this action now to grab them? This is done instead of the usual loading during the App's initial run?

jonathan-longe:
> Interesting. Correct if I'm wrong here, whenever you want to use the current user's details, you would call this action now to grab them? This is done instead of the usual loading during the App's initial run?

That's correct.

jonathan-longe:
> Aren't you also update the old code to use the new store module?

Yes, but I want to call the action from the Consent to Continue Out page (still in progress) so that it's only loaded when needed.

severinbeauvais:
> > Interesting. Correct if I'm wrong here, whenever you want to use the current user's details, you would call this action now to grab them? This is done instead of the usual loading during the App's initial run?
> 
> That's correct.

💡 This just made more sense to me -- as a distinction between "actions" and "mutations". We have not been using these ideally thus far.

src/store/CurrentAccount/actions.ts
```diff
@@ -0,0 +1,48 @@
...
+      for (let i = 0; i < 50; i++) {
+        const currentAccount = sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+        account = JSON.parse(currentAccount)
+        if (account) break
+        sleep(100)
+      }
+      if (account) {
+        return resolve(account)
+      }
+      return reject(new Error('Gave up waiting for CurrentAccount from Session'))
```
jonathan-longe, line 45: @severinbeauvais I think you're right.  It's considered best practice to throw an error.  This still allows the calling logic to `catch()` and handle the error in a graceful way, but the errors can then be tracked.  I really appreciate your suggestion!!

src/store/CurrentAccount/getters.ts
jonathan-longe: @severinbeauvais No, I traced the code back and can see it's pulling from a different endpoint and getting back a different response. I had a 💡 moment when I realized that I after I've logged into via Keycloak, I can change my CurrentAccount -- one login has many accounts.  Here's the response from the current account endpoint:

Screenshot (25)

package.json
```diff
@@ -20,7 +20,7 @@
     "@bcrs-shared-components/confirm-dialog": "1.2.1",
     "@bcrs-shared-components/corp-type-module": "1.0.9",
     "@bcrs-shared-components/court-order-poa": "2.1.4",
-    "@bcrs-shared-components/enums": "1.0.29",
+    "@bcrs-shared-components/enums": "1.0.33",
```
severinbeauvais, line 23: I was hoping you'd update all the other dependencies here that increased due to shared enum changes.

src/store/CurrentAccount/state.ts
```diff
@@ -0,0 +1,8 @@
+import { CurrentAccountInterface, StateIF } from '@/interfaces'
+
+export const state: CurrentAccountInterface = {
```
severinbeauvais, line 3: We've been naming interfaces , "IF". Are you advocating to change that?

src/store/CurrentAccount/getters.ts
```diff
@@ -0,0 +1,15 @@
+import { AccountTypes } from '@bcrs-shared-components/enums'
+import { CurrentAccountInterface } from '@/interfaces'
+
+export default {
+
```
severinbeauvais, line 5: Extra space.

jonathan-longe: Good catch, I'll fix that

src/store/index.ts
```diff
@@ -4,19 +4,21 @@ import actions from './actions'
 import getters from './getters'
 import mutations from './mutations'
 import { state } from './state'
+import CurrentAccount from '@/store/CurrentAccount'
```
severinbeauvais, line 7: I was wondering where `CurrentAccount` came from since it wasn't declared anywhere. Is there any way to tie this back to the export in the index file?

src/store/CurrentAccount/state.ts
jonathan-longe: Nope, I didn't intend to change that.  I'll correct it.

severinbeauvais (approved):
GTG when you're ready.

package.json
jonathan-longe: Sure, good idea, that's not a problem

src/store/CurrentAccount/actions.ts
severinbeauvais: I think you don't have to throw an error -- anything passed in the reject call will be passed to the caller's catch handler.

package.json
severinbeauvais: Thanks!

src/store/index.ts
jonathan-longe: Vuex is pretty clever.  When referencing actions, getters or mutators you can call them directly without having to know the store in which they reside.  For example, both the following will work:
```
dispatch('getCurrentAccount')  // from the current account module
dispatch('someOtherAction') // from another module
```

We could choose to namespace the stores, but I don't think we'll have so many methods that we'll have name collisions.  Are you comfortable with that?

ChrisGabel23 (approved):

src/store/index.ts
severinbeauvais: No need to namespace the modules.

Resolving this comment.

src/store/CurrentAccount/actions.ts
jonathan-longe: That's the way I've done it previously, but the linter is complaining if I don't put an error in inside reject().  The specific error from the linter reads, "expected the promise rejection reason to be an error".  I'm going to do some further reading.... always learning!

severinbeauvais: I bet it'll work if you get rid of that promise code altogether:
- return the value after your async work
- throw an exception if there's a problem

Promises work well with "then" and "catch" blocks -- that's how most of our network services methods work.

PR #456, opened 2023-02-11
15051 add authorization store module

jonathan-longe:

This PR renames the CurrentAccount store module to Authorization.  The Authorization module now includes getters to the following without having to trigger any actions or mutations.
- the Keycloak user profile
- the current account
- the user's Keycloak roles
- the user's KeycloakGUID
- the user's token

This additional information is made available by querying the state of the `<SbcHeader />` common component using getters in the Authorization store module.

By querying the `<SbcHeader />` store, it should -- in the future -- reduce the setup scaffolding not only in filings-ui, but in create-ui and edit-ui too.


src/interfaces/current-account-interface.ts
```diff
@@ -1,8 +1,12 @@
-// Account information interface
+/** The ROOT state model interface. */
```
severinbeauvais, line 1: I'm not sure what you mean by "root" state model. This is the interface for the current account Vuex module, right?

severinbeauvais: Hmm, actually, no, it's in fact the currentAccount object in the account module... 🤔

src/interfaces/root-state-interface.ts
```diff
@@ -0,0 +1,14 @@
...
+export interface RootStateIF {
+  account: {
+    currentUser: CurrentUserIF,
+    currentAccount: CurrentAccountIF
+  },
+  auth: {
+    token: string,
+    idToken: string,
+    refreshToken: string
+  }
```
severinbeauvais, line 13: Should the other 2 properties (kcGuid and loginSource) be added here?

Why did you decide to embed the auth interface here? Or conversely, why not move the other interfaces here (even if separate from RootStateIF?)

src/store/Authentication/getters.ts
```diff
@@ -0,0 +1,51 @@
+import { AccountTypes } from '@bcrs-shared-components/enums'
+import { RootStateIF } from '@/interfaces'
+
+export default {
+  /** Whether the current account is a premium account. */
+  isPremiumAccount (state, getters, rootState: RootStateIF): boolean {
+    return (rootState.account.currentAccount.accountType === AccountTypes.PREMIUM)
+  },
```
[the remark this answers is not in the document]
severinbeauvais: Neat.

Do you think it would be nice to have a getter in here that returns currentAccount, or does that complicate things?

src/store/Authentication/getters.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /** Returns true is the user is Keycloak authenticated. */
+  isKeycloakAuthenticated (state, getters, rootState, rootGetters): boolean {
+    return rootGetters['auth/isAuthenticated']
+  },
+
+  /** Returns the Keycloak user object. */
+  getKeycloakUser (state, getters, rootState: RootStateIF): any {
+    return rootState.account.currentUser
+  },
```
severinbeauvais, line 33: Given that this file contains getters for auth and account, would there be a better name for the current folder?

src/store/getters.ts
```diff
@@ -6,7 +6,8 @@ import {
 import { DateUtilities, EnumUtilities } from '@/services'
 
 export default {
-  /** The user's Keycloak GUID. */
+  /** The user's Keycloak GUID.
+   * DEPRECATED - use authentication/getKeycloakGuid() instead */
```
severinbeauvais, line 10: How much extra work is it to change calls to `getUserKeycloakGuid` to `authentication/getKeycloakGuid` instead? (Less tech debt.)

src/store/Authentication/getters.ts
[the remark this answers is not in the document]
jonathan-longe: Yes, that's a good idea.  I think as soon as someone needs the currentAccount object they should add a getter that returns it.  At present, I'm not aware that the currentAccount object is needed in its entirety.

src/interfaces/root-state-interface.ts
jonathan-longe: Yes, it's the root state, but since we're not using the 'notification' and 'status' objects I didn't include them, but I can add them if you wish.  The kcGuid and loginSource are available via the CurrentUserIF.  Since the CurrentUserIF (Keycloak) is the authoritative source I thought it would be best to leave them there.  Please push back if you disagree.  There are many ways to skin this 🐱

src/store/Authentication/getters.ts
jonathan-longe: I struggle with the best names for things!  I chose 'Authentication' because I see it as encompassing both the Keycloak login and the selected account.  If you have an alternative suggestion, please let me know.

src/store/getters.ts
jonathan-longe: That's a good idea.  It's only used in a couple of places.  I'll fix it.

JazzarKarim (approved):

src/interfaces/root-state-interface.ts
severinbeauvais: Poor cat.

I was thinking that all the other properties under root (ie, our original store instance) weren't declared here. Should they be?

image

I was expecting an object named "root" to include the properties on the right hand side.

src/store/Authentication/getters.ts
severinbeauvais: `auth-module` ?

src/store/getters.ts
severinbeauvais: Also `isRoleStaff`, please.

src/interfaces/root-state-interface.ts
jonathan-longe: > I was expecting an object named "root" to include the properties on the right hand side.

To be honest, I don't know what's appropriate here.  It's likely possible to access our state via the 'rootState' parameter, but that definitely not something we want to encourage.  Instead, we should access our `state` via the first parameter which would be appropriately typed.  Like this:
```
getSomething(state: InterfaceIF, getters, rootState) {
      return [ 'blah', 'blah', 'blah']
    }
```

src/store/Authentication/getters.ts
jonathan-longe: Is 'auth' short for 'authentication'?  I thought about adding "module" to each but then I was thinking that gets repetitive.  You make the call.  I'm happy to use whatever you choose.
- store
  - auth-module
  - current-account-module
  - todo-module
  - filing-history-module

src/interfaces/root-state-interface.ts
severinbeauvais: Ah, fair enough. Then leave it as is and maybe add a comment summarizing the above? Thx.

src/store/getters.ts
jonathan-longe: Ah, good idea, I'll replace `isRoleStaff` too.

jonathan-longe: @severinbeauvais I need to think about how to test this change.  At present, Jest loads either a fake user token or a fake staff token than calls to `isRoleStaff` are successful.  When I point the `isRoleStaff` getter to the rootState many tests break.  Not sure whether it's best to mock the getter or if there's a better way to do this.  Any suggestions?

src/store/Authentication/getters.ts
severinbeauvais: What is this folder exactly -- store stuff that isn't in our own store?

src/store/getters.ts - ours
src/store/other/getters.ts - external?
src/store/external/getters.ts - external?
src/store/imported/getters.ts - external?

🤷‍♂️

src/store/getters.ts
severinbeauvais: I think I mocked this store getters in a recent PR -- much cleaner than having a test KC token.

It might also be possible to set the store state (eg, roles array) so that the getter returns True.

src/store/Authentication/getters.ts
jonathan-longe: This folder is contains our Vuex authentication module.  At present, it contains only getters that reference the rootState or rootGetters for the current Keycloak user or the user's selected account.  In the near future, I'd like to an 'Authorization' module that would contain the code associated with LaunchDarkly as well as the code associated with user authorizations per businessId.  One of the reasons I'd recommend against the name "auth-module" is that "auth" could mean authentication or authorization.

src/store/getters.ts
jonathan-longe: > It might also be possible to set the store state (eg, roles array) so that the getter returns True.

Unfortunately, that doesn't appear possible.  From what I've read, `rootState` is immutable -- which is probably a sensible decision by the Vuex developers.  I'll set up a mock.

src/store/Authentication/getters.ts
severinbeauvais: We also get authorization info from the auth endpoint.

How about naming it Root and just adding all external store things in this folder?

src/store/getters.ts
severinbeauvais: Is what you're asking about different than this?
`store.state.stateModel.businessId = 'BC1234567'`

jonathan-longe: > Is what you're asking about different than this (which is allowed in the unit tests)? `store.state.stateModel.businessId = 'BC1234567'`
Yes, this is permitted:
 `store.state.stateModel.businessId = 'BC1234567'`
but attempting to change rootState is not -- it's immutable:
 `store.rootState.account.currentUser.roles = ['staff']`

severinbeauvais: Ohhhhhh...... I've never tried assigning `store.rootState`.

src/store/Authentication/getters.ts
jonathan-longe: I was hoping to name our store modules by the type of data they grant access to.  For example, the configuration module has configuration data, the filing history module has filing history data, the todo module has ... you get the idea. 

If we name this module "Root" it doesn't tell developers what type of data is contained within this module. Also, consider that this module, may in the future have it's own state, actions and mutators.  For example, we may want to extend the this module to have custom business logic that combines rootState with our own state.  If we name this module "Root" and agree that this module will only contain references to either rootState or rootGetters, where would our own authentication business logic reside?

severinbeauvais: > I was hoping to name our store modules by the type of data they grant access to. For example, the configuration module has configuration data, the filing history module has filing history data, the todo module has ... you get the idea.

Absolutely.

OK, call it Authentication. We can always change it in the future as we get more context/complicated.

src/store/getters.ts
jonathan-longe: @severinbeauvais would it be possible to leave the replacing `isRoleStaff` to another ticket?  Moving the getter from our store to the Authorization module (or whatever we decide to call it) is trivial.  The time-consuming part is fixing our tests.  As discussed above we would have to change the following in each test.
`store.state.keycloakRoles = ['staff']`

At present, I don't have a single line replacement that can update state in another module.  I'm sure I can get this working, but I don't want to hold up the other work I'm doing.  Thoughts?

severinbeauvais: Yes. Create another ticket. We have too many higher-priority things to work on.

severinbeauvais (approved):
GTG when you're happy with it.

PR #457, opened 2023-02-13
15051 - add configuration store

jonathan-longe:

This pull request adds a store module called, Configuration.  The Configuration store module makes available the attributes stored in `configuration.json` as getters.  Now, instead of having to create methods like this:

```
editUrl (): string {
  return sessionStorage.getItem('EDIT_URL')
},
```
You can simply include the appropriate getter, like this: `...mapGetters(['getEditUrl']),` and then call it like this: `this.getEditUrl`.  

The configuration module will reduce code duplication and move slightly more logic to the store.


src/interfaces/configuration-state-interface.ts
```diff
@@ -0,0 +1,27 @@
...
+    PAY_API_URL: string,
+    PAY_API_VERSION: string
+    REGISTRY_HOME_URL: string,
+    SENTRY_DSN: string,
+    SENTRY_ENABLE: string,
+    SITEMINDER_LOGOUT_URL: string,
+    STATUS_API_URL: string,
+    STATUS_API_VERSION: string
+  }
+}
```
severinbeauvais, line 27: Just a thought about this. Did you see at the last tech demo that Patrick Wei will be changing this config system over when apps are ported to Firebase/GCP in the (near?) future, similar to [BC Registry app](https://github.com/bcgov/bcregistry) I think.

src/main.ts
```diff
@@ -30,9 +30,16 @@ Vue.use(Vue2Filters) // needed by SbcFeeSummary
...
 async function start () {
-  // fetch config from environment and API
-  // must come first as inits below depend on config
-  await fetchConfig()
+  // get config from environment
+  const processEnvBaseUrl: string = process.env.BASE_URL // eg, /business/
+  const windowLocationPathname = window.location.pathname // eg, /business/CP1234567/...
+  const windowLocationOrigin = window.location.origin // eg, http://localhost:8080
+  const applicationUrl = windowLocationOrigin + processEnvBaseUrl
+  await setBaseRouteAndBusinessId(windowLocationPathname, processEnvBaseUrl, windowLocationOrigin) // will throw error
```
severinbeauvais, line 38: Haha, no, should be "may throw error" :)

src/main.ts
```diff
@@ -30,9 +30,16 @@ Vue.use(Vue2Filters) // needed by SbcFeeSummary
...
+
+  // fetch the store first as it has no dependencies
+  const store = getVuexStore()
+  await store.dispatch('fetchConfiguration', applicationUrl)
```
severinbeauvais, line 42: Do you have update insight into whether dispatch or commit need to be awaited? (Thinking back to that unit test that failed when `wrapper.vm.$store.commit('someMutation', someVal)` wasn't awaited.

src/store/Configuration/actions.ts
```diff
@@ -0,0 +1,32 @@
+import axios from '@/axios-auth'
+
+export default {
+  /** Fetches stateFiling from the API and, if successful, triggers mutation. */
```
severinbeauvais, line 4: Not state filing. Configuration?

src/store/Configuration/actions.ts
```diff
@@ -0,0 +1,32 @@
...
+            reject(new Error('Invalid configuration.json'))
+          } else {
+            context.commit('setConfiguration', response.data)
+            context.commit('setSessionVariables', response.data) // For backward compatibility
+            context.commit('setAxiosBaseUrl', context.getters.getLegalApiUrl)
+            resolve(response.data)
+          }
+        })
+        .catch((error) => {
+          reject(new Error('Error: ' + error))
```
severinbeauvais, line 28: Redundant?

src/store/Configuration/mutations.ts
```diff
@@ -0,0 +1,76 @@
+import { ConfigurationStateIF } from '@/interfaces'
+import axios from '@/axios-auth'
+
+export default {
+  setConfiguration (state: ConfigurationStateIF, data: any) {
+    state.configuration = data
+  },
+  setSessionVariables (state: ConfigurationStateIF, responseData: any) {
+    const authWebUrl: string = responseData['AUTH_WEB_URL']
+    sessionStorage.setItem('AUTH_WEB_URL', authWebUrl)
```
severinbeauvais, line 10: We should have a look at what needs to be in session storage and what doesn't.

- is the variable needed for a different UI in the same domain?
- can the variable be fetched from the store when it is needed?
- does the variable need to be shared with another session (tab) in the same browser?
- does/should the variable need to survive if a session is closed and then reopened/restored?

src/store/Configuration/mutations.ts
```diff
@@ -0,0 +1,76 @@
...
+    console.info('Set Status API URL to: ' + statusApiUrl)
+
+    const keycloakConfigPath: string = responseData['KEYCLOAK_CONFIG_PATH']
+    sessionStorage.setItem('KEYCLOAK_CONFIG_PATH', keycloakConfigPath)
+    console.info('Set KeyCloak Config Path to: ' + keycloakConfigPath)
+
+    const siteminderLogoutUrl: string = responseData['SITEMINDER_LOGOUT_URL']
+    if (siteminderLogoutUrl) {
+      sessionStorage.setItem('SITEMINDER_LOGOUT_URL', siteminderLogoutUrl)
+      console.info('Set Siteminder Logout Url to: ' + siteminderLogoutUrl)
```
severinbeauvais, line 49: All these `console.info()` were helpful in earlier days, but maybe not needed now?

image

Personally, I like the "initializing..." and "starting..." messages (and "starting app should/could tell us which app this is), but I think the other messages are no longer needed. Thoughts?

src/utils/fetch-config.ts
```diff
@@ -1,110 +1,18 @@
...
  */
-export async function fetchConfig (): Promise<void> {
-  // get config from environment
-  const processEnvBaseUrl: string = process.env.BASE_URL // eg, /business/
-  const windowLocationPathname = window.location.pathname // eg, /business/CP1234567/...
-  const windowLocationOrigin = window.location.origin // eg, http://localhost:8080
 
-  if (!processEnvBaseUrl || !windowLocationPathname || !windowLocationOrigin) {
+export function setBaseRouteAndBusinessId (pathname: string, processEnvBaseUrl: string, rootUrl: string): Promise<any> {
+  if (!processEnvBaseUrl || !pathname || !rootUrl) {
```
severinbeauvais, line 8: Given how this file had changed so significantly, I propose renaming it.

severinbeauvais (commented):
Looks good. If possible, I'd like to not have two systems for fetch config (session storage + store). I'd prefer to see the old system removed. Is that a lot of extra work?

src/main.ts
jonathan-longe: Haha, good point!  I hope it doesn't always throw an error.  I'll fix the comment.

src/interfaces/configuration-state-interface.ts
jonathan-longe: Unfortunately, no. I was invited to the tech demo, but the meeting link wasn't included.  I'll get it sorted in time for the next meeting.  I'm not sure what Patrick has planned, but it would be great to load the configuration via the .env files.  This makes it easy to override for different environments.

src/main.ts
jonathan-longe: My understanding is that actions are asynchronous by default so the caller needs to either await the result or use `.then()` before invoking any logic that needs the result.   Mutations (aka: commits) must be synchronous.  See:  https://v3.vuex.vuejs.org/guide/mutations.html#mutations-must-be-synchronous

src/store/Configuration/actions.ts
jonathan-longe: Good catch!  I'll fix it.

src/store/Configuration/actions.ts
jonathan-longe: Good question, I had to think about this!  My understanding is that the "error" on line 27 (above) is from the Axios promise, so I want to expose that error.  I'm happy to be corrected if I've got this wrong and it's not needed.

src/store/Configuration/mutations.ts
jonathan-longe: Those are really great questions Severin!  I'll answer your quesions point-by-point below:
 - my understanding was that filing-ui, edit-ui and create-ui each load a similar configuration file on start, so I wasn't thinking that session variables needed to be shared across UIs.  
 - Yes, the store is loaded early so I think the configuration can always be accessed via the store.  The only exceptions appear to be the "VUE_ROUTER_BASE" and "BASE_URL" which I haven't moved to the store.
 - my understanding is that if someone opens a new tab in the same browser, the configuration is loaded again as main.ts and App.vue would need to be called.
 - my assumption is that the configuration is immutable once it's loaded, if that's not the case please let me know I'll do further testing.

src/utils/fetch-config.ts
jonathan-longe: Good idea.  I'll rename it.

jonathan-longe:
> Looks good. If possible, I'd like to not have two systems for fetch config (session storage + store). I'd prefer to see the old system removed. Is that a lot of extra work?

That's a good call.   Let me see what's involved.

JazzarKarim (approved):

src/interfaces/configuration-state-interface.ts
severinbeauvais: No action required here then. This will work for now, and the change to env files will be future work.

src/main.ts
severinbeauvais: Ah, so the commit is async but the mutation it calls is sync?

Or maybe there's something _weird_ going on with that unit test 🤷‍♂️ I don't usually like to ignore things like this that I don't understand 🤓

src/store/Configuration/actions.ts
severinbeauvais: If you don't catch that error here, it will still be exposed by this method's caller (since you're essentially re-throwing the error).

This code, I believe, will display the error twice: once because it's caught here, and another time when it's caught (or not) in the caller. Trigger this error then check the developer console.

severinbeauvais: I'm also 99% sure you don't have to create a new promise or resolve or reject it in this method: if you simply return a value (in the axios.get then block) then it gets wrapped in a promise resolve (since the method is async), and if you throw an error then it gets wrapped in a promise reject.

Try it. It's a nice simplification. Promise === Async.

src/store/Configuration/mutations.ts
severinbeauvais: OK, totally satisfactory answers, but I was more wondering if we need to save any of these values in session storage? Maybe not. As for being useful during development, well I think the Vue browser extension handles that nicely.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -400,6 +400,7 @@ import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryIt
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, FilingMixin } from '@/mixins'
 import { EnumUtilities, LegalServices } from '@/services/'
+import {mapGetters} from "vuex";
```
severinbeauvais, line 403: Linting should have caught this.-
- need a space after { and before }
- should be single quotes
- no semi-colon at end

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -427,16 +428,14 @@ import { EnumUtilities, LegalServices } from '@/services/'
     AllowableActionsMixin,
     DateMixin,
     FilingMixin
-  ]
+  ],
+  computed: {
+    ...mapGetters(['getEditUrl', 'getFilings', 'hasCourtOrders', 'isBenBcCccUlc', 'isFirm', 'isRoleStaff'])
+  }
```
severinbeauvais, line 434: Do getters declared this still way still have a type? That was an advantage of the `@Getter` decorator. (You may be able to declare type in the mapGetters array items.)

severinbeauvais:
Thanks for changing the state/getters over. I didn't realize it was so much work 😛

src/main.ts
jonathan-longe: Commits call the mutation so they *should* both be sync.  Can you send me a link to your test again?  I wonder if there's something else going on.

src/store/Configuration/mutations.ts
jonathan-longe: I've done a search for `setItem(` and `getItem(` and I'm pretty sure I have a good handle on it.  But, refactoring always carries some risk especially when you're new on the project and don't know what you don't know.  :-)

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: I switched `@getters` to `...mapGetters()` because you previously commented in [pull 450](https://github.com/bcgov/business-filings-ui/pull/450)
> I'm OK with this format (instead of using @Getter decorators. I think this is more in line 
> with [Vue3 Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) anyway.

src/main.ts
severinbeauvais: https://github.com/bcgov/business-create-ui/blob/531f63ef1c5f1662a1b46f095f426ae1edcf8631/tests/unit/NameTranslation.spec.ts#L50

src/store/Configuration/mutations.ts
severinbeauvais: So... do we need any of these in session storage?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I did write that. Hmm.

What do you think -- are we losing some helpful typings this way? Is there a solution?

src/store/Configuration/mutations.ts
jonathan-longe: No, not as far as I know

severinbeauvais: Sweet. Nuke them?

src/main.ts
jonathan-longe: If I remove the `await` and immediately call a related getter, line 8 passes.  I think this demonstrates that the mutation take place instantaneously.  

Screenshot (27)

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Good catch, I've got figure out how to tweak my editor so it automatically adds the import statements using the correct syntax.

src/main.ts
severinbeauvais: Maybe it needs a `await Vue.nextTick()` for some validation or reactivity to update...?

severinbeauvais: But, weird, async shouldn't do anything on methods that aren't async, so _theoretically_ awaiting line 50 should have no effect. Except it does.

Don't spend too long on this, though I am curious why it's behaving like this.

src/store/Configuration/mutations.ts
jonathan-longe: Yes, that's what I've done.  The only exceptions are the session variables related to routing.  For example:
- BUSINESS_ID
- TEMP_REG_NUMBER
- BASE_URL
- VUE_ROUTER_BASE

We need to keep these around.

src/main.ts
jonathan-longe: Yes, confirmed, adding `await Vue.nextTick()` allows the entire test to pass.  Looking quickly at "NameRequestInfo.vue", `hasNameTranslation` doesn't instantly update when the store is updated.

src/store/Configuration/mutations.ts
severinbeauvais: OK.

src/main.ts
severinbeauvais: Great, thanks for looking into it.

severinbeauvais: I'll follow your lead in awaiting `store.dispatch()` and not `store.commit()` calls.

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Since typing is important, I think we should return the syntax we were using previously:
`@Getter isRoleStaff!: boolean`

It's slightly more verbose and we run the risk of a mismatch between the return type from the store and the type as defined in a component.

jonathan-longe: If you agree, let me know and I'll reverse some of the changes I made in this PR.

severinbeauvais: I agree. Sorry about that.

src/store/Configuration/actions.ts
jonathan-longe: Sure, I'm happy to give it a try.  Is this what you're thinking?
```
return axios.get(url, { headers })
      .then((response) => {
        if (!response.data) {
          return new Error('Invalid configuration.json')
        } else {
          context.commit('setConfiguration', response.data)
          context.commit('setSessionVariables', response.data)
          context.commit('setAxiosBaseUrl', context.getters.getLegalApiUrl)
          return response.data
        }
      })
```

severinbeauvais: Yes! Does it work as you expect?

src/components/Dashboard/FilingHistoryList/CompletedIa.vue
```diff
@@ -23,10 +23,9 @@ import { navigate } from '@/utils'
 @Component({})
 export default class CompletedIa extends Vue {
   @Getter readonly getEntityName!: string
-
+  @Getter readonly getMyBusinessRegistryUrl!: string
   protected returnToMyBusinessRegistry (): void {
```
severinbeauvais, line 27: Whitespace please.

src/components/Dashboard/TodoList.vue
```diff
@@ -451,6 +451,7 @@ import { LegalServices, PayServices } from '@/services/'
 import { AllowableActions, CorpTypeCd, FilingNames, FilingStatus, FilingTypes, Routes } from '@/enums'
 import { ActionBindingIF, ApiTaskIF, BusinessIF, BusinessWarningIF, ConfirmDialogType, TodoItemIF,
   TodoListResourceIF } from '@/interfaces'
+import { mapGetters } from 'vuex'
```
severinbeauvais, line 454: No longer needed.

src/components/Dashboard/TodoList.vue
```diff
@@ -859,7 +839,7 @@ export default class TodoList extends Vue {
       const corpTypeDescription = this.getCorpTypeDescription(business.legalType)
 
       const paymentStatusCode = header.paymentStatusCode
-      const payErrorObj = paymentStatusCode && await PayServices.getPayErrorObj(paymentStatusCode)
+      const payErrorObj = paymentStatusCode && await PayServices.getPayErrorObj(paymentStatusCode, this.getPayApiUrl)
```
severinbeauvais, line 842: How about standardizing on formal parameter order: URL first, then codes/ids/etc (same between all services?)

severinbeauvais: ^^ I wrote this before checking the other services calls. Maybe no need to change...

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.10.5",
+  "version": "5.10.6",
```
severinbeauvais, line 3: How about going to 5.11.0 since you're making some significant (nice) changes?

tests/unit/base-route.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    } as any
+
+    // call method
+    const store = await getVuexStore() as any // remove typings for unit tests
+    const applicationUrl = 'http://localhost/business/'
+    setBaseRouteAndBusinessId('CP1234567', '/business/', window.location.origin)
+    await store.dispatch('fetchConfiguration', applicationUrl)
+
+    // verify data
+    // expect(sessionStorage.getItem('AUTH_API_URL')).toBe('auth api url/auth api version/')
```
severinbeauvais, line 100: Please delete obsolete code.

src/views/auth/Signout.vue
```diff
@@ -6,14 +6,18 @@
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
 import SbcSignout from 'sbc-common-components/src/components/SbcSignout.vue'
+import { mapGetters } from 'vuex'
 
 @Component({
-  components: { SbcSignout }
+  components: { SbcSignout },
+  computed: {
+    ...mapGetters(['getLoginUrl'])
```
severinbeauvais, line 14: Were you going to change these to `@Getter` syntax so we have typing?

src/views/Dashboard.vue
```diff
@@ -231,7 +231,7 @@ export default {
   computed: {
     ...mapGetters(['isBenBcCccUlc', 'isHistorical', 'isRoleStaff', 'isCoaPending', 'getCoaEffectiveDate',
       'isAppTask', 'isAppFiling', 'getParties', 'isFirm', 'isSoleProp', 'isPartnership', 'getIdentifier',
-      'hasMissingInfoWarning', 'hasComplianceWarning', 'isAdminFreeze']),
+      'hasMissingInfoWarning', 'hasComplianceWarning', 'isAdminFreeze', 'getEditUrl']),
```
severinbeauvais, line 234: I think there's 2 place in this component where you have to change editUrl -> getEditUrl.

src/utils/base-route.ts
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-console */
+
+/**
+ * stores BUSINESS_ID, TEMP_REG_NUMBER, BASE_URL, VUE_ROUTER_BASE from initial route.
+ */
+
+export function setBaseRouteAndBusinessId (pathname: string, processEnvBaseUrl: string, rootUrl: string): Promise<any> {
```
severinbeauvais, line 7: I kinda like the file named the same as the function it exports. I don't mind long filenames :)

src/utils/base-route.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!processEnvBaseUrl || !pathname || !rootUrl) {
+    return Promise.reject(new Error('Missing environment variables'))
```
severinbeauvais, line 9: As discussed, this _should_ work without the Promise.reject call. You should be able to simply throw a new error to reject this "wrapped" promise.

Actually, I don't see anything async in here -- you can simplify it!

src/store/Configuration/mutations.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** Use this mutator to set a specific attribute in unit testing.
+   * Like this: store.commit('setTestConfiguration', ['AUTH_WEB_URL', 'https://auth.web.url/'])
+   * */
+  setTestConfiguration (state: ConfigurationStateIF, payload: any) {
+    if (!state.configuration) {
+      Vue.set(state, 'configuration', {})
+    }
+    Vue.set(state.configuration, payload[0], payload[1])
+  },
+  setSessionVariables (state: ConfigurationStateIF, responseData: any) {
```
severinbeauvais, line 18: How about a blank line between functions?

src/store/Configuration/mutations.ts
```diff
@@ -0,0 +1,38 @@
...
+import axios from '@/axios-auth'
+
+export default {
+  setConfiguration (state: ConfigurationStateIF, data: any) {
+    state.configuration = data
+  },
+  /** Use this mutator to set a specific attribute in unit testing.
+   * Like this: store.commit('setTestConfiguration', ['AUTH_WEB_URL', 'https://auth.web.url/'])
+   * */
+  setTestConfiguration (state: ConfigurationStateIF, payload: any) {
```
severinbeauvais, line 12: If you have time, please create a type for payload.

Does it have to be an array, or is an object OK? Eg,
```
store.commit('setTestConfiguration', { key: 'AUTH_WEB_URL', val: 'https://auth.web.url/' })
```

src/store/Configuration/mutations.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (!state.configuration) {
+      Vue.set(state, 'configuration', {})
+    }
+    Vue.set(state.configuration, payload[0], payload[1])
+  },
+  setSessionVariables (state: ConfigurationStateIF, responseData: any) {
+    const hotjarId: string = responseData['HOTJAR_ID'];
+    (<any>window).hotjarId = hotjarId
```
severinbeauvais, line 20: Looks like responseData is a dictionary?

src/store/Configuration/actions.ts
severinbeauvais: If returning a Promise is needed because this method is called via the store, please add a comment. (We talked about this. I'm just leaving this comment here to remind myself to go back and check your comment.)

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -400,6 +400,7 @@ import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF, DocumentIF, HistoryIt
   from '@/interfaces'
 import { AllowableActionsMixin, DateMixin, FilingMixin } from '@/mixins'
 import { EnumUtilities, LegalServices } from '@/services/'
+import { mapGetters } from 'vuex'
```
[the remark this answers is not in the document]
jonathan-longe: Good catch, I'll fix it.

src/components/Dashboard/FilingHistoryList/CompletedIa.vue
jonathan-longe: Good call, I'll fix it

src/components/Dashboard/TodoList.vue
jonathan-longe: Good catch, I'll fix it

tests/unit/base-route.spec.ts
jonathan-longe: Done, will be visible after the next push

src/views/auth/Signout.vue
jonathan-longe: Yes, good catch, looks like I missed this one

src/views/Dashboard.vue
jonathan-longe: Good catch, I'll fix it.

src/utils/base-route.ts
jonathan-longe: Great, I like that too!

src/utils/base-route.ts
jonathan-longe: Good point, I'll change the return type to 'void' and remove the Promise rejects

ketaki-deodhar (approved):
Lots of learning here. Thank you!

src/store/Configuration/mutations.ts
jonathan-longe: That's a good idea.  It's not how the /store/mutations.js is formatted but I'm happy to set that as a pattern going forward.

src/store/Configuration/mutations.ts
jonathan-longe: I like your suggestion of a creating an object way better!  I'm on it ...

src/store/Configuration/mutations.ts
jonathan-longe: Does JavaScript have a dictionary type?  Would you be okay if I used an Object instead?

src/components/Dashboard/TodoList.vue
jonathan-longe: Yes, I agree, normally I'd place the URL first, but in this case, I placed it second so that it could be optional.  
```
static async getPayErrorObj (code: string, payApiUrl: string = ''): Promise<PaymentErrorIF> {
    const url = `${payApiUrl}codes/errors/${code}`
    return axios.get(url)
      .then(response => response?.data)
      .catch(() => {}) // ignore errors
  }
```

Let me know if you think I should make the url parameter required and, thereby, allow the parameter order to be reversed.

tests/unit/base-route.spec.ts
severinbeauvais: You mean INvisible :)

src/store/Configuration/mutations.ts
severinbeauvais: Sort of. Google it.

Looks to me like dictionary/object is sort of the same thing in JS.

I commented as I was trying to understand your pattern.

src/components/Dashboard/FilingHistoryList/CompletedIa.vue
severinbeauvais: I meant whitespace between getters and local methods.

src/interfaces/set-configuration-interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface SetConfigurationIF  {
```
severinbeauvais, line 1: Or maybe call it something generic that can be reused?

src/components/Dashboard/TodoList.vue
severinbeauvais: Is it ever not required?

src/store/Configuration/actions.ts
severinbeauvais: Thanks for the comment.

src/views/auth/Signout.vue
```diff
@@ -6,15 +6,14 @@
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
 import SbcSignout from 'sbc-common-components/src/components/SbcSignout.vue'
-import { mapGetters } from 'vuex'
+import {Getter} from "vuex-class";
```
severinbeauvais, line 9: Please fix format.

src/components/Dashboard/FilingHistoryList/CompletedIa.vue
jonathan-longe: Ah, got it, thanks for the clarification.  I'll fix it.

src/components/Dashboard/TodoList.vue
jonathan-longe: Fair point!  I had made it optional because I thought that I wouldn't have to change a series of tests that call the function, but I now see there's only one test.  I'll change it.

src/interfaces/set-configuration-interface.ts
jonathan-longe: Good idea.  I'll rename it 'KeyValueIF'

severinbeauvais (approved):
LGTM

PR #459, opened 2023-03-03
15286 Refactored business info into new store module + cleanup

severinbeauvais:
*Issue #:* bcgov/entity#15286

*Description of changes:*

This is a refactor of part of the store to try to simplify our architecture. Next, I will refactor the Allowable Actions Mixin using this new data model.

- moved business info properties from main store to new business module
- moved business info actions from main store to new business module
- moved business info getters from main store to new business module
- moved business info mutations from main store to new business module
- moved business info state from main store to new business module
- now call loadBusinessInfo() action to fetch and commit business info object
- replaced some `@State` with `@Getter`
- renamed some getters to match data name (eg, getEntityName -> getLegalName)
- sorted misc lists
- misc cleanup
- added/updated/fixed unit tests
- app version = 5.11.2


src/App.vue
```diff
@@ -171,8 +172,16 @@ export default {
...
+    ...mapGetters([
+      'getAuthApiUrl',
+      'getBusinessUrl',
+      'getCreateUrl',
+      'getEntityName',
+      'getEntityType',
+      'getIdentifier',
+      'getRegHomeUrl',
+      'isRoleStaff'
+    ]),
```
severinbeauvais, line 184: Sorted and reformatted for easier finding of getters.

Ditto below for actions and mutations.

src/App.vue
```diff
@@ -137,7 +137,6 @@ export default {
       notInGoodStandingDialog: false,
       nigsMessage: null as NigsMessage,
       localNrNumber: null as string,
-      corpTypeCd: null as CorpTypeCd,
```
severinbeauvais, line 140: Now in store instead of local.

src/App.vue
```diff
@@ -357,20 +389,27 @@ export default {
...
-      await this.storeBusinessInfo(data[1])
       this.storeTasks(data[2])
       this.storeFilings(data[3])
       this.storeParties(data[4])
+
+      // if the URL was provided in business info, load state filing
+      if (this.getStateFilingUrl) await this.loadStateFiling(this.getStateFilingUrl)
+
+      // now that we know entity type, store config object
+      this.storeConfigObject()
```
severinbeauvais, line 412: Both of these were previously done inside storeBusinessInfo() (which is now a store action). They depend on business info data, and I didn't want them in the store action (for now), so I moved them here.

src/App.vue
```diff
@@ -565,11 +562,11 @@ export default {
       // Draft Applications are always in good standing
       this.setGoodStanding(true)
 
-      // store NR Number if present
+      // save local NR Number if present
       this.localNrNumber = nameRequest.nrNumber || null
 
       // store Legal Name if present
-      this.setEntityName(nameRequest.legalName || null)
+      if (nameRequest.legalName) this.setEntityName(nameRequest.legalName)
```
severinbeauvais, line 569: This is better -- leave the initial store value alone if we don't set an entity name here.

src/interfaces/business-state-interface.ts
```diff
@@ -0,0 +1,20 @@
+import { EntityState, CorpTypeCd } from '@/enums'
+import { BusinessWarningIF } from '@/interfaces'
+
+/** The state model for the business store module. */
+export interface BusinessStateIF {
+  businessInfo: any // *** TODO: use an interface for this
```
severinbeauvais, line 6: WIP - see comment in Business/state.ts below.

src/store/Business/state.ts
```diff
@@ -0,0 +1,18 @@
+import { BusinessStateIF } from '@/interfaces'
+
+export const state: BusinessStateIF = {
+  businessInfo: {}, // *** TODO: populate defaults as below
```
severinbeauvais, line 4: I'm trying to figure out what to do here (instead of needing to commit all the individual properties). But it looks funny having this business module with just 1 property (businessInfo).

tests/unit/AGMDate.spec.ts
```diff
@@ -24,8 +24,8 @@ describe('AgmDate', () => {
     store.state.ARFilingYear = 2019
     store.state.arMinDate = '2019-01-01'
     store.state.arMaxDate = '2019-12-31'
-    store.state.entityType = 'CP'
-    store.state.lastAnnualReportDate = '2018-07-15'
+    store.state.business.entityType = 'CP'
+    store.state.business.lastAnnualReportDate = '2018-07-15'
```
severinbeauvais, line 28: Update needed for business module (since these store properties are no longer in the root store).

tests/unit/ConfigurationActions.spec.ts
```diff
@@ -7,7 +7,9 @@ import { nextTick } from 'vue'
 // mock the console.info function to hide the output
 console.info = jest.fn()
 
-describe('Fetch Config', () => {
+describe('Configuration Actions', () => {
+  const store = getVuexStore() as any // remove typings for unit tests
```
severinbeauvais, line 11: Should be done just once per test suite, or you get duplicate stores.

src/store/Business/getters.ts
```diff
@@ -0,0 +1,160 @@
...
+    return state.businessInfo.stateFiling
+  },
+
+  /** Is true of the business has a court order filing */
+  hasCourtOrders (state: BusinessStateIF): boolean {
+    return state.hasCourtOrders
+  },
+
+  /** Is True if a firm has at least one "compliance" warning. */
+  hasComplianceWarning (_state: BusinessStateIF, getters): boolean {
```
severinbeauvais, line 66: Using `_state` so this unused parameter doesn't cause a warning.

Note the use of additional parameter `getters` (from this module). There's also `rootState` and `rootGetters` (from rest of store).

src/mixins/common-mixin.ts
```diff
@@ -15,8 +15,8 @@ export default class CommonMixin extends Vue {
   /**
    * Removes the specified properties from nested objects.
    * @param baseObj the base object
-   * @param keys    the nested object keys which to omit properties from
-   * @param prop    the properties to be removed
+   * @param keys the nested object keys which to omit properties from
+   * @param prop the properties to be removed
```
severinbeauvais, line 19: It's still formatted correctly:

image

src/store/Business/actions.ts
```diff
@@ -0,0 +1,93 @@
...
+          // hopefully ops will see this error in Sentry
+          if (corpTypeCd && corpTypeCd !== businessInfo.legalType) {
+            // eslint-disable-next-line no-console
+            console.error('WARNING: Legal Type in Legal db does not match Corp Type in Auth db!')
+          }
+
+          // commit data to store
+          commit('setBusinessInfo', businessInfo)
+
+          // *** TODO: delete these and use getters to access individual properties
```
severinbeauvais, line 41: WIP

src/App.vue
```diff
@@ -560,16 +556,16 @@ export default {
...
       // Draft Applications are always in good standing
       this.setGoodStanding(true)
 
-      // store NR Number if present
+      // save local NR Number if present
       this.localNrNumber = nameRequest.nrNumber || null
 
       // store Legal Name if present
-      this.setEntityName(nameRequest.legalName || null)
+      if (nameRequest.legalName) this.setLegalName(nameRequest.legalName)
```
severinbeauvais, line 568: I think the new code is better -- leave the initial store value alone if we don't set an entity name here.

src/store/Business/mutations.ts
```diff
@@ -0,0 +1,33 @@
+// import Vue from 'vue'
+import { BusinessIF } from '@/interfaces'
+import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
+
+export default {
+  // *** TODO: discuss this with Jonathan
```
severinbeauvais, line 6: WIP

src/store/Business/getters.ts
```diff
@@ -0,0 +1,161 @@
...
+    return state.stateFiling
+  },
+
+  /** Is true of the business has a court order filing */
+  hasCourtOrders (state: BusinessIF): boolean {
+    return state.hasCourtOrders
+  },
+
+  /** Is True if a firm has at least one "compliance" warning. */
+  hasComplianceWarning (_state: BusinessIF, getters): boolean {
```
severinbeauvais, line 67: Using `_state` so this unused parameter doesn't cause a warning.

Note the use of additional parameter `getters` (from this module). There's also `rootState` and `rootGetters` (from rest of store).

src/store/Business/actions.ts
```diff
@@ -0,0 +1,28 @@
...
+    return new Promise((resolve, reject) => {
+      const businessId = sessionStorage.getItem('BUSINESS_ID')
+
+      // if there is no business id, return error
+      if (!businessId) reject(new Error('Missing business id'))
+
+      LegalServices.fetchBusinessInfo(businessId)
+        .then(businessInfo => {
+          // commit data to store
+          commit('setBusinessInfo', businessInfo)
```
severinbeauvais, line 19: This is significant: we fetch the business info and store the whole thing.

src/store/Business/getters.ts
```diff
@@ -0,0 +1,161 @@
...
+  },
+
+  /** The business state. */
+  getBusinessState (state: BusinessIF): EntityState {
+    return state.state
+  },
+
+  /** The founding date. */
+  getFoundingDate (state: BusinessIF): Date {
+    return DateUtilities.apiToDate(state.foundingDate)
```
severinbeauvais, line 23: This is significant: we put some of the logic in the getters. In this example, we return something "different" from the business info object in the store.

JazzarKarim (approved):
Looks good to me. Lots of reading and learning from me since I'm unfamiliar with the Filings UI codebase yet.

src/App.vue
```diff
@@ -475,56 +513,14 @@ export default {
...
-      this.setEntityType(business.legalType)
-      this.setBusinessNumber(business.taxId || null) // may be empty
-      this.setIdentifier(business.identifier)
-      this.setEntityFoundingDate(this.apiToDate(business.foundingDate))
-      this.setLastAnnualReportDate(business.lastAnnualReportDate) // may be empty
-      this.setLastAddressChangeDate(business.lastAddressChangeDate) // may be empty
-      this.setLastDirectorChangeDate(business.lastDirectorChangeDate) // may be empty
-      this.setBusinessWarnings(Array.isArray(business.warnings) ? business.warnings : [])
-      this.setGoodStanding(business.goodStanding)
-      this.setHasCourtOrders(business.hasCourtOrders)
```
jonathan-longe, line 517: Excellent work!  Good riddance to these functions!

src/components/common/Directors.vue
```diff
@@ -582,12 +582,12 @@ export default class Directors extends Vue {
...
-  @Getter getIdentifier!: string
   @Getter getCurrentDate!: string
+  @Getter getFoundingDate!: Date
+  @Getter getIdentifier!: string
+  @Getter getLastAnnualReportDate!: string
+  @Getter getLastDirectorChangeDate!: string
   @Getter isBenBcCccUlc!: boolean
-  @State lastAnnualReportDate!: string
-  @State entityFoundingDate!: Date
-  @State lastDirectorChangeDate!: string
```
jonathan-longe, line 590: Nice work!  It's great to see that we're no longer mutating state directly

src/mixins/date-mixin.ts
```diff
@@ -2,7 +2,7 @@ import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 import { ApiDateTimeUtc, IsoDatePacific } from '@bcrs-shared-components/interfaces'
-import DateUtilities from '@/services/date-utilities'
+import { DateUtilities } from '@/services/'
```
jonathan-longe, line 5: Sorry, that's my fault.  I like the way you wrote it.

src/services/legal-services.ts
```diff
@@ -11,11 +11,20 @@ export default class LegalServices {
...
     return axios.get(url)
+      .then(response => {
+        const businessInfo = response?.data?.business as BusinessIF
+        if (!businessInfo) {
+          // eslint-disable-next-line no-console
+          console.log('fetchBusinessInfo() error - invalid response =', response)
+          throw new Error('Invalid business info')
+        }
+        return businessInfo
+      })
```
jonathan-longe, line 27: Do we need to return a new Promise here?  If "businessInfo" is not returned from the API, does `fetchBusinessInfo` return a Promise.resolve?

severinbeauvais: Because we're returning the then block of an async function, this is inherently a promise.

Then -> resolved

Catch -> rejected

This code works perfectly and as you expect it to, and it's fairly simple-looking 👍

severinbeauvais: Technically, my comment on line 14 is incorrect -- in fact, what is returned is a promise to return the data object -- but that's pretty wordy so I wrote it this way.

src/store/Business/mutations.ts
```diff
@@ -0,0 +1,33 @@
...
+  setBusinessInfo (state: BusinessIF, val: BusinessIF) {
+    // state = val // doesn't work
+
+    // Vue.set(state, 'business', val) // doesn't work
+
+    Object.keys(val).forEach(key => {
+      // Vue.set(state, key, val[key]) // works but more complex than necessary
+      state[key] = val[key]
+    })
+  },
```
jonathan-longe, line 16: After I suggested using `state = val` only to have you discover it doesn't work (sorry!), I think we should follow convention and use `state.businessInfo = val`.

src/store/Configuration/index.ts
```diff
@@ -3,12 +3,10 @@ import getters from './getters'
...
+/** The store configuration module. */
 export default {
-  actions: actions,
-  getters: getters,
-  mutations: mutations,
-  state: state
+  actions,
+  getters,
+  mutations,
+  state
```
jonathan-longe, line 11: Nice, that's cleaner!

src/store/Business/mutations.ts
severinbeauvais: I'm doing more work in this area for the subject ticket. Can I change this in my next commit (so I can merge this PR asap)?

tests/unit/AnnualReport.spec.ts
```diff
@@ -2077,9 +2077,9 @@ describe('Annual Report - payment required error', () => {
 
   beforeEach(() => {
     // init store
-    store.state.identifier = 'CP0001191'
-    store.state.entityType = 'CP'
-    store.state.entityName = 'Legal Name - CP0001191'
+    store.state.business.identifier = 'CP0001191'
+    store.state.business.legalType = 'CP'
+    store.state.business.legalName = 'Legal Name - CP0001191'
```
jonathan-longe, line 2082: When we're writing new tests, what do you think about setting the state by calling a mutator rather than mutating the state directly.  Calling a mutator gives us flexibility to change our store in the future without having to refactor larger number of tests.

tests/unit/ConfigurationActions.spec.ts
jonathan-longe: Good point!

tests/unit/Correction.spec.ts
```diff
@@ -59,10 +59,10 @@ describe('Correction - UI', () => {
...
     // init store
     store.state.currentDate = '2020-03-04'
-    store.state.entityType = 'CP'
-    store.state.entityName = 'My Test Entity'
-    store.state.identifier = 'CP1234567'
-    store.state.entityFoundingDate = new Date('1971-05-12T00:00:00-00:00')
+    store.state.business.legalType = 'CP'
+    store.state.business.legalName = 'My Test Entity'
+    store.state.business.identifier = 'CP1234567'
+    store.state.business.foundingDate = '1971-05-12T00:00:00-00:00'
```
jonathan-longe, line 65: Just checking, is it okay that the JavaScript date is changing to a date string?

jonathan-longe (approved):
Great work @severinbeauvais

tests/unit/Correction.spec.ts
severinbeauvais: In the store (ie, from business object -- from API), it is a string, but to the store users (via getter), it's a JS Date.

tests/unit/AnnualReport.spec.ts
severinbeauvais: I'll update the unit tests I touched in my next PR.

src/services/legal-services.ts
jonathan-longe: If the Axios call is successful but returns bad data, I'm pretty sure Axios returns a Promise.resolve.  In OpenShift (and other hosting platforms), I've seen API calls return successful responses but bad data.  When bad data is returned, we need to return a Promise.reject to get the downstream caller to respond properly.  Didn't you and I test this previously and determine that we needed to `return new Promise ...`?

tests/unit/AnnualReport.spec.ts
jonathan-longe: I'd call a mutation directly using something like this `store.commit('mutationName', newValue)`.  There shouldn't be any need to `await` since mutations are synchronous.  Depending on what store object you need to set, it might be handy to create a mutation specifically for testing.

src/services/legal-services.ts
severinbeauvais: Axios returns an exception for a 4xx or 5xx response, otherwise it returns a resolved promise (then block).

severinbeauvais: What we previously tested is whether the action needs to return a wrapped promise.

tests/unit/AnnualReport.spec.ts
severinbeauvais: ^^ That's what I'm finding: I need to create some business property mutations so that I don't set the entire business object every time.

severinbeauvais: I updated the unit tests in my next PR: https://github.com/bcgov/business-filings-ui/pull/461

tests/unit/Correction.spec.ts
severinbeauvais: Remember: the old code converted the string to a Date in App.vue (on fetch/store); now it's done in the getter.

src/store/Business/mutations.ts
severinbeauvais: I updated this in my next PR: https://github.com/bcgov/business-filings-ui/pull/461/commits/2873ac9eebd85d2e8a1c6c57b541e450b9daab86

PR #462, opened 2023-03-08
15286 Business store module and unit test cleanup

severinbeauvais:
*Issue #:* bcgov/entity#15286

*Description of changes:*
- renamed BusinessIF -> ApiBusinessIF
- fixed "allowed actions" property and type
- now store business info as separate object in state
- renamed fetchBusinessInfo() -> fetchBusiness()
- added business mutations to store individual properties
- replaced store state assignments with store commits in unit tests
- app version = 5.11.3


src/components/Dashboard/TodoList.vue
```diff
@@ -701,7 +701,7 @@ export default class TodoList extends Vue {
   /** Loads a NEW Annual Report todo. */
   private loadAnnualReportTodo (task: ApiTaskIF): void {
     const todo = task.task.todo
-    const business = todo.business as BusinessIF
+    const business = todo.business as ApiBusinessIF
```
severinbeauvais, line 704: I think the new name makes more sense since it's the business object from the API.

src/store/Business/mutations.ts
```diff
@@ -1,33 +1,57 @@
...
 export default {
-  // *** TODO: discuss this with Jonathan
-  setBusinessInfo (state: BusinessIF, val: BusinessIF) {
-    // state = val // doesn't work
+  setAdminFreeze (state: BusinessStateIF, val: boolean) {
+    state.businessInfo.adminFreeze = val
+  },
+
+  setBusinessInfo (state: BusinessStateIF, val: ApiBusinessIF) {
+    state.businessInfo = val
```
severinbeauvais, line 11: The business info object is now a store property, instead of being "the whole store" as before.

severinbeauvais: When the business info object is read from the API, this ^^ mutation is used.

The other mutations in here are primarily for test purposes (ie, where we want to set just one of the business info properties).

tests/unit/AGMDate.spec.ts
```diff
@@ -24,8 +24,8 @@ describe('AgmDate', () => {
     store.state.ARFilingYear = 2019
     store.state.arMinDate = '2019-01-01'
     store.state.arMaxDate = '2019-12-31'
-    store.state.business.legalType = 'CP'
-    store.state.business.lastAnnualReportDate = '2018-07-15'
+    store.commit('setLegalType', 'CP')
+    store.commit('setLastAnnualReportDate', '2018-07-15')
```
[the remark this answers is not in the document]
severinbeauvais: Up to you if you want to review the remaining unit tests but it's this change in all of them -- all assignments to store.state.business now use commits instead.

tests/unit/StaffNotation.spec.ts
```diff
@@ -107,8 +106,8 @@ describe('StaffNotation', () => {
 
   it('renders drop-down menu correctly - active and limited restoration', async () => {
     // set store specifically for this test
-    store.state.business.legalType = CorpTypeCd.COOP
-    store.state.business.state = EntityState.ACTIVE
```
severinbeauvais, line 111: Should use values, not enums, to make the unit tests sensitive to enum changes.

jonathan-longe (approved):
Nice work @severinbeauvais !

PR #463, opened 2023-03-08
15286 Refactor to use allowed actions from API

severinbeauvais:
*Issue #:* bcgov/entity#15286 and bcgov/entity#15079

*Description of changes:*

Commit 1:
- app version = 5.12.0 (since this is a significant, but not major, change)
- imported latest shared Corp Type Module
- updated/added Allowable Actions mixin
- deleted obsolete hasBlockerTask/hasBlockerFiling code (components, store, etc)
- updated EntityInfo, Dashboard, etc to use updated isAllowed() method (including small tweaks to rendered vs disabled)
- updated isAllowed() to use data from Legal API instead of local business logic
- updated unit tests

Commit 2:
- replaced some getLegalName with getEntityName
- cleaned up EntityInfo title and sub-title
- renamed some IDs
- deleted some unneeded getter refs
- added some getLegalName fallback strings
- added getEntityName getter
- updated unit tests

Commit 3:
- moved EntityInfo tombstone data to new component
- added some getters to get rid of `@State` usage
- misc cleanup
- created/updated unit tests


src/components/EntityInfo.vue
```diff
@@ -5,15 +5,15 @@
         <v-col class="flex-column d-flex justify-space-between" cols="12" md="9">
           <header>
             <!-- Entity Name / IA Title -->
-            <div v-if="businessId" id="entity-legal-name" aria-label="Business Legal Name">
+            <div v-if="!!businessId" id="entity-legal-name" aria-label="Business Legal Name">
```
severinbeauvais, line 8: Change from string to boolean.

src/components/EntityInfo.vue
```diff
@@ -46,50 +46,48 @@
...
-                  small text color="primary"
-                  id="company-information-button"
-                  :disabled="hasBlocker"
-                  @click="promptChangeCompanyInfo()"
-                >
-                  <v-icon medium>mdi-file-document-edit-outline</v-icon>
-                  <span class="font-13 ml-1">View and Change Business Information</span>
-                </v-btn>
+            <!-- View and Change Business Information -->
+            <span v-if="!!businessId && !isHistorical">
```
severinbeauvais, line 50: I tweaked the logic a bit here to make more sense:
- don't show unless this is a business (not a temporary registration)
- don't show if business is historical
- disable if not allowed

Same below for Dissolve Business.

src/components/EntityInfo.vue
```diff
@@ -129,70 +127,62 @@
...
               <dd id="nr-number">{{ nameRequestNumber }}</dd>
             </template>
 
             <!-- Email -->
-            <template v-if="businessId">
+            <template v-if="!!businessId">
               <dt class="mr-2">Email:</dt>
               <dd
                 id="entity-business-email"
-                :class="isAllowed(AllowableActions.EDIT_BUSINESS_PROFILE) ? 'cursor-pointer' : 'pointer-events-none'"
```
severinbeauvais, line 167: This is now always true, so I just made it a static class attribute.

Same below for Phone.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -2,101 +2,155 @@ import Vue from 'vue'
...
+        if (this.isFirm) {
+          return this.isAllowedFiling(FilingTypes.CHANGE_OF_REGISTRATION)
+        }
+        return this.isAllowedFiling(FilingTypes.CHANGE_OF_ADDRESS)
       }
 
-      case AllowableActions.ADD_STAFF_COMMENT: {
-        return this.isRoleStaff
+      case AllowableActions.ADMINISTRATIVE_DISSOLUTION: {
+        // FUTURE: check dissolution type
```
severinbeauvais, line 34: Dissolution type is not yet provided by the API - see ticket 15548. There is a note in that ticket to also patch this code and Voluntary Dissolution below, but the current code works as-is.

src/router.ts
```diff
@@ -47,12 +47,17 @@ export function getVueRouter () {
...
+   * Returns True when this is a digital credential route and the credentials
+   * feature flag is OFF (for the current user).
+   */
+  function isRedirectDigitalCredentialRoute (route: Route): boolean {
+    // Initially this is True for all but a few select users.
+    // See DIGITAL_CREDENTIALS in Allowable Actions Mixin.
+    return (
+      !getFeatureFlag('enable-digital-credentials') &&
       [Routes.DIGITAL_CREDENTIALS].includes(route.name as Routes)
+    )
```
severinbeauvais, line 60: Just cleanup in here.

src/store/getters.ts
```diff
@@ -36,27 +36,6 @@ export default {
...
-  /** Is True if there is a blocker including firm compliance. */
-  hasBlocker (_state: StateIF, _getters, _rootState, rootGetters): boolean {
-    // check for compliance warning
-    if (rootGetters.hasComplianceWarning) return true
-
-    // check for missing info warning
-    if (rootGetters.hasMissingInfoWarning) return true
-
-    return rootGetters.hasBlockerExceptStaffApproval
-  },
```
severinbeauvais, line 58: These 2 getters are no longer needed (including the "blocker task" and "blocker filing" states).

src/views/Dashboard.vue
```diff
@@ -48,7 +48,6 @@
                 <StaffNotation
                   v-if="isRoleStaff && !!businessId"
                   addScrollbarOffset="true"
-                  :disabled="!isAllowed(AllowableActions.FILE_STAFF_NOTATION)"
```
severinbeauvais, line 51: This check was useless -- it is always true if the current user is staff -- which is already checked on line 49.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -66,7 +66,7 @@
             <v-list-item
               data-type="registrars-notation"
               @click="showRegistrarsNotationDialog()"
-              :disabled="disabled"
+              :disabled="!isAllowed(AllowableActions.REGISTRARS_NOTATION)"
```
severinbeauvais, line 69: This is cleanup and more concise (ie, driven by BE logic).

Basically, I display all items that are applicable for the current state (active or historical), and then disable the ones that aren't allowed.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -86,91 +86,88 @@
...
               @click="goToConversionFiling()"
-              :disabled="disabled || !isFirm || isAdminFreeze"
+              :disabled="!isAllowed(AllowableActions.RECORD_CONVERSION)"
             >
               <v-list-item-title>
                 <span class="app-blue">Record Conversion</span>
               </v-list-item-title>
             </v-list-item>
 
-            <template v-if="(isFirm || isCoop || isBenBcCccUlc) && !isHistorical">
```
severinbeauvais, line 106: This diff is easier to view if you hide whitespace changes.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -208,40 +206,38 @@ import { mapGetters } from 'vuex'
     ...mapGetters([
       'getIdentifier',
       'isAdminFreeze',
-      'isBenBcCccUlc',
-      'isCoop',
-      'isEntityInLimitedRestoration',
```
severinbeauvais, line 213: Obsolete getters.

src/interfaces/business-interfaces.ts
```diff
@@ -5,6 +5,7 @@ export interface FilingTypeIF {
   displayName: string
   feeCode: string
   name: FilingTypes
+  type?: string // FUTURE: use an enum
```
severinbeauvais, line 8: There are only 4 subtypes at the moment (all related to restorations). There may be more values for dissolution types. When we know more (future) then we can improve this.

package.json
```diff
@@ -20,7 +20,7 @@
     "@bcrs-shared-components/confirm-dialog": "1.2.1",
     "@bcrs-shared-components/corp-type-module": "1.0.9",
     "@bcrs-shared-components/court-order-poa": "2.1.4",
-    "@bcrs-shared-components/enums": "1.0.33",
+    "@bcrs-shared-components/enums": "1.0.37",
```
severinbeauvais, line 23: Contains updated FilingNames enum (ie, for 4 restoration types).

src/App.vue
```diff
@@ -229,7 +230,7 @@ export default {
       const breadcrumbs = this.$route?.meta?.breadcrumb
       const crumbs: Array<BreadcrumbIF> = [
         {
-          text: this.getLegalName || this.getCorpTypeNumberedDescription(this.getLegalType),
+          text: this.getLegalName || GetCorpNumberedDescription(this.getLegalType),
```
severinbeauvais, line 233: Using function right from shared repo (not locally re-exported) -- as I removed it from Enum Mixin (which I hope to get rid of entirely soon).

src/App.vue
```diff
@@ -617,7 +618,7 @@ export default {
         businessIdentifier: this.getIdentifier,
         commentsCount: filedApplication.commentsCount,
         commentsLink: filedApplication.commentsLink,
-        displayName: this.filingTypeToName(header.name),
+        displayName: EnumUtilities.filingTypeToName(header.name),
```
severinbeauvais, line 621: Now using method from utilities class instead of Enum Mixin.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -504,19 +503,8 @@ export default class FilingHistoryList extends Vue {
...
-    const blockerFiling = this.historyItems.some(item => {
-      if (EnumUtilities.isStatusPaid(item)) {
-        if (item.isFutureEffectiveCoaPending) {
-          isCoaPending = true
-          coaEffectiveDate = item.effectiveDate
-        }
-        return true
-      }
-      return false
-    })
```
severinbeauvais, line 516: Obsolete. This business logic is now computed in the Legal API.

src/components/Dashboard/TodoList.vue
```diff
@@ -615,18 +620,6 @@ export default class TodoList extends Vue {
...
-    // This is a blocker because it needs to be completed first.
-    const hasBlockerTask = this.todoItems.some(task => (
-      this.isStatusDraft(task) ||
-      this.isStatusPending(task) ||
-      this.isStatusError(task) ||
-      this.isStatusPaid(task) ||
-      this.isTypeCorrection(task) ||
-      this.isTypeAlteration(task)
-    ))
-    this.setHasBlockerTask(hasBlockerTask)
```
severinbeauvais, line 628: Obsolete. This business logic is now computed in the Legal API.

src/components/EntityInfo.vue
```diff
@@ -266,13 +274,40 @@ export default class EntityInfo extends Vue {
...
+    return this.dateToPacificDate(date, true)
+  }
+
+  /**
+   * True if the business is authorized to continue out, ie, the state filing is a
+   * Consent to Continuation Out filing.
+   */
+  get isAuthorizedToContinueOut (): boolean {
+    return (this.getStateFiling?.hasOwnProperty(FilingTypes.CONSENT_CONTINUATION_OUT) || false)
+  }
```
severinbeauvais, line 302: I moved these here from the store getters file since this is the only place they are used.

src/enums/dissolutionTypesNames.ts
```diff
@@ -1,47 +0,0 @@
-// FUTURE: update these as designs evolve
```
severinbeauvais, line 1: This is now rolled into FilingSubTypes.

src/enums/restorationTypes.ts
```diff
@@ -1,7 +0,0 @@
-// FUTURE: move these to shared enum and import that instead
```
severinbeauvais, line 1: This is now rolled into FilingSubTypes.

src/interfaces/history-item-interface.ts
```diff
@@ -67,6 +67,6 @@ export interface HistoryItemIF {
 
   // limited restorations only
   isTypeLimitedRestoration?: boolean
-  expiry?: Date // also used for consent to continuation out
+  expiry?: string // also used for consent to continuation out
```
severinbeauvais, line 70: This is a fix. It's never been a JS Date; always a string.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -2,101 +2,158 @@ import Vue from 'vue'
...
-          this.isSoleProp
+      case AllowableActions.REGISTRARS_ORDER: {
+        return this.isAllowedFiling(FilingTypes.REGISTRARS_ORDER)
+      }
+
+      case AllowableActions.RESTORATION: {
+        return (
+          this.isAllowedFiling(FilingTypes.RESTORATION, FilingSubTypes.RESTORATION_FULL) ||
+          this.isAllowedFiling(FilingTypes.RESTORATION, FilingSubTypes.RESTORATION_LIMITED)
         )
```
severinbeauvais, line 123: There is still code that uses this case. Ideally, in the future, code would look at only FullRestoration and LimitedRestoration cases.

This is actually even incorrect, since there are 2 other restoration cases that aren't included here. Oh well.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -2,101 +2,158 @@ import Vue from 'vue'
...
-        return (!this.isHistorical && !!businessId && isAllowedEntityType)
       }
 
-      case AllowableActions.VIEW_ADD_DIGITAL_CREDENTIALS: {
-        // Pilot is targeting specific Benefit Companies: Handled on LD Server Side
-        const isFeatureEnabled = !!getFeatureFlag('enable-digital-credentials')
-        const isNotaDcRoute = !(this.$route.matched.some(route => route.name === Routes.DIGITAL_CREDENTIALS))
+      case AllowableActions.STAFF_COMMENT: {
+        return (isBusiness && this.isRoleStaff)
```
severinbeauvais, line 127: Eventually, this and maybe some other cases, will also come from the Legal API. For now this is fine.

src/mixins/enum-mixin.ts
```diff
@@ -1,30 +1,12 @@
 import Vue from 'vue'
```
severinbeauvais, line 1: Gradually making less and less use of this mixin so we can delete it.

src/services/enum-utilities.ts
```diff
@@ -137,19 +132,50 @@ export default class EnumUtilities {
...
+  /**
+   * Returns True if filing is an Administrative Dissolution.
+   * @param item a todo item, state filing or filing item
+   */
+  static isTypeDissolutionAdministrative (item: any): boolean {
+    return (
+      item.filingSubType === FilingSubTypes.DISSOLUTION_ADMINISTRATIVE ||
+      item.dissolution?.dissolutionType === FilingSubTypes.DISSOLUTION_ADMINISTRATIVE ||
+      item.data?.dissolution?.dissolutionType === FilingSubTypes.DISSOLUTION_ADMINISTRATIVE
+    )
```
severinbeauvais, line 144: I realized this method, and the ones below, are called from different components using different data objects. So, I addd the "param" comment indicating the items the method supports, and the code checks all possibilities.

I don't like this, but fixing it would require some API changes (since the API doesn't provide filing sub-type consistently) and maybe splitting this method into separate cases for API objects or UI objects. I think this is OK for now but I admit it's fragile.

src/services/enum-utilities.ts
```diff
@@ -251,15 +312,20 @@ export default class EnumUtilities {
       case FilingTypes.CONVERSION: return FilingNames.CONVERSION
       case FilingTypes.CORRECTION: return FilingNames.CORRECTION
       case FilingTypes.COURT_ORDER: return FilingNames.COURT_ORDER
-      case FilingTypes.DISSOLUTION: return FilingNames.DISSOLUTION
+      case FilingTypes.DISSOLUTION:
+        // FUTURE: move dissolution subtype checks here
+        return FilingNames.DISSOLUTION
```
severinbeauvais, line 317: I didn't move the dissolution subtype checks here because they check yet another parameter (isFirm), which THIS method doesn't know about. Sigh, the webs we weave.

tests/unit/enum-mixin.spec.ts
```diff
@@ -1,100 +0,0 @@
-import Vue from 'vue'
```
severinbeauvais, line 1: Actually, I renamed this file (to enum-utilities) and changed what it tests.

src/components/EntityInfo.vue
```diff
@@ -5,30 +5,44 @@
...
           <header>
             <!-- Entity Name / IA Title -->
-            <div v-if="businessId" id="entity-legal-name" aria-label="Business Legal Name">
+            <div v-if="!!businessId" id="entity-legal-name" aria-label="Business Legal Name">
               <span>{{ getLegalName || 'Unknown Name' }}</span>
             </div>
-            <div v-if="tempRegNumber" id="incorp-app-title" aria-label="Incorporation Application Title">
-              <span>{{ getLegalName || getCorpTypeNumberedDescription(getLegalType)}}</span>
+            <div v-if="!!tempRegNumber" id="incorp-app-title" aria-label="Incorporation Application Title">
+              <span>{{ getLegalName || GetCorpNumberedDescription(getLegalType)}}</span>
```
jonathan-longe, line 12: What do you think of taking the opportunity to refactor the EntityInfo component?  It looks like we might be mixing a number of different display elements:
- entity name, 
- entity status, 
- entity header menu, and; 
- entity key facts

src/components/EntityInfo.vue
severinbeauvais: OK, I'll move it back.

src/components/EntityInfo.vue
jonathan-longe: A single component is fine until it either gets too big or too complicated.  In this case, I was thinking that splitting the component into separate parts makes sense because we use an `EntityInfo` component in the other repos but without the menu and the key facts.  By having a component that just displays the entity name and status, it would allow us to reuse the component in multiple places and have a consistent look as users jump from application to application.  (I'm stepping out on a limb here because I'm assuming the UX folks haven't asked from the `EntityInfo` header to vary between applications.)

src/mixins/enum-mixin.ts
```diff
@@ -184,16 +146,6 @@ export default class EnumMixin extends Vue {
...
-  /** DEPRECATED Returns True if filing is a Limited Restoration. */
-  isTypeLimitedRestoration (item: any): boolean {
-    return EnumUtilities.isTypeLimitedRestoration(item)
-  }
-
-  /** DEPRECATED Returns True if filing is a Staff Only filing. */
-  isTypeStaff (item: any): boolean {
-    return EnumUtilities.isTypeStaff(item)
-  }
-
```
jonathan-longe, line 196: Nice!  It's great to see these deprecated methods deleted.

src/services/enum-utilities.ts
jonathan-longe: Good work @severinbeauvais

src/App.vue
jonathan-longe: What do you think of moving `this.getLegalName || GetCorpNumberedDescription(this.getLegalType)` to a store getter?  If I remember correctly, we reuse this logic in the EntityInfo component.   As we've seen with the sole proprietorship discussion, the logic behind what business name to display can get complicated.  I would suggest that any logic like this belongs in a single place where all components can access it.

src/components/EntityInfo.vue
severinbeauvais: Good points. I'm glad you're thinking clearly, haha.

I'll check on what blocks might be common between our 3 UIs, and if it makes sense then I'll create sub-components for them here. Later, we can reuse those sub-components in the other UIs.

src/App.vue
severinbeauvais: Good idea, will do.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -86,83 +86,78 @@
...
                 <span class="app-blue">Add Court Order</span>
               </v-list-item-title>
             </v-list-item>
 
             <v-list-item
+              v-if="isFirm"
               data-type="record-conversion"
               @click="goToConversionFiling()"
-              :disabled="disabled || !isFirm || isAdminFreeze"
+              :disabled="!isAllowed(AllowableActions.RECORD_CONVERSION)"
```
severinbeauvais, line 100: My thought process here is that this is never applicable unless it's a firm, so only show it for a firm (line 97) and then enable/disable it accordingly (line 100).

src/components/Dashboard/StaffNotation.vue
severinbeauvais, line 1: This diff is easier to view if you hide whitespace changes.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -86,83 +86,78 @@
...
-                <v-list-item-title>
-                  <span class="app-blue">Consent to Continuation Out</span>
-                </v-list-item-title>
-              </v-list-item>
-            </template>
+            <v-list-item
+              v-if="!isHistorical"
+              data-type="administrative-dissolution"
+              @click="showAdministrativeDissolutionDialog()"
+              :disabled="!isAllowed(AllowableActions.ADMINISTRATIVE_DISSOLUTION)"
```
severinbeauvais, line 111: My thought process here is that this is never applicable unless the businss is historical.

JazzarKarim (approved):
LGTM! 👍

src/components/Dashboard/StaffNotation.vue
```diff
@@ -86,83 +86,78 @@
...
+              <v-list-item-title>
+                <span class="app-blue">{{ isAdminFreeze ? 'Unfreeze Business' : 'Freeze Business' }}</span>
+              </v-list-item-title>
+            </v-list-item>
 
             <v-list-item
-              v-if="isInLimitedRestoration"
+              data-type="consent-continue-out"
+              @click="goToConsentContinuationOutFiling()"
+              :disabled="!isAllowed(AllowableActions.CONSENT_CONTINUATION_OUT)"
```
severinbeauvais, line 151: I don't know if this should be shown in all cases (eg, historical?) but we can change this easily in the future as we get UX feedback.

src/enums/allowableActions.ts
```diff
@@ -1,13 +1,24 @@
...
+  LIMITED_RESTO_CONVERT = 'limitedRestoConvert',
+  LIMITED_RESTO_EXTEND = 'limitedRestoExtend',
+  PUT_BACK_ON = 'putBackOn',
+  RECORD_CONVERSION = 'recordConversion',
+  REGISTRARS_NOTATION = 'registrarsNotation',
+  REGISTRARS_ORDER = 'registrarsOrder',
+  RESTORATION = 'restoration',
+  STAFF_COMMENT = 'staffComment',
+  TRANSITION = 'transition',
+  VOLUNTARY_DISSOLUTION = 'voluntaryDissolution',
```
severinbeauvais, line 23: These don't all come from the Legal API yet, but most do 🎉

src/interfaces/api-task-interface.ts
```diff
@@ -36,24 +36,25 @@ export interface ApiHeaderIF {
...
-  alteration?: AlterationIF
   registrarsNotation?: any
   registrarsOrder?: any
-  courtOrder?: any
   registration?: any
-  changeOfRegistration?: any
-  conversion?: any
-  specialResolution?: SpecialResolutionIF
   restoration?: RestorationIF
+  specialResolution?: SpecialResolutionIF
```
severinbeauvais, line 57: Added missing consentContinuationOut and sorted properties.

src/views/Dashboard.vue
```diff
@@ -114,50 +116,54 @@
               </v-card>
             </section>
 
-            <!-- Current Directors-->
-            <section v-if="!isSoleProp && !isPartnership">
+            <!-- Proprietor / Partners -->
+            <section v-if="isFirm">
```
severinbeauvais, line 120: I moved the "is True" condition first and the "else" condition below. I think it's easier to read/understand this way: First, what does this apply to? Then, what's the fallback?

tests/unit/allowable-actions-mixin.spec.ts
```diff
@@ -10,12 +10,17 @@ Vue.use(Vuetify)
...
 describe('Allowable Actions Mixin', () => {
   let vm: any
 
+  function setAllowedFilingType (filingType = {}) {
+    store.commit('setAllowedActions', { filing: { filingTypes: [filingType] } })
+  }
+
+  function setFeatureFlag (val: any) {
+    jest.spyOn((FeatureFlags as any), 'getFeatureFlag').mockReturnValue(val)
+  }
```
severinbeauvais, line 22: ^^ scoped helper functions

src/App.vue
severinbeauvais: Updated in second commit. But there wasn't as much commonality as we thought.

src/components/Tombstone.vue
severinbeauvais, line 1: This is a terrible name but I'm still not thinking clearly. Ideas?

src/components/EntityInfo.vue
severinbeauvais: I identified that only the right hand column ("tombstone") data might be common between Create/Edit/Filings UI, so I refactored that into its own sub-component. See third commit of this PR.

src/components/Tombstone.vue
JazzarKarim: How about TombstoneProfile.vue? Do you reckon it's more specific in this case?

severinbeauvais: It's more than just the profile info.

I was thinking TombstoneInfo, but that's long and redundant.

Keep up the ideas...

ketaki-deodhar (approved):

severinbeauvais:
As this PR was built upon previous code (release 5.11.5), I will merge it to that branch. This will allow PR 467, which is built upon THIS PR, to be more easily compared for review purposes.

Still to do:
1. create a new PR for the content in THIS PR and merge it to main branch
2. rebase PR 467 to main

severinbeauvais:
I created new PR https://github.com/bcgov/business-filings-ui/pull/478 that is rebased to the latest main branch and then contains the allowed actions changes. That PR is ready to merge once the BE changes are merged.

PR #464, opened 2023-03-11
15079 Implemented LIMITED RESTORATION badge + cleaned up enum mixin/utilities

severinbeauvais:
*Issue #:* bcgov/entity#15079

*Description of changes:*
- app version = 5.11.4
- imported latest shared enums
- removed re-export of GetCorpFullDescription and GetCorpNumberedDescription and imported them directly in the components that use them
- replaced some EnumMixin methods with EnumUtilities methods
- changed some businessId (string) to !!businessId (boolean)
- changed some tempRegNumber (string) to !!tempRegNumber (boolean)
- added LIMITED RESTORATION span with badge and text
- removed obsolete "is allowed edit business profile" functionality and checks
- added local getters for isInLimitedRestoration, activeUntil and isAuthorizedToContinueOut
- cleaned up usage of isTypePutBackOn and isTypeDissolutionAdministrative
- replaced DissolutionNames with FilingNames
- replaced DissolutionTypes and RestorationTypes with FilingSubTypes
- cleaned up some interfaces
- updated Enum Utilities (missing cases, comments, etc)
- added/updated/fixed unit tests


src/components/Dashboard/TodoList.vue
```diff
@@ -189,7 +189,7 @@
 
                 <!-- draft correction or conversion or restoration or consent to continuation Out (staff only) -->
                 <template v-else-if="isStatusDraft(item) && (isTypeCorrection(item) || isTypeConversion(item) ||
-                  isTypeRestoration(item) || isTypeConsentContinuationOut(item)) && isRoleStaff"
+                  EnumUtilities.isTypeRestoration(item) || isTypeConsentContinuationOut(item)) && isRoleStaff"
```
severinbeauvais, line 192: I only changed some of these "enum methods" over from EnumMixin to EnumUtilities. But I did delete the old ones along the way.

src/components/EntityInfo.vue
```diff
@@ -160,39 +174,23 @@
             </template>
 
             <!-- Email -->
-            <template v-if="businessId">
+            <template v-if="!!businessId">
               <dt class="mr-2">Email:</dt>
-              <dd
-                id="entity-business-email"
-                :class="isAllowed(AllowableActions.EDIT_BUSINESS_PROFILE) ? 'cursor-pointer' : 'pointer-events-none'"
```
severinbeauvais, line 167: This is always allowed so I replaced the bound class with a static class.

Also below for phone.

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -189,7 +195,7 @@ export default class AddStaffNotationDialog extends Vue {
 
   /** Whether this filing is an Administrative Dissolution. */
   get isAdministrativeDissolution (): boolean {
-    return this.isTypeAdministrativeDissolution({ name: this.name, dissolutionType: this.dissolutionType })
+    return EnumUtilities.isTypeDissolutionAdministrative({ filingSubType: this.dissolutionType })
```
severinbeauvais, line 198: I updated `isTypeDissolutionAdministrative` (and others) to work for different object types, so I was able to simplify this code (and other places).

src/enums/dissolutionTypesNames.ts
```diff
@@ -1,47 +0,0 @@
-// FUTURE: update these as designs evolve
```
severinbeauvais, line 1: I replaced this with FilingSubTypes.

src/enums/restorationTypes.ts
```diff
@@ -1,7 +0,0 @@
-// FUTURE: move these to shared enum and import that instead
```
severinbeauvais, line 1: I merged this into FilingSubTypes.

src/router.ts
```diff
@@ -47,12 +47,17 @@ export function getVueRouter () {
...
+   * Returns True when this is a digital credential route and the credentials
+   * feature flag is OFF (for the current user).
+   */
+  function isRedirectDigitalCredentialRoute (route: Route): boolean {
+    // Initially this is True for all but a few select users.
+    // See DIGITAL_CREDENTIALS in Allowable Actions Mixin.
+    return (
+      !getFeatureFlag('enable-digital-credentials') &&
       [Routes.DIGITAL_CREDENTIALS].includes(route.name as Routes)
+    )
```
severinbeauvais, line 60: Just cleanup in here. (I was in this code for another reason.)

src/services/enum-utilities.ts
```diff
@@ -138,18 +133,43 @@ export default class EnumUtilities {
...
   /** Returns True if filing is an Administrative Dissolution. */
-  static isTypeAdministrativeDissolution (item: any): boolean {
-    return (item.name === FilingTypes.DISSOLUTION && item.dissolutionType === DissolutionTypes.ADMINISTRATIVE)
+  static isTypeDissolutionAdministrative (item: any): boolean {
+    return (
+      // the property in a todo item or filing item:
+      item.filingSubType === FilingSubTypes.DISSOLUTION_ADMINISTRATIVE ||
+      // the property in a state filing:
+      item.dissolution?.dissolutionType === FilingSubTypes.DISSOLUTION_ADMINISTRATIVE
+    )
```
severinbeauvais, line 142: I realized this method, and the ones below, are called from different components using different data objects. So, I added the comments indicating the items the method supports, and the code checks all possibilities.

I don't like this, but fixing it would require some API changes (since the API doesn't provide filing sub-type consistently) and maybe splitting this method into separate cases for API objects or UI objects. I think this is OK for now but I admit it's fragile (*).

(*) The purpose of these tickets/PRs is to implement some new stuff, not refactor all the old stuff. The code changes above and below are actually fixes -- I found a few broken scenarios that now work.

src/services/enum-utilities.ts
```diff
@@ -251,15 +302,20 @@ export default class EnumUtilities {
       case FilingTypes.CONVERSION: return FilingNames.CONVERSION
       case FilingTypes.CORRECTION: return FilingNames.CORRECTION
       case FilingTypes.COURT_ORDER: return FilingNames.COURT_ORDER
-      case FilingTypes.DISSOLUTION: return FilingNames.DISSOLUTION
+      case FilingTypes.DISSOLUTION:
+        // FUTURE: move dissolution subtype checks here
```
severinbeauvais, line 306: I didn't move the dissolution subtype checks here because they check yet another parameter (isFirm), which THIS method doesn't know about. Sigh, the webs we weave.

tests/unit/enum-mixin.spec.ts
```diff
@@ -1,100 +0,0 @@
-import Vue from 'vue'
```
severinbeauvais, line 1: Actually, I renamed this enum-utilities.spec.ts and update it accordingly.

src/mixins/enum-mixin.ts
```diff
@@ -1,30 +1,12 @@
...
-} from '@bcrs-shared-components/corp-type-module'
-import EnumUtilities from '@/services/enum-utilities'
-import { Getter } from 'vuex-class'
+import { EffectOfOrderTypes, FilingStatus, FilingSubTypes, FilingTypes, PaymentMethod } from '@/enums'
 
 /**
  * Mixin that provides some useful enum-related utilities.
  */
 @Component({})
 export default class EnumMixin extends Vue {
```
JazzarKarim, line 9: Is it possible to delete this file since it's deprecated now?

severinbeauvais: There are still many code references to this mixin. I've only moved some of them over (as a favour to my reviewers so they don't have to review a gazillion files).

JazzarKarim (approved):
LGTM! Lots of cleanup and refactoring which is great.

src/components/EntityInfo.vue
```diff
@@ -266,9 +271,15 @@ export default class EntityInfo extends Vue {
     return sessionStorage.getItem('TEMP_REG_NUMBER')
   }
 
+  /** The limited restoration active-until date, if it exists, otherwise null. */
+  get activeUntil (): string {
+    const date = this.yyyyMmDdToDate(this.getStateFiling?.restoration?.expiry)
+    return this.dateToPacificDate(date, true)
+  }
+
```
jonathan-longe, line 279: What do you think about moving this method to a store getter?  At the very least, I think it would be wise to have a getter for the expiry date -- `this.getStateFiling?.restoration?.expiry`.   If we change the getStateFiling API response in the future, having this method as a getter allows us to easily refactor.  While this logic exists in a component, it's more difficult to find.

jonathan-longe: Also, what do you think about bringing the `|| 'Unknown` that exists in the template back into `activeUntil()`?

src/components/EntityInfo.vue
```diff
@@ -5,30 +5,44 @@
...
-              <v-chip
-                v-if="isAuthorizedToContinueOut"
-                id="authorized-to-continue-out-chip"
-                class="primary mt-n1 ml-3 pointer-events-none" small label text-color="white"
-              >
-                <strong>AUTHORIZED TO CONTINUE OUT</strong>
-              </v-chip>
+
+              <span id="limited-restoration" class="ml-3" v-if="isInLimitedRestoration">
+                <span class="ml-3">Active until {{ activeUntil || 'Unknown' }}</span>
```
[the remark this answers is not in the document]
severinbeauvais: On the one hand, yes, having the fallback in the script makes the template cleaner and may make it easier to internationalize. (I think you can call internationalize functions in the template also.)

On the other hand, I kind of like the method returning a falsy value if it doesn't have a real answer, and deferring the display of a fallback value as late as possible in the code. This allows easy logic in the script, if needed, for example: `v-if="activeUntil"`. Providing a fallback in the script doesn't permit this.

I'm hoping to leave this as-is. Thoughts?

src/components/EntityInfo.vue
severinbeauvais: > What do you think about moving this method to a store getter?

Done.

src/store/getters.ts
```diff
@@ -187,23 +193,28 @@ export default {
...
-   * the last filing to change state was a Consent to Continue Out filing.
+   * True if the business is authorized to continue out, ie, the state filing is a
+   * Consent to Continuation Out filing.
    */
-  isAuthorizedToContinueOut (_state: StateIF, _getters, rootState): boolean {
+  isAuthorizedToContinueOut (_state: StateIF, _getters: any, rootState: any): boolean {
     const stateFiling = rootState.stateFiling // may be null
 
-    return (stateFiling?.hasOwnProperty(FilingTypes.CONSENT_CONTINUATION_OUT))
+    return (stateFiling?.hasOwnProperty(FilingTypes.CONSENT_CONTINUATION_OUT) || false)
```
jonathan-longe, line 225: Isn't it possible to access the `stateFiling` via `state` and not `rootState`?  If at all possible, I think we should avoid using `rootState` unless there are no other options.

src/store/getters.ts
```diff
@@ -187,23 +193,28 @@ export default {
...
+  isInLimitedRestoration (_state: StateIF, _getters: any, rootState: any): boolean {
     const stateFiling = rootState.stateFiling // may be null
 
-    return (stateFiling?.hasOwnProperty(FilingTypes.RESTORATION))
+    return (
+      !!stateFiling && (
+        EnumUtilities.isTypeRestorationLimited(stateFiling) ||
+        EnumUtilities.isTypeRestorationExtension(stateFiling)
+      )
+    )
```
jonathan-longe, line 215: Similar to my other comment, is it possible to access `stateFiling` via `state` and not `rootState`?

src/components/EntityInfo.vue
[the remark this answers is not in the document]
jonathan-longe: That's a very good point!  Let's leave it as is.

src/store/getters.ts
severinbeauvais: Yes, I think `stateFiling` is in this store's module, so I should be able to use state. Stand by...

severinbeauvais: I've update this. I've decided to use `getters` to get the state filing, so that's there only 1 place where the state is fetched.

src/store/getters.ts
severinbeauvais: Yes, fixed all 3.

src/store/getters.ts
jonathan-longe: I ❤️ it!  Beautiful solution.

jonathan-longe (approved):
Great work as always @severinbeauvais

PR #465, opened 2023-03-14
15092 - separate restoration endpoints in edit-ui

jonathan-longe:
*Issue #:* /bcgov/entity#15092

In edit-ui, we're overloading the restoration endpoint and reusing it for ~~full restorations,~~ limited-to-full restorations and limited-extension restorations.  By separating restorations to different ~~endpoints~~ routes (thanks Sev!), we can keep the code in edit-ui cleaner and easier to read.  

### Caution, this change will break existing functionality until the receiving endpoints are configured.

### DIFF
--- a/src/components/Dashboard/StaffNotation.vue
+++ b/src/components/Dashboard/StaffNotation.vue
@@ -307,8 +307,7 @@ export default class StaffNotation extends Vue {
         url = `${this.getCreateUrl}?id=${this.getIdentifier}`
       }
       if (applicationName === ApplicationTypes.EDIT_UI) {
-        url = `${this.getEditUrl}${this.getIdentifier}/restoration` +
-          `?restoration-id=${id}`
+        url = `${this.getEditUrl}${this.getIdentifier}/` + restorationType + `?restoration-id=${id}`
       }
       navigate(url)
     } catch (error) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-filings-ui",
-  "version": "5.11.4",
+  "version": "5.11.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "business-filings-ui",
-      "version": "5.11.4",
+      "version": "5.11.5",
       "dependencies": {
         "@babel/compat-data": "^7.19.1",
         "@bcrs-shared-components/breadcrumb": "2.1.11",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "5.11.4",
+  "version": "5.11.5",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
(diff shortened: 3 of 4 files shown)


src/components/Dashboard/StaffNotation.vue
```diff
@@ -307,8 +307,7 @@ export default class StaffNotation extends Vue {
         url = `${this.getCreateUrl}?id=${this.getIdentifier}`
       }
       if (applicationName === ApplicationTypes.EDIT_UI) {
-        url = `${this.getEditUrl}${this.getIdentifier}/restoration` +
-          `?restoration-id=${id}`
+        url = `${this.getEditUrl}${this.getIdentifier}/` + restorationType + `?restoration-id=${id}`
```
jonathan-longe, line 310: Where `restorationType` is one of:
- limitedRestorationToFull
- limitedRestorationExtension
- fullRestoration
- limitedRestoration

JazzarKarim: I think it's cleaner this way. When they first started talking about restorations, there were only two types: Full and Limited. This is why I believe the previous code didn't specify the type in the Url. Having now two more restoration filing types, I believe having the type in the URL is much cleaner. We have to handle the endpoints though like you mentioned!

JazzarKarim: Do you think if we go with this change, a lot of other functionalities might break? We have to dig out what might be affected since everything is so intertwined?

jonathan-longe: No, I'm confident that we can create endpoints for each of the restorationTypes (listed above) and, for now, point them all to the same component.  I'll create an edit-ui PR with just that change so the approval isn't held up by other changes.

jonathan-longe: Ah, good point @severinbeauvais "routes" is more precise!  I would suggest that, ideally, the draft should be created in edit-ui to keep the logic together, but I'd suggest we leave that change for another day.

@ketaki-deodhar, you're correct, this will affect the route you use, but we don't want to have the logic duplicated in two places.  What do you think about removing the `goToRestorationFiling()` method that currently resides in StaffNotation.vue and creating a store action instead?  Then both the ToDo list and the Staff menu can dispatch the same functionality?

severinbeauvais: It would be relatively easy to have restorations create their own draft in Edit UI -- correction filings already do that. That's _if_ we don't need to pass any other information from Filings UI to Edit UI.

Yes, out of scope for this ticket, though could be done in "review and confirm page" ticket. I'm afraid if we create a Tech Debt ticket to change this then we won't get around to it for a while.

However, while we have an ideal design, the current design is satisfactory -- IMHO we have higher priority things to work on.

jonathan-longe: Agreed, thanks @severinbeauvais  @ketaki-deodhar pointed out that the ToDo list has buttons that navigate to edit-ui which don't need to create a draft filing (it's already been created) whereas the StaffNotation needs to create a draft filing prior to navigation.  If we move the creation of the draft to edit-ui, we'd need to create a draft if one doesn't already exist.

severinbeauvais: > If we move the creation of the draft to edit-ui, we'd need to create a draft if one doesn't already exist.

Yes. A number of Edit UI filings already do that.

severinbeauvais (approved):

ketaki-deodhar (approved):

PR #467, opened 2023-03-20
15692 Ledger sub-component refactoring (compared to main branch)

severinbeauvais:
*Issue #:* bcgov/entity#15692

*Description of changes:*
- app version = 6.2.0
- changed FetchingDataSpinner to use store flag (and removed it from FHL)
- changed fetchFilings + storeFilings to a store action
- simplified Pending COA getters
- changed FHL dialogs to use store flags
- refactored to use filings list from API
- moved template blocks and script blocks to separate filing components
- moved many methods to store (actions)
- created common filing template component (with slots for override content)
- created common header actions component
- created common body components
- created filing components
- created common subtitle components
- created FilingHistoryList store module with actions, mutations and getters
- changed some Enum Mixin calls to Enum Utilities
- deleted obsolete Enum Mixin methods
- renamed isAdminFreeze -> isAdminFrozen
- misc cleanup
- updated (or disabled) unit tests -- to be completed in ticket 15762

#### Sample Conversion to Full Restoration Application (collapsed):
image

#### Sample Conversion to Full Restoration Application (expanded):
image

#### Sample Limited Restoration Extension Application (expanded):
image

#### Sample Limited Restoration Extension Application (collapsed):
image


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -530,6 +519,26 @@ export default class FilingHistoryList extends Vue {
...
+      // }
+      // if (EnumUtilities.isTypeAdminFreeze(filing) && (filing.data?.adminFreeze?.freeze === false)) {
+      //   // change this to a Limited Restoration Conversion filing for testing
+      //   filing = {
+      //     ...filing,
+      //     name: FilingTypes.RESTORATION,
+      //     displayName: 'Conversion To Full Restoration Application',
+      //     filingSubType: FilingSubTypes.RESTORATION_CONVERSION
+      //   }
+      // }
```
severinbeauvais, line 540: This is to "fake" restoration extension and conversion filings for development purposes, since the Legal API doesn't return this data yet. I will delete this eventually.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -686,6 +695,12 @@ export default class FilingHistoryList extends Vue {
         item.expiry = filing.data.restoration?.expiry
       }
 
+      // add properties for limited restoration extensions
+      if (EnumUtilities.isTypeRestorationExtension(filing)) {
+        item.isTypeLimitedRestorationExtension = true
+        item.expiry = filing.data.restoration?.expiry
+      }
```
severinbeauvais, line 702: This should be OK since a limited restoration already returns `expiry` like this.

src/components/dialogs/ConfirmDissolutionDialog.vue
```diff
@@ -13,7 +13,7 @@
           {{getModalTitle}}
         </p>
         <p class="warning-text" id="dialog-text">
-          You are about to {{ dissolutionType }} <strong>{{ getLegalName }}</strong>;
+          You are about to {{ dissolutionType }} <strong>{{ getLegalName ||'this company' }}</strong>;
```
severinbeauvais, line 16: I fixed another missing fallback for `getLegalName`.

src/interfaces/api-task-interface.ts
```diff
@@ -36,24 +36,25 @@ export interface ApiHeaderIF {
  * Note that only `business` and `header` are required.
  */
 export interface TaskTodoIF {
+  alteration?: AlterationIF
   annualReport?: any
   business: ApiBusinessIF
   changeOfAddress?: any
   changeOfDirectors?: any
+  changeOfRegistration?: any
+  consentContinuationOut?: any
```
severinbeauvais, line 45: Added missing `consentContinuationOut` property (object).

src/App.vue
```diff
@@ -50,21 +50,22 @@
...
     </v-fade-transition>
 
-    <!-- Alternate Loading Spinner -->
+    <!-- Fetching Data spinner -->
     <v-fade-transition>
-      <div class="loading-container grayed-out" v-show="showSpinner">
+      <div class="loading-container grayed-out" v-show="showFetchingDataSpinner">
         <div class="loading__content">
           <v-progress-circular color="primary" size="50" indeterminate />
+          <div class="loading-msg white--text">Fetching Data</div>
```
severinbeauvais, line 68: This spinner was also on the FilingHistoryList, so I changed the flag to a store variable and just show it here.

src/components/Dashboard/AddressListSm.vue
```diff
@@ -231,16 +231,21 @@ export default class AddressListSm extends Vue {
...
   @Getter isHistorical!: boolean
 
   // FUTURE: change these to getters
   @State registeredAddress!: OfficeAddressIF
   @State recordsAddress!: OfficeAddressIF
 
+  /** Whether a COA is pending. */
+  get isCoaPending (): boolean {
+    return !!this.getPendingCoa
+  }
```
severinbeauvais, line 247: Instead of having a getter for the pending COA filing, and a getter whether it exists, I simplified it to this.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1,373 +1,58 @@
 <template>
   <div id="filing-history-list">
     <AddCommentDialog
-      :dialog="addCommentDialog"
-      :filing="currentFiling"
+      :dialog="showAddCommentDialog"
+      :filing="getCurrentFiling"
```
severinbeauvais, line 5: These dialogs now use store actions/mutations/getters.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1,373 +1,58 @@
...
+    <div class="scrollable-container" v-if="showHistoryPanel">
+      <v-expansion-panels :value="getPanel">
+        <component
+          v-for="(filing, index) in getFilings"
+          :is="is(filing)"
+          :filing="filing"
+          :index="index"
           :key="index"
-        >
-          <v-expansion-panel-header class="no-dropdown-icon pa-0">
```
severinbeauvais, line 53: This expansion panel header is in each filing component.

All the blocks below are also in each filing component, or defaulted (using fallback slots).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -379,346 +64,97 @@
...
-    return (this.historyItems.length > 0)
-  }
-
-  /** Returns whether the action button is visible or not. */
-  protected displayAction (filing): string {
-    return filing.availableOnPaperOnly || filing.isTypeStaff || filing.documentsLink
-  }
-
-  /** Loads list of filings from the API into History Items array. */
-  private loadData (): void {
```
severinbeauvais, line 484: Both loadData() and loadFiling() are obsolete. The template now iterates over getFilings (from store).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -379,346 +64,97 @@
...
+      case EnumUtilities.isTypeRestorationLimited(filing): return 'limited-restoration'
+      case EnumUtilities.isTypeRestorationConversion(filing): return 'limited-restoration-conversion'
+      case EnumUtilities.isTypeRestorationExtension(filing): return 'limited-restoration-extension'
+      case EnumUtilities.isTypeRegistration(filing): return 'registration-filing'
+      case EnumUtilities.isTypeStaff(filing): return 'staff-filing'
+      default: return 'default-filing'
     }
-
-    // report number of items back to parent (dashboard)
-    this.$emit('history-count', this.historyItems.length)
```
severinbeauvais, line 500: This is no longer needed. Instead `getFilings.length` is used (in only 2 places).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -379,346 +64,97 @@
...
-          isCoaPending = true
-          coaEffectiveDate = item.effectiveDate
-        }
-        return true
-      }
-      return false
-    })
-    this.setIsCoaPending(isCoaPending)
-    this.setCoaEffectiveDate(coaEffectiveDate)
-    this.setHasBlockerFiling(blockerFiling)
```
severinbeauvais, line 519: This blocker stuff is part of the allowed actions change and will go away after a rebase.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -379,346 +64,97 @@
...
-        submittedDate,
-        submitter: filing.submitter,
-
-        commentsLink: filing.commentsLink,
-        documentsLink: filing.documentsLink,
-        filingLink: filing.filingLink,
-
-        comments: null, // null until loaded
-        documents: null // null until loaded
-      }
```
severinbeauvais, line 552: No need to do this any more -- it's the same list fetched from the API (plus a couple of properties).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -379,346 +64,97 @@
...
-
-      // add properties for correction filings
-      // (a correction filing has the id of the filing that it corrects)
-      if (filing.correctedFilingId) {
-        item.correctedFilingId = filing.correctedFilingId
-        item.correctedLink = filing.correctedLink
-      }
```
severinbeauvais, line 559: This block and the ones below are now in filing-specific components.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -737,218 +173,23 @@ export default class FilingHistoryList extends Vue {
...
+      this.setFetchingDataSpinner(false)
 
       // eslint-disable-next-line no-console
       console.log('Error creating correction =', error)
-      this.loadCorrectionDialog = true
-    }
-  }
-
-  /** Called by File a Correction button to correct the subject filing. */
-  protected async correctThisFiling (item: HistoryItemIF): Promise<void> {
```
severinbeauvais, line 749: This is now moved to a common HeaderActions component (where it's used).

src/components/Dashboard/FilingHistoryList/FilingTemplate.vue
```diff
@@ -0,0 +1,172 @@
+<template>
```
severinbeauvais, line 1: This template -- with slots for override content -- is common across all filings.

src/components/Dashboard/FilingHistoryList/HeaderActions.vue
```diff
@@ -0,0 +1,182 @@
+<template>
```
severinbeauvais, line 1: This is the common header actions used by FilingTemplate.

This is the fallback "actions" if a filing doesn't override the slot.

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
```diff
@@ -1,15 +1,19 @@
 <template>
```
severinbeauvais, line 1: This common component is used as the body for future effective filings.

severinbeauvais, line 1: This common component is used as the body for future effective filings.

src/components/Dashboard/FilingHistoryList/bodies/FutureEffectivePending.vue
```diff
@@ -1,15 +1,19 @@
 <template>
```
severinbeauvais, line 1: This common component is used as the body for future effective pending filings.

src/components/Dashboard/FilingHistoryList/filings/AlterationFiling.vue
```diff
@@ -0,0 +1,153 @@
+<template>
```
severinbeauvais, line 1: This is the alteration filing component.

As you see below, not all slots are provided -- so the fallback slot content from FilingTemplate will be used.

src/components/Dashboard/FilingHistoryList/filings/AlterationFiling.vue
```diff
@@ -0,0 +1,153 @@
...
+        :index="index"
+      />
+
+      <FutureEffectivePaid
+        v-else-if="isFutureEffective"
+        class="item-header-subtitle"
+        :filing="filing"
+        :index="index"
+      />
+    </template>
```
severinbeauvais, line 17: Here we check 2 conditionals for the "subtitle" slot. If both are false the the fallback (common/generic) slot content is used.

src/components/Dashboard/FilingHistoryList/filings/DefaultFiling.vue
```diff
@@ -0,0 +1,20 @@
+<template>
```
severinbeauvais, line 1: The is the default filing component if we don't need a custom component for a filing type.

src/components/Dashboard/FilingHistoryList/filings/StaffFiling.vue
```diff
@@ -0,0 +1,94 @@
+<template>
```
severinbeauvais, line 1: Special case for all staff filings (of different types) instead of having all separate components.

src/components/Dashboard/FilingHistoryList/subtitles/FiledAndPaid.vue
```diff
@@ -0,0 +1,28 @@
+<template>
```
severinbeauvais, line 1: As above, this time for the "subtitle" slot.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -66,7 +66,7 @@
             <v-list-item
               data-type="registrars-notation"
               @click="showRegistrarsNotationDialog()"
-              :disabled="disabled"
+              :disabled="!isAllowed(AllowableActions.REGISTRARS_NOTATION)"
```
severinbeauvais, line 69: This "is allowed" stuff is part of the allowed actions change and will go away after a rebase.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -293,7 +288,7 @@ export default class StaffNotation extends Vue {
     let url: string
     try {
       // show spinner since the network calls below can take a few seconds
-      this.$root.$emit('showSpinner', true)
+      this.setFetchingDataSpinner(true)
```
severinbeauvais, line 291: Now using store flag for this spinner instead of emitting an event.

src/interfaces/filing-history-list-state-interface.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiFilingIF } from '@/interfaces'
```
severinbeauvais, line 1: Basically, these are the properties that were in FilingHistoryList.vue but are now used by all different sub-components.

src/interfaces/history-item-interface.ts
```diff
@@ -1,72 +0,0 @@
-import { CorpTypeCd, FilingTypes } from '@/enums'
-import { CommentIF, DocumentIF, IsoDatePacific } from '@/interfaces'
-
-/**
- * A Filing History List item (ie, local object).
- * See also Api Filing interface.
- */
-export interface HistoryItemIF {
```
severinbeauvais, line 8: This was a list item in the FHL, but now FHL uses the objects from the API (via a getter).

The "special" properties below are no longer on any object. Instead, they are getters in the components that need them (eg, IA component determines if completed, future effective, etc).

src/interfaces/state-interface.ts
```diff
@@ -28,26 +27,20 @@ export interface StateIF {
...
-
-  hasBlockerTask: boolean
-  hasBlockerFiling: boolean
-  isCoaPending: boolean
-  coaEffectiveDate: Date
+  nameRequest: any
+  parties: Array<PartyIF>
+  recordsAddress: OfficeAddressIF
+  registeredAddress: OfficeAddressIF
+  tasks: Array<ApiTaskIF> // "tasks" data from API
```
severinbeauvais, line 45: I sorted this (and moved `filings` to a separate store module/interface/state).

src/main.ts
```diff
@@ -70,7 +70,12 @@ async function start () {
 
   // configure Keycloak Service
   console.info('Starting Keycloak service...') // eslint-disable-line no-console
-  await KeycloakService.setKeycloakConfigUrl(store.getters.getKeycloakConfigPath)
+  const keycloakConfig: any = {
+    url: 'https://dev.loginproxy.gov.bc.ca/auth',
+    realm: 'bcregistry',
+    clientId: 'entity-web'
+  }
+  await KeycloakService.setKeycloakConfigUrl(keycloakConfig)
```
severinbeauvais, line 78: I imported this change from a change Patrick recently made. It will go away after a rebase.

src/mixins/filing-mixin.ts
```diff
@@ -35,23 +34,6 @@ export default class FilingMixin extends DateMixin {
   @Getter getRegisteredOfficeAddress!: OfficeAddressIF
   @Getter getBusinessAddress!: OfficeAddressIF
 
-  /**
-   * Flattens and sorts an array of comments.
-   * @param comments the array of comments to sort and deconstruct
-   * @returns the sorted and flattened array of comments
-   */
-  flattenAndSortComments (comments: Array<CommentIF>): Array<CommentIF> {
```
severinbeauvais, line 43: I moved this to FHL actions (the only place it's used).

src/services/legal-services.ts
```diff
@@ -41,11 +41,20 @@ export default class LegalServices {
...
     return axios.get(url)
+      .then(response => {
+        const filings = response?.data?.filings as ApiFilingIF[]
+        if (!filings) {
+          // eslint-disable-next-line no-console
+          console.log('fetchFilings() error - invalid response =', response)
+          throw new Error('Invalid filings list')
+        }
+        return filings
+      })
```
severinbeauvais, line 57: This now returns a list of API Filing objects.

When we refactor the Todo List, we should do the same thing for tasks.

src/store/Business/actions.ts
```diff
@@ -11,7 +11,10 @@ export default {
       const businessId = sessionStorage.getItem('BUSINESS_ID')
 
       // if there is no business id, return error
-      if (!businessId) reject(new Error('Missing business id'))
+      if (!businessId) {
+        reject(new Error('Missing business id'))
+        return
+      }
```
severinbeauvais, line 17: I tested this, and sometimes code after "reject" is executed, so a return is needed.

src/store/FilingHistoryList/actions.ts
```diff
@@ -0,0 +1,230 @@
+import { FilingSubTypes } from '@/enums'
```
severinbeauvais, line 1: There isn't much to say about the actions in here. Except for the first one, they were methods in FHL and I moved them here.

src/store/FilingHistoryList/getters.ts
```diff
@@ -0,0 +1,80 @@
...
+  /** The list of filings from the API. */
+  getFilings (state: FilingHistoryListStateIF): ApiFilingIF[] {
+    // return only the valid filings
+    return state.filings.filter(filing => {
+      // safety check for required fields
+      if (!filing.name || !filing.displayName || !filing.effectiveDate || !filing.submittedDate || !filing.status) {
+        console.log('Invalid filing =', filing) // eslint-disable-line no-console
+        return false
+      }
+      return true
```
severinbeauvais, line 18: Returns only the valid filings.

src/views/ConsentContinuationOut.vue
```diff
@@ -612,7 +612,7 @@ export default class ConsentContinuationOut extends Vue {
     const data: any = {
       [FilingTypes.CONSENT_CONTINUATION_OUT]: {
         // FUTURE: add more filing properties below
-        comment: `${this.defaultComment}\n${this.detailComment}`
+        details: `${this.defaultComment}\n${this.detailComment}`
```
severinbeauvais, line 615: Two bugfixes in here...

According to [the schema](https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/consent_continuation_out.json), the property is called details (not comment) and the max length is 2000.

tests/unit/AlterationFiling.spec.ts
```diff
@@ -2,16 +2,16 @@ import Vue from 'vue'
...
-import CompletedAlteration from '@/components/Dashboard/FilingHistoryList/CompletedAlteration.vue'
+import AlterationFiling from '@/components/Dashboard/FilingHistoryList/filings/AlterationFiling.vue'
 
 Vue.use(Vuetify)
 
 const vuetify = new Vuetify({})
 const store = getVuexStore() as any // remove typings for unit tests
 
-describe('Alteration Filing', () => {
+xdescribe('Alteration Filing', () => {
```
severinbeauvais, line 12: The old version of this test suite does not work with this new filing component. This will be completed in ticket 15762. Ditto for other "xdescribe" blocks in files below.

src/store/FilingHistoryList/actions.ts
```diff
@@ -0,0 +1,196 @@
...
+      const businessId = sessionStorage.getItem('BUSINESS_ID')
+      const tempRegNumber = sessionStorage.getItem('TEMP_REG_NUMBER')
+
+      // if there is no business id, return error
+      if (!businessId && !tempRegNumber) {
+        reject(new Error('Missing business id or temp reg number'))
+        return
+      }
+
+      LegalServices.fetchFilings(businessId || tempRegNumber)
```
jonathan-longe, line 23: Would it be possible to pass the `businessId` or `tempRegNumber` to the action as a parameter?  Then we wouldn't have to check for its existence, and it makes `loadFilings()` a pure function.

jonathan-longe: ... well relatively pure for a function that gets from a database. :-)

src/components/Tombstone.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <dl id="tombstone">
+    <!-- Registration Date -->
+    <template v-if="!!registrationDate">
+      <dt class="mr-2">Registration Date:</dt>
+      <dd id="registration-date">{{ registrationDate }}</dd>
+    </template>
```
jonathan-longe, line 7: This looks like a file that's part of a separate pull request.  Is it possible that changes from two PRs are shown here?

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Beautiful!

src/services/legal-services.ts
jonathan-longe: Instead of throwing an error, what do you think about having `fetchFilings()` always return a promised instead.  This approach would then allow downstream logic to gracefully handle the error in the most appropriate way.  When we throw an error, doesn't the browser just display an error in the console log?  The regular, non-technical user wouldn't know what had happened -- right?

src/store/Business/actions.ts
jonathan-longe: That's a good find!  Can you share your test so we can all learn the circumstances where a return is required.

src/views/Dashboard.vue
```diff
@@ -43,19 +43,17 @@
...
                   addScrollbarOffset="true"
-                  :disabled="!isAllowed(AllowableActions.FILE_STAFF_NOTATION)"
                   @close="reloadDataIfNeeded($event)"
                 />
               </header>
               <FilingHistoryList
                 class="mt-3"
                 :highlightId="filingId"
-                @history-count="historyCount = $event"
               />
```
jonathan-longe, line 57: What do you think of passing the `businessId` as prop into `FilingHistoryList`?  Since `businessId` is required, we should be explicit and show what's needed.

jonathan-longe: This also fits with our long-term plan of having `businessId` become a route parameter that's passed from VueRouter to the view as a prop.

src/store/FilingHistoryList/actions.ts
severinbeauvais: I think it's possible to pass in 1 extra parameter, but it could be an object with 2 properties.

I don't think businessId or tempRegNumber need to be parameterized, since they are set once when the app starts and don't change thereafter.

I would love to get these values from somewhere else than session state every time. Maybe the store, if they're set before the getter sees them.

However, the focus of this ticket was the ledger refactoring and I consider this a bit out of scope. Yes, I'll fix small things along the way, but I tried to leave existing functions alone as much as possible. In the example above, I pretty copied the code out of App.vue.

We'll see how today goes. Maybe I'll have time to update this and maybe higher priority things will take precedence.

src/components/Tombstone.vue
severinbeauvais: You're correct. This development builds upon the allowed actions work and crosses with the entity info refactor I did recently. I haven't untangled this PR yet. I'm hoping the rebase is relatively smooth....

src/services/legal-services.ts
severinbeauvais: This error handling both logs an error to the console (with the subject response) and throws an error, which is super-easy to handle in the caller. For example:
```
const filings = await fetchFilings(123)
   .then (/* do something with the return value */)
   .catch (/* do something with the error */)
```
or
```
try {
   const filings = await fetchFilings(123)
   /* do something with the return value */
} catch (error) {
   /* do something with the error */
}
```

src/store/Business/actions.ts
severinbeauvais: I don't have a specific test for this -- I added a console.log statement after the reject call and checked if it ran or not.

In fact, in this action, the regular statement after the reject call was not called. However, in loadFilings() (in FHL actions file), the regular statement after the reject call WAS called. So I kept it consistent between both actions.

I just Googled it again and it looks like reject is supposed to act like a return statement. Except I saw above that it sometimes doesn't. To add to the mystery, my code tools tell me when subsequent statements are never run (eg, due to a throw or return) but here they don't say anything.

Thoughts?

src/views/Dashboard.vue
severinbeauvais: Yes, I like being explicit about what parameters a component needs (eg, props).

Again, businessId doesn't change once the app has started, so why parameterize it? But it would be nice if it came from somewhere central and internal (like the store, not session storage).

Open to other ideas here, but I think that should be handled in a different (new) ticket.

severinbeauvais: > This also fits with our long-term plan of having `businessId` become a route parameter that's passed from VueRouter to the view as a prop.

Sure, but businessId never changes, so why parameterize it? (Other than for visibility.) I mean, the FHL will never run on its own or in a new app, so why can't we consider the businessId to be pretty much static (like, say, the environment variables).

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1,373 +1,58 @@
...
       </v-expansion-panels>
     </div>
 
     <!-- No Results Message -->
     <v-card class="no-results" flat v-if="!showHistoryPanel">
       <v-card-text>
-        <template v-if="!!tempRegNumber">
+        <template v-if="isTemporaryRegistration">
           <div class="no-results__subtitle">Complete your filing to display</div>
         </template>
```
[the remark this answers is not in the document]
severinbeauvais: I think it's only a few lines and not "unclean" enough to make me want another component.

What changing this would do is move this logic (isTemporaryRegistration vs isBusiness) to the parent component (Dashboard), so you wouldn't be getting rid of any code.

I can do it if you insist, but I don't see a strong reason to do it.

src/components/Dashboard/FilingHistoryList/DocumentsList.vue
```diff
@@ -31,32 +31,56 @@
...
+  @Getter getLoadingOneIndex!: number
+  @Getter isLoadingAll!: boolean
+  @Getter isLoadingOne!: boolean
 
-  /** Whether all documents are currently loading. */
-  @Prop({ default: false }) readonly loadingAll!: boolean
+  @Mutation mutateDownloadErrorDialog!: (x: boolean) => void
+  @Mutation mutateLoadingAll!: (x: boolean) => void
+  @Mutation mutateLoadingOne!: (x: boolean) => void
+  @Mutation mutateLoadingOneIndex!: (x: number) => void
```
JazzarKarim, line 51: Is Mutations something that we'll be using moving on instead of actions? Actions will be used when there is an async call right?

src/store/Business/actions.ts
jonathan-longe: From memory, I didn't think that anything after a `resolve()` or `reject()` would be executed, but it's always the unknown, unknowns that surprise me.

src/components/Dashboard/FilingHistoryList/bodies/FutureEffectivePending.vue
```diff
@@ -24,49 +28,59 @@
...
       filingLabel: 'filing'
     }
   }
 
-  /** The future effective datetime of the subject filing. */
+  /** The effective date-time of this filing. */
   get effectiveDateTime (): string {
-    return (this.dateToPacificDateTime(this.filing.effectiveDate) || 'Unknown')
+    return this.filing.effectiveDate
+      ? DateUtilities.dateToPacificDateTime(new Date(this.filing.effectiveDate))
```
JazzarKarim, line 73: Why do we need to create a new date here? When we create a new date passing to it a parameter of type `FormattedDateTimeGmt`, it will return the correct output needed to convert to pacific date time?

src/store/Business/actions.ts
jonathan-longe: Looking at the docs, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise it looks like a return is not needed, but the resolve or reject logic should be contained, so no further code can be executed.

For example:
```javascript
function promiseGetWord(parityInfo) {
  return new Promise((resolve, reject) => {
    const { value, isOdd } = parityInfo;
    if (value >= THRESHOLD_A - 1) {
      reject(`Still too large: ${value}`);
    } else {
      parityInfo.wordEvenOdd = isOdd ? "odd" : "even";
      resolve(parityInfo);
    }
  });
}
```

src/App.vue
```diff
@@ -419,7 +412,7 @@ export default {
         AuthServices.fetchEntityInfo(this.getAuthApiUrl, this.businessId),
         this.loadBusinessInfo(),
         LegalServices.fetchTasks(this.businessId),
-        LegalServices.fetchFilings(this.businessId || this.tempRegNumber),
+        this.loadFilings(),
```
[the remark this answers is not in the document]
jonathan-longe: Would it be possible to call `this.loadFilings()` from the FilingHistoryList instead of here in App.vue?  It looks like FilingHistoryList is the only component that uses the data.

src/store/Business/actions.ts
severinbeauvais: > From memory, I didn't think that anything after a `resolve()` or `reject()` would be executed, but it's always the unknown, unknowns that surprise me.

That's my expectation, too, but I know what I saw. How about we leave the "extra" return in there just to be safe?

In your example, there is no other code after the reject or resolve, so 🤷‍♂️

src/components/Dashboard/StaffNotation.vue
```diff
@@ -240,8 +236,7 @@ export default class StaffNotation extends Vue {
   /** Prop for the scrollbar offset to be added. */
   @Prop() readonly addScrollbarOffset!: string
 
-  /** Prop for disabling the menu items. */
-  @Prop({ default: false }) readonly disabled!: boolean
+  @Action setFetchingDataSpinner!: ActionBindingIF
```
JazzarKarim, line 239: Can we use the Mutation instead?

src/components/Dashboard/FilingHistoryList/DocumentsList.vue
severinbeauvais: If it's only a matter of updating the store state then we can keep it simple and just call the mutations.

On the other hand, if it's more complicated then we would declare actions who in turn call the mutations.

src/interfaces/api-task-interface.ts
```diff
@@ -36,24 +36,25 @@ export interface ApiHeaderIF {
...
+  conversion?: any
   correction?: any
+  courtOrder?: any
```
JazzarKarim, line 48: Can we change it to CourtOrderIF instead of any?

src/components/Dashboard/FilingHistoryList/bodies/FutureEffectivePending.vue
severinbeauvais: `FormattedDateTimeGmt` is actually a string. I created this type to give more meaning to the string.
```
/**
  * Formatted date-time string with explicit GMT timezone (ie, UTC).
  * @example "Tue, 19 Apr 2022 16:40:18 GMT"
  */
export type FormattedDateTimeGmt = string
```

We need to create a new date because it needs to be formatted as a Pacific date.

severinbeauvais: The API provides several time/date formats. In this case we need to convert this:
image

to this:
image

src/App.vue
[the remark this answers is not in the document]
severinbeauvais: Yes but then we lose the asynchronicity of fetching all data at the same time (ie, Promise.all()).

It's true that the FHL is the only one to use the list items, but the Dashboard displays the count.

I'd rather not refactor App.vue in this ticket. Are you OK with that?

src/components/Dashboard/FilingHistoryList/bodies/FutureEffectivePending.vue
JazzarKarim: Oh okay. Thank you so much for the clarification Sev 👍

src/services/date-utilities.ts
```diff
@@ -350,4 +350,14 @@ export default class DateUtilities {
...
   }
+
+  /** Whether the subject date string is in the past. */
+  static isDatePast (date: FormattedDateTimeGmt): boolean {
+    return (new Date(date) <= new Date())
+  }
+
+  /** Whether the subject date string is in the future. */
+  static isDateFuture (date: FormattedDateTimeGmt): boolean {
+    return (new Date(date) > new Date())
```
JazzarKarim, line 361: Cool methods!

src/components/Dashboard/StaffNotation.vue
severinbeauvais: Yes! You'll see this in the next commit.

src/interfaces/api-task-interface.ts
severinbeauvais: I have done this (see next commit) but it doesn't help us in any way. First of all, this data from the API (which we don't control). Secondly, we don't get or set any properties from this object. I think I'll undo this change.

However, the ApiFilingIF contains a data.order object that could use this type... Except the definition in Filings UI:
image
is better than the definition in the shared repo:
image

Yes, this is a bit of tech debt that would be nice to resolve, but I'm calling it out of scope for this ticket. Are you OK with that?

src/components/Dashboard/FilingHistoryList/FilingTemplate.vue
```diff
@@ -0,0 +1,172 @@
...
+  <v-expansion-panel class="filing-template">
+    <v-expansion-panel-header>
+      <div class="item-header d-flex justify-space-between">
+        <div>
+          <h3 class="item-header-title">
+            <slot name="title">
+              <span>{{filing.displayName}}</span>
```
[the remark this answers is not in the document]
severinbeauvais: Believe it or not, I tried to not change every little thing I saw 😁 

I tried to leave existing code alone. Personally I prefer the extra whitespace, but the code works both ways.

src/interfaces/api-task-interface.ts
JazzarKarim: Okay, got it. Thank you for the explanation! We can resolve that in another ticket by having a uniform interface. I'll be resolving this conversation.

src/components/Dashboard/FilingHistoryList/FilingTemplate.vue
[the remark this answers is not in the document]
JazzarKarim: Okay Sev. Sorry, I didn't mean to be super nitpicky. I was just asking for future reference. Thank you!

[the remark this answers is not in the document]
severinbeauvais: In the future, you can do it either way and I promise I won't complain about it.

I still prefer the extra whitespace though 😁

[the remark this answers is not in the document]
JazzarKarim: It indeed does look nicer! I'll be going with the extra whitespace too then 😉

tests/unit/AlterationFiling.spec.ts
JazzarKarim: Didn't know we can do something like this. Might come in handy later. Nice!

src/components/Tombstone.vue
jonathan-longe: image

tests/unit/AlterationFiling.spec.ts
severinbeauvais: You can also do `xit`.

You can also do `describe.only` or `it.only`.

JazzarKarim: I just looked up what each one does. Cool stuff actually. Thanks a bunch Sev.

src/components/Tombstone.vue
severinbeauvais: I didn't want to overwrite any files in this PR so I created another PR with the same commits, but based on release-5.12.0 branch. Let me know if you like it better:

https://github.com/bcgov/business-filings-ui/pull/477

src/App.vue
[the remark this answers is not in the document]
jonathan-longe: Sure, I'll create a refactoring ticket to move call `loadFilings()` from Dashboard.vue.  As you point out, the dashboard needs the count.

src/components/Dashboard/FilingHistoryList.vue
ketaki-deodhar: This is awesome! It is going to be really easy to read/understand the different filings and conditionals for each one.

ketaki-deodhar:
@severinbeauvais, just noticed that there are 6 code-smells. Not sure how many are critical.

severinbeauvais:
> @severinbeauvais, just noticed that there are 6 code-smells. Not sure how many are critical.

I'll look at them.

severinbeauvais:
@jonathan-longe @ketaki-deodhar @JazzarKarim I now have 2 PRs for the same feature. They're in different branches so I can't work on both at the same time. Would you prefer to add your comments to this PR or the other one? Would you prefer me to make fixes to PR or the other one?

JazzarKarim:
> @jonathan-longe @ketaki-deodhar @JazzarKarim I now have 2 PRs for the same feature. They're in different branches so I can't work on both at the same time. Would you prefer to add your comments to this PR or the other one? Would you prefer me to make fixes to PR or the other one?

Whatever you prefer and makes your life easier Sev. Is one way or the other better for you?

JazzarKarim:
> > Whatever you prefer and makes your life easier Sev. Is one way or the other better for you?
> 
> I thought about it but I don't know which is easier for me or you.
> 
> On the one hand, all my comments and all your comments are in this PR.
> 
> On the other hand, the other PR contains a smaller diff (less changes from other PRs).
> 
> Either way it will be some effort for me to rebase and merge to main branch.
> 
> 🤷‍♂️
> 
> How about we stick with this PR? If you want a purer diff then look at the other one but add your comments here.

Makes sense. Alright, sounds good with me!

severinbeauvais:
> @severinbeauvais, just noticed that there are 6 code-smells. Not sure how many are critical.

Fixed in latest commit.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -1,373 +1,58 @@
...
+        impacts to business information.</span>
     </v-card>
 
-    <div class="scrollable-container">
-      <v-expansion-panels v-if="showHistoryPanel" v-model="panel">
-        <v-expansion-panel
-          class="align-items-top filing-history-item px-6 py-5"
-          v-for="(filing, index) in historyItems"
+    <div class="scrollable-container" v-if="showHistoryPanel">
+      <v-expansion-panels :value="getPanel">
```
jonathan-longe, line 37: What do you think about tucking these two lines into the `FilingTemplate`?  I think this makes it look cleaner and easier to read.  The `showHistoryPanel` then becomes redundant since if getFilings doesn't return any data, the `<component>` won't get rendered.

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
```diff
@@ -1,15 +1,19 @@
...
 
-    <p v-if="filing.courtOrderNumber">Court Order Number: {{filing.courtOrderNumber}}</p>
+    <p v-if="fileNumber">
+      Court Order Number: {{filing.courtOrderNumber}}
+    </p>
 
-    <p v-if="filing.isArrangement">Pursuant to a Plan of Arrangement</p>
+    <p v-if="hasEffectOfOrder">
+      Pursuant to a Plan of Arrangement
+    </p>
```
jonathan-longe, line 16: It looks like there's some duplication between FutureEffective and FutureEffectivePending.  What do you think about separating the elements and listing them in the parent?  Something like this:

```javascript
<FutureEffective>
  <h4>{{ subtitle }}</h4>
  <EffectiveDateTimeString />
  <CourtOrder />
  <PlanOfArrangement />
  <OneHourToProcess v-if="isFutureEffectivePending" />
  <WithDrawalNotice v-if="isFutureEffective" />
  <ContactInfo />
</FutureEffective>
```

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
```diff
@@ -27,52 +31,73 @@
...
   /** Data for the subject filing. */
   get _ (): any {
-    if (this.filing.isFutureEffectiveIa) {
+    if (EnumUtilities.isTypeIncorporationApplication(this.filing)) {
       return {
         subtitle: 'Future Effective Incorporation Date',
         filingLabel: 'incorporation',
         filingTitle: 'Incorporation Application'
       }
     }
```
jonathan-longe, line 55: Hey, why we are we using `get _()` to determine the subtitle, filingLabel and filingTitle, when we've used `<component :is="" />' to separate the filings by filing type from the dashboard.  Isn't the filing type known at this point?

src/components/Dashboard/FilingHistoryList/subtitles/FutureEffectivePaid.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <div class="future-effective-paid">
+    <span v-if="isTypeIncorporationApplication">FUTURE EFFECTIVE INCORPORATION</span>
+    <span v-else-if="isTypeAlteration">FUTURE EFFECTIVE ALTERATION</span>
+    <span v-else-if="isTypeDissolutionVoluntary">FUTURE EFFECTIVE DISSOLUTION</span>
+    <span v-else>FUTURE EFFECTIVE FILING</span>
+    <span class="vert-pipe" />
+    <span>PAID <FiledLabel :filing="filing" /></span>
```
jonathan-longe, line 8: It seems like this component is called from a filing component that already "knows" the filing type.  What do you think of either passing in the title as a prop or as a slot with the default as "Future Effective Filing"?  This way could eliminate the need for this `if-then-else` logic.

src/components/Dashboard/FilingHistoryList/subtitles/FutureEffectiveFiledAndPendingCoa.vue
```diff
@@ -0,0 +1,55 @@
...
+export default class FutureEffectiveFiledAndPendingCoa extends Vue {
+  @Prop({ required: true }) readonly filing!: ApiFilingIF
+  @Prop({ required: true }) readonly index!: number
+
+  /** The effective date-time of this filing. */
+  get effectiveDateTime (): string {
+    return this.filing.effectiveDate
+      ? DateUtilities.dateToPacificDateTime(new Date(this.filing.effectiveDate))
+      : '[unknown]'
+  }
```
jonathan-longe, line 36: To reduce duplication, what do you think about creating a store getter to replace this method?

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I think it is the FilingHistoryList.vue's job to display the list vs the no-results message.

Also I don't see how I can move the expansion panels component inside the template, since the template is iterated upon (many copies of "panel" inside 1 "panels").

What am I not seeing?

src/store/FilingHistoryList/actions.ts
jonathan-longe: > I don't think businessId or tempRegNumber need to be parameterized, since they are set once when the app starts and don't change thereafter.

Didn't we groom bcgov/entity#15657 to pass businessId or the tempRegNumber as parameters from VueRouter?  I thought we agreed, it would be better to follow best practices than a roll-your-own approach?

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
severinbeauvais: There's actually more duplication than you noted above, but I really don't want to break up things in components with just a few lines. How about a new ticket to tackle this (as part of future work probably)?

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
severinbeauvais: Yes, the filing type is know in this partial. This is actually previous code that I didn't change. Believe it or not, I tried to not change every single thing I touched. It was really hard. I was very tempted.

In order to implement what you suggest, we could pass props (or use slots). However, I've reached my limit for things I want to change in this ticket (other than any bugs I created -- fingers crossed!).

src/components/Dashboard/FilingHistoryList.vue
jonathan-longe: Whoops, that's my mistake, you're absolutely right.  That makes no sense!

src/components/Dashboard/FilingHistoryList/subtitles/FutureEffectivePaid.vue
severinbeauvais: You're right. It's the same scenario as FutureEffective above.

Again, I left the existing code. My plan for this ticket was to clean up the bulk of the pasta in the FilingHistoryList template and script -- to separate filing types and to make it easier to add new ones. I wasn't planning on re-writing all sub-components (though I did many). How about a new ticket to handle this comment and the one above?

src/components/Dashboard/FilingHistoryList/subtitles/FutureEffectiveFiledAndPendingCoa.vue
severinbeauvais: We'd have to pass the filing itself, which is not so much a store getter as a utility function... in a mixin? I think doing that would be as ugly as the current code. How much do you want this, in this ticket?

src/App.vue
```diff
@@ -430,7 +424,7 @@ export default {
...
```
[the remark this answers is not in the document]
severinbeauvais: Yes, good point. I'll update this now.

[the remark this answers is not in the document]
severinbeauvais: Done.

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
jonathan-longe: > However, I've reached my limit for things I want to change in this ticket...

Got it.  PR comments > /dev/null  :-)

jonathan-longe (approved):
Huge amount of work here! Thanks @severinbeauvais

src/store/FilingHistoryList/actions.ts
severinbeauvais: Yes, we groomed that ticket for Edit UI, and I agree that we should look at that approach.

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
severinbeauvais: Haha, no, keep the comments coming, but don't add extra work :)

(I've been buried under several linked/blocked features/PRs for 3 sprints now. I'm almost at the surface and desperately need to breathe a little ;) )

src/App.vue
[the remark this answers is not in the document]
severinbeauvais: Good to resolve this conversation?

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -737,218 +173,23 @@ export default class FilingHistoryList extends Vue {
...
-    // load data initially and when filings list changes
-    this.loadData()
+    // if needed, highlight a specific filing
+    if (this.highlightId) {
+      const index = this.getFilings.findIndex(f => (f.filingId === this.highlightId))
+      if (index >= 0) { // safety check
+        this.toggleFilingHistoryItem(index, this.getFilings[index])
+      }
+    }
   }
```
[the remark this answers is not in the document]
severinbeauvais: Good to resolve this conversation?

src/components/Dashboard/StaffNotation.vue
severinbeauvais: It's gone now!

src/main.ts
severinbeauvais: It's gone now (from this PR; obviously it's still in the main branch).

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: It's gone now!

src/store/FilingHistoryList/actions.ts
severinbeauvais: This action method is now passed "id" instead of getting it from session storage. Good to resolve this conversation?

severinbeauvais:
@ketaki-deodhar @JazzarKarim Do you have any further comments or concerns? If not, can I get your review approval, please?

JazzarKarim (approved):
Lots and lots of learning for me whilst I was reviewing this PR. Thanks for the opportunity to review Sev! Awesome stuff 👍

severinbeauvais:
I'm going to merge this now. If there are any follow-up issues then I'll create a new PR (or ticket). FYI

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
jonathan-longe: Done:  bcgov/entity#15822

src/components/Dashboard/FilingHistoryList/bodies/FutureEffective.vue
jonathan-longe: Yes, I totally get it.  I'll create refactoring tickets for issues that can't be resolved in the PR.

PR #468, opened 2023-03-21
14670 Added ledger component for restoration extension filings

severinbeauvais:
*Issue #:* bcgov/entity#14670

*Description of changes:*

First commit:
- app version = 5.11.6
- imported latest shared corp-type-module
- imported latest Vuetify
- replaced some getLegalName with getEntityName
- added some getLegalName fallback strings
- fixed some typing errors
- added handling for Restoration Extension filings
- added Limited Restoration Extension Filing component
- renamed isAdminFreeze -> isAdminFrozen
- changed some Enum Mixin calls to Enum Utilities
- deleted obsolete Enum Mixin methods
- cleaned up EntityInfo first and second lines
- moved EntityInfo tombstone data to new component
- renamed some IDs
- added some getters to get rid of `@State` usage
- deleted some unneeded getter refs
- created Tombstone component
- added getEntityName getter
- moved some computed to store getters
- created/updated/fixed unit tests
- misc cleanup

Second commit:
- moved Entity Info header to new component
- moved Entity Info menu to new component
- renamed Tombstone -> EntityDefinitions
- misc cleanup
- fixed a getter name
- updated/created unit test suites for affected components


src/views/Dashboard.vue
```diff
@@ -114,50 +117,52 @@
               </v-card>
             </section>
 
-            <!-- Current Directors-->
-            <section v-if="!isSoleProp && !isPartnership">
+            <!-- Proprietor / Partners -->
+            <section v-if="isSoleProp || isPartnership">
```
severinbeauvais, line 121: I just moved the "if True" code above the "else" code.

src/components/EntityInfo.vue
```diff
@@ -331,27 +254,22 @@ export default class EntityInfo extends Vue {
     this.emitConfirmDissolution()
   }
 
-  /** Navigates to the Auth UI to update business profile. */
-  protected editBusinessProfile (): void {
-    navigate(this.getBusinessProfileUrl)
-  }
```
severinbeauvais, line 337: This is moved to Tombstone component.

src/components/EntityInfo.vue
```diff
@@ -406,32 +318,6 @@ menu > span + span {
...
-  display: none;
-}
-
-#change-email-button,
-#change-phone-button {
-  height: 1.125rem;
-  padding: 0.25rem 0.5rem;
-  margin-top: -0.125rem;
-  margin-left: 0.125rem;
-}
```
severinbeauvais, line 433: This is moved to Tombstone component.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -686,6 +710,12 @@ export default class FilingHistoryList extends Vue {
         item.expiry = filing.data.restoration?.expiry
       }
 
+      // add properties for limited restoration extensions
+      if (EnumUtilities.isTypeRestorationExtension(filing)) {
+        item.isTypeLimitedRestorationExtension = true
+        item.expiry = filing.data.restoration?.expiry
+      }
+
```
jonathan-longe, line 718: Would it be possible to add these attributes inside the `LimitedRestorationExtension` filing history component?

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
```diff
@@ -0,0 +1,47 @@
...
+import { Component, Prop } from 'vue-property-decorator'
+import { ContactInfo } from '@/components/common'
+import { HistoryItemIF } from '@/interfaces'
+import { DateUtilities } from '@/services'
+
+@Component({
+  components: { ContactInfo }
+})
+export default class LimitedRestorationExtensionFiling extends Vue {
+  /** The subject filing. */
```
jonathan-longe, line 28: Since this is a new component, what do you think about using the more the Options API rather than the Class API so it's natively supported by Vue3.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -76,7 +76,7 @@
             <v-list-item
               data-type="registrars-order"
               @click="showRegistrarsOrderDialog()"
-              :disabled="disabled || isAdminFreeze"
+              :disabled="disabled || isAdminFrozen"
```
jonathan-longe, line 79: Is this change related to the ticket?

src/components/EntityInfo.vue
```diff
@@ -4,40 +4,40 @@
...
-                  small label text-color="white"
-                >
-                  LIMITED RESTORATION
-                </v-chip>
-              </span>
+            <template v-if="!!businessId">
+              <!-- First line -->
+              <div id="entity-legal-name" aria-label="Entity Legal Name">
+                {{ getEntityName || 'Unknown Name' }}
+              </div>
```
jonathan-longe, line 11: What do you think of creating two components, `EntityInfoBusinessId` and `EntityInfoTempRegId`?  It feels like we're mixing concerns by combining the two.

src/components/EntityInfo.vue
```diff
@@ -141,62 +141,7 @@
...
-              <dd id="entity-business-phone" class="cursor-pointer" @click="editBusinessProfile()">
-                <span>{{fullPhoneNumber || 'Not Available'}}</span>
-                <v-btn id="change-phone-button" small text color="primary">
-                  <v-icon small>mdi-pencil</v-icon>
-                  <span>Change</span>
-                </v-btn>
-              </dd>
-            </template>
-          </dl>
+          <Tombstone />
```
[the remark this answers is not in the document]
jonathan-longe: I think it's recommended to use two words for component names.  What about `EntityKeyFacts`?  See:  https://v2.vuejs.org/v2/style-guide/?redirect=true#Multi-word-component-names-essential

src/components/Tombstone.vue
```diff
@@ -0,0 +1,171 @@
...
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component } from 'vue-property-decorator'
+import { Getter } from 'vuex-class'
+import { navigate } from '@/utils'
+import { DateUtilities } from '@/services'
+
+@Component({})
+export default class Tombstone extends Vue {
```
jonathan-longe, line 67: What about using the Options API so it's compatible with Vue3?

src/components/Tombstone.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <dl id="tombstone">
+    <!-- Registration Date -->
+    <template v-if="!!registrationDate">
+      <dt class="mr-2">Registration Date:</dt>
+      <dd id="registration-date">{{ registrationDate }}</dd>
+    </template>
```
jonathan-longe, line 7: What do you think about creating a component for the items?  It looks like you have only two variants, a display version and an edit version.   The benefit is that if the business needs to modify the key fact displayed all we need to do is change the data supplied.  We don't have to add an if-then.

src/components/Tombstone.vue
```diff
@@ -0,0 +1,171 @@
...
+  @Getter getFoundingDate!: Date
+  @Getter getFullPhoneNumber!: string
+  @Getter getIdentifier!: number
+  @Getter getNameRequest!: any
+  @Getter isFirm!: boolean
+
+  /** The Business ID string, if available. */
+  get businessId (): string {
+    return sessionStorage.getItem('BUSINESS_ID')
+  }
```
jonathan-longe, line 80: What do you think about passing the `businessId` in as a prop?  If we eventually want to have the businessId passed to the view as route parameter, this makes this refactoring easier in the future.

src/mixins/enum-mixin.ts
```diff
@@ -1,32 +1,12 @@
...
-  /** DEPRECATED Returns True if item status is Corrected. */
-  isStatusCorrected (item: any): boolean {
-    return (item.status === FilingStatus.CORRECTED)
-  }
-
-  /** DEPRECATED Returns True if item status is Deleted. */
-  isStatusDeleted (item: any): boolean {
-    return (item.status === FilingStatus.DELETED)
-  }
-
```
jonathan-longe, line 29: Nice!  Get rid of those deprecated methods.

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
severinbeauvais: A quick Google search suggests Class API is well supported in Vue3:
https://github.com/vuejs/vue-class-component/issues/402
https://levelup.gitconnected.com/from-vue-class-component-to-composition-api-ef3c3dd5fdda
https://medium.com/@takeit_/vue-3-composition-api-with-typescript-class-components-dbe7cc9bb468

In THIS PR, I am following existing code. In my next (refactoring) PR, I will look at this and change it over if it's not too much of a PITA, ok?

src/components/Dashboard/StaffNotation.vue
severinbeauvais: No.

`adminFreeze` is the name of the variable, whereas the getter returns the state.

src/components/EntityInfo.vue
severinbeauvais: I'll look into this but it is a bit beyond the scope of this PR.

src/components/Tombstone.vue
severinbeauvais: See other comment. I want to move forward this way for now. This shouldn't be much additional work to change later if we change all other components over at the same time.

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
jonathan-longe: Apologies, that's my mistake, when we discussed this issue previously (see below) I shared a link that I misread to read the Class API support had been abandoned in Vue3.  

image

src/components/Tombstone.vue
jonathan-longe: You're correct, the Class API is supported in Vue3.  Please ignore the suggestion.

src/components/Tombstone.vue
severinbeauvais: There is no edit version here; only items that may or may not apply to the currently loaded entity.

Sorry if I don't understand your suggestion. What do you mean?

src/components/Tombstone.vue
severinbeauvais: Good idea!

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+  <div v-if="filing" class="limited-restoration-extension-filing body-2">
+    <h4>Extension of Limited Restoration</h4>
+
+    <p>
+      The period of restoration was successfuly extended and is active
+      <strong>until {{ expiryDateFriendly }}</strong>. At the end of the extended limited
```
JazzarKarim, line 7: Nice! This will help my extension section ticket! 😄

JazzarKarim (approved):
Lots and LOTS of learning for me. Especially since it's Filings UI. Awesome stuff!

ketaki-deodhar (approved):
Thanks for the opportunity to review and learn!

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
severinbeauvais: Are you OK if I don't change this component to Options API, and if so can you please resolve this conversation? TIA

src/components/Tombstone.vue
severinbeauvais: Are you OK to resolve this conversation?

src/components/EntityInfo.vue
severinbeauvais: I'm working on this (and old "Tombstone" sub-component) atm. I think you'll like what I've done 😁

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Are you OK with leaving this until Phase 2 of this work (next PR)?

src/components/EntityInfo.vue
severinbeauvais: Please see new commit and resolve this conversation if you are satisfied.

src/components/Tombstone.vue
severinbeauvais: Please see new commit and resolve this conversation if you are satisfied.

src/components/Dashboard/StaffNotation.vue
jonathan-longe: Yes, you bet, just clarifying to make sure I wasn't missing something important.

src/components/EntityInfo/EntityHeader.vue
```diff
@@ -0,0 +1,105 @@
...
+          <v-chip class="primary mt-n1 ml-3 pointer-events-none font-weight-bold"
+            small label text-color="white">LIMITED RESTORATION</v-chip>
+        </span>
+
+        <span id="authorized-to-continue-out" v-if="isAuthorizedToContinueOut">
+          <v-chip class="primary mt-n1 ml-3 pointer-events-none font-weight-bold"
+            small label text-color="white">AUTHORIZED TO CONTINUE OUT</v-chip>
+        </span>
+      </div>
+    </template>
```
jonathan-longe, line 24: I know this may seem like a pain, but can we separate the code for entities with a `businessId` from entities with a `tempRegNumber` into their own components?  I eventually would like to have separate dashboard routes for entities with a `businessID` and entities with a `tempRegNumber`.  Separating these concerns will make this easier.

src/components/EntityInfo/EntityHeader.vue
```diff
@@ -0,0 +1,105 @@
...
+      <!-- Title -->
+      <div id="ia-reg-name" aria-label="Incorporation Application or Registration Entity Name">
+        {{ getEntityName || 'Unknown Name' }}
+      </div>
+
+      <!-- Subtitle -->
+      <div id="ia-reg-description" aria-label="Incorporation Application or Registration Description">
+        {{ iaRegDescription }}
+      </div>
+    </template>
```
jonathan-longe, line 36: Here's the template for the EntityHeaderTempReg

src/components/EntityInfo/EntityDefinitions.vue
```diff
@@ -0,0 +1,168 @@
+<template>
+  <dl id="entity-definitions">
+    <!-- Registration Date -->
+    <template v-if="!!registrationDate">
+      <dt class="mr-2">Registration Date:</dt>
+      <dd id="registration-date">{{ registrationDate }}</dd>
+    </template>
```
jonathan-longe, line 7: Apologies for my cryptic comment earlier!  What do you think of making the following a component?

```javascript
  <template>
      <dt class="mr-2">{{ title }}:</dt>
      <dd :id="id">{{ value }}</dd>
    </template>
```

And this a component too?

```javascript
   <template>
      <dt class="mr-2">{{ title }}:</dt>
      <dd :id="id" class="cursor-pointer" @click="editBusinessProfile()">
        <span>{{ value }}</span>
        <v-btn :id="'change' + id + 'button'" small text color="primary">
          <v-icon small>mdi-pencil</v-icon>
          <span>Change</span>
        </v-btn>
      </dd>
    </template>
```

I know this may seem crazy to create tiny components like this but I think you'll see it makes our code more flexible.  By displaying these components by iterating over an array of data, it makes it very easy to change in the future if the business changes the requirements.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -0,0 +1,186 @@
...
+
+    <span v-if="isHistorical">
+      <v-chip
+        id="historical-chip"
+        class="primary mt-n1 ml-4 pointer-events-none" small label text-color="white"
+      >
+        <strong>HISTORICAL</strong>
+      </v-chip>
+      <span class="font-14 mx-3">{{ getReasonText || 'Unknown Reason' }}</span>
+    </span>
```
jonathan-longe, line 20: This looks like a separate concern that deserves it's own component.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -0,0 +1,186 @@
+<template>
+  <menu id="entity-menu">
+    <!-- Staff Comments -->
+    <span v-if="!!businessId && isAllowed(AllowableActions.ADD_STAFF_COMMENT)">
+      <StaffComments
+        :axios="axios"
+        :businessId="businessId"
+        :maxLength="2000"
+      />
+    </span>
```
jonathan-longe, line 10: Rather than writing conditionals for each allowable action, would it be possible to iterate over the allowable actions returned by the API and then use a `<component :is="">` to render these menu items dynamically?   It seems to me that this would make our code easier to extend and maintain.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -0,0 +1,186 @@
+<template>
+  <menu id="entity-menu">
```
jonathan-longe, line 2: What do you think about either moving `<menu>` back to the parent or creating a wrapper with a slot so the subcomponents can be listed like we were thinking about doing with the `YourCompany` component?

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -0,0 +1,186 @@
...
+    <!-- Download Business Summary -->
+    <span v-if="isAllowed(AllowableActions.DOWNLOAD_BUSINESS_SUMMARY)">
+      <v-tooltip top content-class="top-tooltip" transition="fade-transition">
+        <template v-slot:activator="{ on }">
+          <v-btn
+            small text color="primary"
+            id="download-summary-button"
+            @click="emitDownloadBusinessSummary()"
+            v-on="on"
+          >
```
jonathan-longe, line 73: When this button is clicked, it:
- calls a method, `emitDownloadBusinessSummary()`
- which emits an event called, `downloadBusinessSummary`
- which is received by the parent, "EntityInfo"
- which calls another method called, `emitDownloadBusinessSummary`
- which emits another method called `downloadBusinessSummary`
- which is received by the parent, "App.vue"
- which calls another method called, `downloadBusinessSummary`
- which calls the API and downloads the requested document

Would it be possible to call a store action instead or, if you don't have time, create a refactoring ticket that explains the issue and describes a resolution?

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -0,0 +1,186 @@
...
+      <span v-if="isAllowed(AllowableActions.DISSOLVE_COMPANY)">
+        <v-tooltip top content-class="top-tooltip" transition="fade-transition">
+          <template v-slot:activator="{ on }">
+            <v-btn
+              small text color="primary"
+              id="dissolution-button"
+              :disabled="hasBlocker"
+              @click="promptDissolve()"
+              v-on="on"
+            >
```
jonathan-longe, line 53: This button and several other buttons in this component appear to also have a long chains of emits that are passed back to the parent and then the grandparent.  Can we either reduce the length of the chains or create tech debt tickets to refactor this?

src/components/EntityInfo/EntityHeader.vue
severinbeauvais: I'd rather not, at this time anyway. Separate routes sound a good idea for the future, but it's still fair to have separate views use common components that check a variable to display A vs B. Also, I like the fact that the Entity Info's header content is all in one file. Also, I don't want to assume that businesses vs pre-businesses (temp reg number) will continue to be handled separately in the current way. Also, this isn't a complex component to break apart in the future if the need arises (the need isn't here yet). Finally, there are many higher-priority, higher-gain things to work first/instead.

Here's an idea of what we'll be up against if we try to split up businesses from pre-businesses -- the A vs B code is in a lot of places so that we reuse components.

image

image

src/components/EntityInfo/EntityHeader.vue
severinbeauvais: If you don't want to see this template when you're in this file, code-fold it.

At the moment, I prefer to keep like content together instead of managing more component files. Also see my previous comment on this file.

src/components/EntityInfo/EntityDefinitions.vue
severinbeauvais: We don't need that flexibility yet. Let's change it when/if we need to.

Right now I prefer to keep this entire "description list" together.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: That's a good idea but I doubt it: the possible actions (and whether they're allowed or not) are sprinkled into different places in the code/components. It boils down to our desire to show a consistent page with various actions but those actions are in different places (header, right-side components, staff filings dropdown, etc).

This could be investigated further, but IMO it's not the highest priority in this UI, not even in the top 3-4 (ie, behind Filing History, Todo List, App.vue, local filings, and more).

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/15726

severinbeauvais:
@jonathan-longe Thanks for all your forward-looking comments -- I really appreciate (and support) your efforts to move us towards a more flexible architecture. However, this ticket/PR was only meant to add a specific functionality using the existing architecture -- I am using ticket 15692 to refactor a variety of things, but even then the scope needs to be limited so we can balance cleanup vs new features.

I heard we have support to spend more time up front on this (pay now instead of pay later) but I'm trying to be careful to be concise in approved tickets, and creating other tickets for future work (or not creating tickets if the value isn't clear today -- they can be created later). For example, THIS ticket was estimated as a 3 and it's already used up 5 worth of time. And 15692 was also estimated as a 3 and maybe should be a 5 (due to revamped unit tests and QA). I'm OK with more tickets but I don't want to blow up current tickets.

Please continue your thorough reviews and spring of ideas, and please don't be upset if I push back a little :)

severinbeauvais:
I'm not going to overrule you and merge this ticket as is. Please let me know if I have your approval (and if you're comfortable that outstanding issues are recorded in other tickets).

jonathan-longe:
Understood @severinbeauvais I definitely don't want to hold you up or become a PITA :-)  Please go ahead and merge the code you're comfortable merging.

PR #478, opened 2023-03-29
15286 Refactor to use allowed actions from API

severinbeauvais:
*Issue #:* bcgov/entity#15286

**This is the same as PR https://github.com/bcgov/business-filings-ui/pull/463 except it's rebased to the latest main branch. It has already been reviewed and is ready to merge.**

*Description of changes:*
- app version = 6.1.0 (since this is a significant, but not major, change)
- imported latest shared Corp Type Module
- updated/added Allowable Actions
- deleted obsolete hasBlockerTask/hasBlockerFiling code (components, store, etc)
- updated EntityInfo, Dashboard, etc to use new isAllowed() method (small tweaks to rendered vs disabled)
- updated isAllowed() to use data from Legal API instead of local business logic
- updated unit tests


src/components/Dashboard/StaffNotation.vue
```diff
@@ -97,73 +97,67 @@
...
-                <v-list-item-title>
-                  <span class="app-blue">Consent to Continuation Out</span>
-                </v-list-item-title>
-              </v-list-item>
-            </template>
+            <v-list-item
+              v-if="!isHistorical"
+              data-type="administrative-dissolution"
+              @click="showAdministrativeDissolutionDialog()"
+              :disabled="!isAllowed(AllowableActions.ADMINISTRATIVE_DISSOLUTION)"
```
severinbeauvais, line 111: This item always displays as long as the business is not already historical, but it's enabled only when allowed.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -97,73 +97,67 @@
...
+            >
```
severinbeauvais, line 112: This item always displays as long as the business is not already historical, but it's enabled only when allowed.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -19,50 +19,48 @@
...
-              <span class="font-13 ml-1">Dissolve this Business</span>
-            </v-btn>
-          </template>
-          Dissolving the business will make this business historical
-          and it will be struck from the corporate registry.
-        </v-tooltip>
-      </span>
-    </template>
+    <!-- View and Change Business Information -->
+    <span v-if="isBusiness && !isHistorical">
```
severinbeauvais, line 23: Here I took a different approach: only show the action when it's appropriate but enable it only if allowed.

Ditto for Dissolve Business below.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -19,50 +19,48 @@
...
+          </v-btn>
+        </template>
+        Dissolving the business will make this business historical
+        and it will be struck from the corporate registry.
+      </v-tooltip>
+    </span>
 
     <!-- Download Business Summary -->
-    <span v-if="isAllowed(AllowableActions.DOWNLOAD_BUSINESS_SUMMARY)">
+    <span v-if="isAllowed(AllowableActions.BUSINESS_SUMMARY)">
```
severinbeauvais, line 63: This is a show/no-show item (unlike above where it's show/no-show and enabled/disabled).

The API should only allow these when appropriate.

PR #479, opened 2023-03-31
Fixed the content path issue.

pwei1018:
*Issue #:* [15499](https://github.com/bcgov/entity/issues/15499)

*Description of changes:*
The current VUE_APP_PATH of GCP CD setting cause a null value exception after deploy to firebase. It must be changed to "/" value to avoid exception.

Also, open a ticket to investigate BASE_URL issue. https://github.com/bcgov/entity/issues/15817

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "6.1.0",
+  "version": "6.1.1",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "6.1.0",
+  "version": "6.1.1",
```
severinbeauvais, line 3: Keep an eye on this as I hope to merge a PR with v6.2.0 this morning.

severinbeauvais (approved):
LGTM. Did you run it locally to verify that a business loaded up, and a local filing (eg, BEN or CP AR/COA/COD) could be filed, fine?

PR #488, opened 2023-04-21
15892 Director & Address change correction

JazzarKarim:
*Issue #:* bcgov/entity#15892 & bcgov/entity#15893

*Description of changes:*
- Open File Correction dialog when filing a correction for Director & Address change
- Redirect to Edit UI
- Save and resume redirects to Edit UI also
- The ability to correct is dependent on `supported-correction-entities` feature flag. If entity is not supported in a specific environment, the "File a Correction" button will be disabled.

For reference:

https://user-images.githubusercontent.com/122301442/233743788-a1a235ee-b5e3-411f-8c4b-5effd03fdb88.mp4

### DIFF
--- a/src/components/Dashboard/FilingHistoryList/HeaderActions.vue
+++ b/src/components/Dashboard/FilingHistoryList/HeaderActions.vue
@@ -115,6 +115,8 @@ export default class HeaderActions extends Vue {
     conditions[4] = () => (EnumUtilities.isTypeChangeOfRegistration(this.filing) && !this.isFirm)
     conditions[5] = () => (EnumUtilities.isTypeCorrection(this.filing) && !this.isFirm && !this.isBenBcCccUlc)
     conditions[6] = () => (EnumUtilities.isTypeRegistration(this.filing) && !this.isFirm)
+    conditions[7] = () => (EnumUtilities.isTypeChangeOfAddress(this.filing) && !this.isBenBcCccUlc)
+    conditions[8] = () => (EnumUtilities.isTypeChangeOfDirectors(this.filing) && !this.isBenBcCccUlc)
 
     // check if any condition is True
     return conditions.some(condition => condition())
--- a/src/components/Dashboard/FilingHistoryList/HeaderActions.vue
+++ b/src/components/Dashboard/FilingHistoryList/HeaderActions.vue
@@ -115,8 +115,6 @@ export default class HeaderActions extends Vue {
     conditions[4] = () => (EnumUtilities.isTypeChangeOfRegistration(this.filing) && !this.isFirm)
     conditions[5] = () => (EnumUtilities.isTypeCorrection(this.filing) && !this.isFirm && !this.isBenBcCccUlc)
     conditions[6] = () => (EnumUtilities.isTypeRegistration(this.filing) && !this.isFirm)
-    conditions[7] = () => (EnumUtilities.isTypeChangeOfAddress(this.filing) && !this.isBenBcCccUlc)
-    conditions[8] = () => (EnumUtilities.isTypeChangeOfDirectors(this.filing) && !this.isBenBcCccUlc)
 
     // check if any condition is True
     return conditions.some(condition => condition())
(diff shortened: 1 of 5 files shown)


severinbeauvais:
> * The ability to correct is dependent on `supported-correction-entities` feature flag. If entity is not supported in a specific environment, the "File a Correction" button will be disabled.

What happens for CP corrections?

JazzarKarim:
> > * The ability to correct is dependent on `supported-correction-entities` feature flag. If entity is not supported in a specific environment, the "File a Correction" button will be disabled.
> 
> What happens for CP corrections?

Please see the latest commit Sev. It should be OK now.

Screenshot (63)

severinbeauvais:
> Please see the latest commit Sev. It should be OK now.

Please ask Mihai whether Coops should continue to use the local correction filing.

src/components/Dashboard/FilingHistoryList/HeaderActions.vue
```diff
@@ -125,18 +125,32 @@ export default class HeaderActions extends Vue {
...
         this.mutateCurrentFiling(filing)
         this.mutateFileCorrectionDialog(true)
         break
 
-      case FilingTypes.ANNUAL_REPORT:
       case FilingTypes.CHANGE_OF_ADDRESS:
       case FilingTypes.CHANGE_OF_DIRECTORS:
+        if (this.isBenBcCccUlc) {
+          // correction via Edit UI if current type is BC, CC, ULC, or BEN
+          this.mutateCurrentFiling(filing)
```
JazzarKarim, line 141: I know there's a bit of code duplication here but I couldn't think of a cleaner way to achieve this.

severinbeauvais: This doesn't bother me.

src/components/Dashboard/TodoList.vue
```diff
@@ -1403,18 +1403,33 @@ export default class TodoList extends Vue {
...
             break
 
+          case FilingNames.CHANGE_OF_ADDRESS:
+          case FilingNames.CHANGE_OF_DIRECTORS:
+            if (this.isBenBcCccUlc) {
+              // resume correction via Edit UI if current type is BC, CC, ULC, or BEN
+              const correctionUrl = `${this.getEditUrl}${this.getIdentifier}/correction/?correction-id=${item.filingId}`
+              navigate(correctionUrl)
+              break
+            } else {
```
severinbeauvais, line 1422: This one's messier and does bother me 😁 

How about implementing a local function that you can call, eg.
```
/** Resumes a draft filing. */
protected doResumeFiling (item: TodoItemIF): void {
  ...
  case XXX
    GoToEditUi()
    break
  case YYY
    if (...) {
      GoToEditUi()
      break
    }

  function GoToEditUi (): void {
     ...
  }
}
```

JazzarKarim: Incredible idea Sev. I will be implementing that, thanks a bunch!

JazzarKarim:
> I think you'll find that Coops are still not able to do local corrections due to the FF check in the Allowable Actions mixin. Probably you'll have to refactor that FF check out of the mixin and move it to HeaderActions instead.

Are you sure Sev? I've tested in local and I'm getting the expected output currently.

When I try to File a correction in Coops, I'm going to the local correction page.
When I try to File a correction in BCs, I'm going to the Edit UI correction page.

The FF supports CPs:
correction local

severinbeauvais:
> The FF supports CPs: correction local

I was concerned that this would break in Test or Prod because they have a different feature flag (list of supported entities), but as it turns out, both Test and Prod allow Coops, so this should be fine.

severinbeauvais:
Please fix broken unit tests.

src/components/Dashboard/TodoList.vue
```diff
@@ -1520,6 +1510,20 @@ export default class TodoList extends Vue {
...
+
+    function navigateToCorrectionEditUi (editUrl: string, identifier: string): void {
+      // resume correction via Edit UI
+      const correctionUrl = `${editUrl}${identifier}/correction/?correction-id=${item.filingId}`
+      navigate(correctionUrl)
+    }
+
+    function routeToLocalCorrection (router: VueRouter): void {
+      // resume local correction
+      router.push({
```
severinbeauvais, line 1522: Can you use `this$router` here to simplify things?

JazzarKarim: I'm unable to access the router inside the function (reading properties of undefined).

severinbeauvais: OK, fine.

JazzarKarim:
@severinbeauvais Should I add/fix anything else Sev?

severinbeauvais:
> @severinbeauvais Should I add/fix anything else Sev?

LGTM but please get some other reviews.

ketaki-deodhar (approved):
Great work!

severinbeauvais:
Please re-run the failed linking check before merging.

PR #491, opened 2023-04-26
15051 Second update to local filings to respond to UX observations

JazzarKarim:

*Description of changes:*
- Yui's comment: https://github.com/bcgov/entity/issues/15051#issuecomment-1523874685
- Fixed allignment
- Fixed text colors
- Fixed validation trigger
screenshot (2)

### DIFF
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -16,19 +16,21 @@
           />
         </div>
       </div>
-      <v-checkbox
-        :value="isCertified"
-        @change="emitIsCertified($event)"
-      >
-        <template v-slot:label>
-          <div class="certify-stmt">
-            I, <strong>{{trimmedCertifiedBy || '[Legal Name]'}}</strong>, certify that I have relevant knowledge
-            of the {{entityDisplay || 'association'}} and that I am authorized to make this filing.
-          </div>
-        </template>
-      </v-checkbox>
-      <p class="certify-clause">Date: {{getCurrentDate}}</p>
-      <p class="certify-clause">{{message}}</p>
+      <div class="certify-statement-section">
+        <v-checkbox
+          :value="isCertified"
+          @change="emitIsCertified($event)"
+        >
+          <template v-slot:label>
+            <div class="certify-stmt">
+              I, <strong>{{trimmedCertifiedBy || '[Legal Name]'}}</strong>, certify that I have relevant knowledge
+              of the {{entityDisplay || 'association'}} and that I am authorized to make this filing.
+            </div>
+          </template>
+        </v-checkbox>
+        <p class="certify-clause">Date: {{getCurrentDate}}</p>
+        <p class="certify-clause">{{message}}</p>
+      </div>
     </div>
   </v-card>
 </template>
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -1,22 +1,20 @@
 <template>
   <v-card flat>
-    <div class="certify-container">
-      <div class="certifiedby-container">
-        <label>Legal Name</label>
-        <div class="value certifiedby">
-          <v-text-field
-            filled
-            persistent-hint
-            id="certified-by-textfield"
-            label="Person's legal name"
-            hint="Legal name of authorized person"
-            :value="certifiedBy"
-            :rules="[ v => !!v || 'A person\'s legal name is required.']"
-            @input="emitCertifiedBy($event)"
-          />
-        </div>
-      </div>
-      <div class="certify-statement-section">
+    <v-row no-gutters class="pl-4 pr-4 pt-4">
+      <v-col cols="12" sm="3">
+        <label class="title-label">Legal Name</label>
+      </v-col>
+      <v-col cols="12" sm="9">
+        <v-text-field
+          filled
+          persistent-hint
+          id="certified-by-textfield"
+          label="Person's legal name"
+          hint="Legal name of authorized person"
+          :value="certifiedBy"
+          :rules="[ v => !!v || 'A person\'s legal name is required.']"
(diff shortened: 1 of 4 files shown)


src/components/common/Certify.vue
```diff
@@ -16,19 +16,21 @@
...
-        <template v-slot:label>
-          <div class="certify-stmt">
-            I, <strong>{{trimmedCertifiedBy || '[Legal Name]'}}</strong>, certify that I have relevant knowledge
-            of the {{entityDisplay || 'association'}} and that I am authorized to make this filing.
-          </div>
-        </template>
-      </v-checkbox>
-      <p class="certify-clause">Date: {{getCurrentDate}}</p>
-      <p class="certify-clause">{{message}}</p>
+      <div class="certify-statement-section">
```
JazzarKarim, line 19: I added a div here since I needed to push this entire section and align it with the text area component.

src/views/ConsentContinuationOut.vue
```diff
@@ -945,5 +945,19 @@ h2 {
   .value.certifiedby {
     padding-left: 3rem;
   }
+
+  .certify-statement-section {
+    padding-left: 15rem;
+  }
+
+  @media (max-width: 768px) {
```
JazzarKarim, line 953: For large screens:
large

For small screens:
small

src/components/common/Certify.vue
severinbeauvais: Did you test the other local filings that use this component?

src/views/ConsentContinuationOut.vue
```diff
@@ -945,5 +945,19 @@ h2 {
   .value.certifiedby {
     padding-left: 3rem;
   }
+
+  .certify-statement-section {
+    padding-left: 15rem;
```
severinbeauvais, line 950: No, don't do this. Use the grid system like the other sections (and/or v-rows/v-cols).

src/components/common/Certify.vue
JazzarKarim: Yes! That's also why I tried not to touch this component's style in this file. I didn't want to ruin the style in other local filings.

src/views/ConsentContinuationOut.vue
severinbeauvais: While this looks OK, it's reinventing the grid system and its built-in breakpoints/responsiveness.

Ref: https://v2.vuetifyjs.com/en/components/grids/

src/components/common/Certify.vue
severinbeauvais: (I don't expect any issues since you're only adding a div, which does nothing unless there is CSS or a unit test that looks for it.)

src/views/ConsentContinuationOut.vue
```diff
@@ -74,8 +74,9 @@
                       <strong>Detail</strong>
                     </v-col>
                     <v-col cols="12" sm="9">
-                      <p class="detail-text">{{defaultComment}}</p>
+                      <strong class="grey-text">{{defaultComment}}</strong>
```
severinbeauvais, line 77: I'm not sure that the strong element can have classes... More typical would be a paragraph element with a `font-weight-bold` class.

Ref: https://vuetifyjs.com/en/styles/text-and-typography/#font-emphasis

src/views/ConsentContinuationOut.vue
```diff
@@ -65,7 +65,7 @@
             <section>
               <header>
                 <h2>Ledger Detail</h2>
-                <p>Enter a detail that will appear on the ledger for this entity.</p>
+                <p class="grey-text">Enter a detail that will appear on the ledger for this entity.</p>
```
severinbeauvais, line 68: `<jedi>`I don't think this is the colour you're looking for.`</jedi>`

The UX team specified a colour of `#495057`, which we declare in theme.scss as "$gray7". You can use this variable in CSS, but in the template (as here), you can use a named colour (eg, magenta) or a Vuetify theme colour -- declared [here](https://github.com/bcgov/business-filings-ui/blob/b7c7f60caec253bb45a78ae0d6fd8765ad5eba35/src/main.ts#L100).

Your choices are:
- add gray7 to the theme (and ideally update all the code that does it the other way), or
- use a class that specifies the needed colour

severinbeauvais: Search the project for "$gray7".

If you're lucky, there will already be a base class (in base.scss) that specifies what you need (assuming you haven't overridden it somehow).

JazzarKarim: Wait Sev, I'm confused. In the CSS, I have grey-text set the color as $gray7. Line: (https://github.com/JazzarKarim/business-filings-ui/blob/eb63b323e2786770409f6501de3b12d008da02f2/src/views/ConsentContinuationOut.vue#L959)

Isn't this what you mean?

severinbeauvais: Oh, my fault, I didn't see this.

I skipped to a conclusion -- it's possible to specify a `color` in some Vuetify components, and those follow the theme as above. But this isn't what you're doing 😛 

Yes, you're doing the right thing.

src/views/ConsentContinuationOut.vue
JazzarKarim: Sev, I had a discussion with Ethan regarding the Certify component. Apparently, this design of the certify statement being on the same line as the text-field is needed everywhere in other local filings.

So I went ahead and refactored the Certify component's code and cleaned it up. It now uses the grid system.

severinbeauvais: Looks awesome!

severinbeauvais (approved):

ketaki-deodhar (approved):
LGTM!

PR #493, opened 2023-04-27
Added a missing getter

severinbeauvais:
*Issue #:* none

*Description of changes:*
- app version = 6.3.5
- added a missing getter


src/components/Dashboard/TodoList.vue
```diff
@@ -515,6 +515,7 @@ export default class TodoList extends Vue {
   @Getter getTasks!: Array<ApiTaskIF>
   @Getter getTodoListResource!: TodoListResourceIF
   @Getter isBenBcCccUlc!: boolean
+  @Getter isGoodStanding!: boolean
```
severinbeauvais, line 518: This was being inherited from a mixin, but I think it's better to directly import it in each file it's used (otherwise you lose visibility of which component uses which getter).

JazzarKarim (approved):

PR #495, opened 2023-05-02
16182 Fixed null filing fetch error + fixed Consent orderDetails and effectOfOrder

severinbeauvais:
*Issue #:* bcgov/entity#16182

*Description of changes:*

Commit 1:
- app version = 6.3.8
- get order details from correct filing object (in consent)
- display newlines in order details (consents and staff filings)
- fixed store action fetching null filing
- fixed some lint errors (maxlen)
- fixed broken consent tests
- fixed broken actions tests

Commit 2:
- now restore hasPlanOfArrangement flag from effectOfOrder property
- now save effectOfOrder property depending on hasPlanOfArrangement flag

### DIFF
--- a/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
+++ b/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
@@ -4,9 +4,9 @@
       <p class="mt-4">
         This consent is valid <strong>until {{ expiry }} at 12:01 am Pacific time</strong>.
       </p>
-      <p v-if="orderDetails" class="mt-4">{{ orderDetails }}</p>
+      <p v-if="orderDetails" class="mt-4" v-html="orderDetails" />
       <p v-if="fileNumber" class="mt-4 mb-0">Court Order Number: {{ fileNumber }}</p>
-      <p v-if="planOfArrangement" class="mt-0">{{ planOfArrangement }}</p>
+      <p v-if="hasEffectOfOrder" class="mt-0">Pursuant to a Plan of Arrangement</p>
     </template>
   </FilingTemplate>
 </template>
@@ -26,7 +26,6 @@ export default class ConsentContinuationOut extends Vue {
   @Prop({ required: true }) readonly index!: number
 
   get expiry (): string {
-    // FUTURE: expiry date may come from business object
     const expiry = this.filing.data?.consentContinuationOut?.expiry
     if (expiry) {
       return DateUtilities.apiToPacificDate(expiry, true)
@@ -35,7 +34,7 @@ export default class ConsentContinuationOut extends Vue {
   }
 
   get orderDetails (): string {
-    return this.filing.data?.consentContinuationOut?.orderDetails
+    return this.filing.data?.order?.orderDetails?.replace('\n', '<br/>')
   }
 
   /** The court order file number. */
@@ -43,9 +42,9 @@ export default class ConsentContinuationOut extends Vue {
     return this.filing.data?.order?.fileNumber
   }
 
-  /** The court order plan of arrangement. */
-  get planOfArrangement (): string {
-    return this.filing.data?.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : null
+  /** Whether the court order has an effect of order. */
+  get hasEffectOfOrder (): boolean {
+    return Boolean(this.filing.data?.order?.effectOfOrder)
   }
 }
 </script>
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -83,7 +83,10 @@ export default {
       const stateFilingUrl = rootGetters.getStateFilingUrl
 
       // if there is no state filing url, return null
-      if (!stateFilingUrl) resolve(null)
+      if (!stateFilingUrl) {
+        resolve(null)
+        return
+      }
 
       LegalServices.fetchFiling(stateFilingUrl)
         .then(filing => {
--- a/src/views/ConsentContinuationOut.vue
+++ b/src/views/ConsentContinuationOut.vue
@@ -92,7 +92,9 @@
             <section>
               <header>
                 <h2>Documents Delivery</h2>
-                <p class="grey-text">Copies of the consent to continue out documents will be sent to the email addresses listed below.</p>
+                <p class="grey-text">
+                  Copies of the consent to continue out documents will be sent to the email addresses listed below.
+                </p>
               </header>
               <div :class="{ 'invalid-section': !documentDeliveryValid && showErrors }" id="document-delivery-section">
                 <v-card flat class="py-8 px-5">
@@ -112,7 +114,9 @@
             <section>
               <header>
                 <h2>Certify</h2>
-                <p class="grey-text">Enter the legal name of the person authorized to complete and submit this correction.</p>
+                <p class="grey-text">
+                  Enter the legal name of the person authorized to complete and submit this correction.
+                </p>
               </header>
               <div :class="{ 'invalid-section': !certifyFormValid && showErrors }" id="certify-form-section">
                 <Certify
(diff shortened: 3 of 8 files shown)


src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
```diff
@@ -4,9 +4,9 @@
       <p class="mt-4">
         This consent is valid <strong>until {{ expiry }} at 12:01 am Pacific time</strong>.
       </p>
-      <p v-if="orderDetails" class="mt-4">{{ orderDetails }}</p>
+      <p v-if="orderDetails" class="mt-4" v-html="orderDetails" />
```
severinbeauvais, line 7: Now using `v-html` to display formatted text.

src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
```diff
@@ -4,9 +4,9 @@
...
       <p v-if="fileNumber" class="mt-4 mb-0">Court Order Number: {{ fileNumber }}</p>
-      <p v-if="planOfArrangement" class="mt-0">{{ planOfArrangement }}</p>
+      <p v-if="hasEffectOfOrder" class="mt-0">Pursuant to a Plan of Arrangement</p>
```
severinbeauvais, line 9: This is a simpler implementation and is exactly how StaffFiling.vue does it.

src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
```diff
@@ -35,17 +34,17 @@ export default class ConsentContinuationOut extends Vue {
   }
 
   get orderDetails (): string {
-    return this.filing.data?.consentContinuationOut?.orderDetails
+    return this.filing.data?.order?.orderDetails?.replace('\n', '<br/>')
```
severinbeauvais, line 37: This is the standard path to Order Details (same as staff filings).

severinbeauvais: Also, this code now separates multiple comment lines.

src/store/actions.ts
```diff
@@ -83,7 +83,10 @@ export default {
       const stateFilingUrl = rootGetters.getStateFilingUrl
 
       // if there is no state filing url, return null
-      if (!stateFilingUrl) resolve(null)
+      if (!stateFilingUrl) {
+        resolve(null)
+        return
```
severinbeauvais, line 88: It appears that `resolve()` doesn't return. (I know if that's "sometimes" or "always", but the return statement is always safe.)

The bug was that line 91 ran even when no state filing was specified (empty), resulting in a console error. Also it's worth mentioned that this had no functional impact, since "no filing" was fetched when "no filing" was specified.

src/views/ConsentContinuationOut.vue
```diff
@@ -92,7 +92,9 @@
             <section>
               <header>
                 <h2>Documents Delivery</h2>
-                <p class="grey-text">Copies of the consent to continue out documents will be sent to the email addresses listed below.</p>
+                <p class="grey-text">
+                  Copies of the consent to continue out documents will be sent to the email addresses listed below.
```
severinbeauvais, line 96: Fixed lint errors in this file (lines too long).

tests/unit/ConsentContinuationOut.spec.ts
```diff
@@ -153,29 +149,29 @@ describe('Consent to Continuation Out view', () => {
   })
 
   it('saves draft consent to continuation out properly', async () => {
-    // mock "save draft" endpoint (garbage response data - we aren't testing that)
-    sinon.stub(axios, 'post')
```
severinbeauvais, line 157: I changed this over to simply mocking the service method.

tests/unit/ConsentContinuationOut.spec.ts
```diff
@@ -153,29 +149,29 @@ describe('Consent to Continuation Out view', () => {
...
-              header: { filingId: 456 },
-              consentContinuationOut: { details: 'test' },
-              annualReport: {}
-            }
-          }
-        })
-      ))
+    // mock "has pending tasks" legal service
+    jest.spyOn(LegalServices, 'hasPendingTasks').mockImplementation((): any => {
+      return Promise.resolve(false)
```
severinbeauvais, line 154: This was one bug: this call wasn't mocked and didn't return a value.

However, due to another bug, this was never called.

tests/unit/ConsentContinuationOut.spec.ts
```diff
@@ -153,29 +149,29 @@ describe('Consent to Continuation Out view', () => {
...
+        business: {},
+        header: { filingId: 456 },
+        consentContinuationOut: { details: 'test' },
+        annualReport: {}
+      })
+    })
 
     // mock $route
-    const $route = { params: { filingId: '456' } }
+    const $route = { params: { filingId: '0' } }
```
severinbeauvais, line 169: This was another bug: as we are saving a new filing, the filing ID was expected to be 0.

Because this wasn't zero, some code was bypassed.

tests/unit/ConsentContinuationOut.spec.ts
```diff
@@ -153,29 +149,29 @@ describe('Consent to Continuation Out view', () => {
...
 
     // create local Vue and mock router
     createLocalVue().use(VueRouter)
     const router = mockRouter.mock()
-    router.push({ name: 'consent-continuation-out', params: { filingId: '0' } })
```
severinbeauvais, line 178: We don't need to pass the filing ID here as it's passed on the route above.

tests/unit/ConsentContinuationOut.spec.ts
```diff
@@ -194,14 +190,8 @@ describe('Consent to Continuation Out view', () => {
...
-    const button = wrapper.find('#consent-save-btn')
-    expect(button.attributes('disabled')).toBeUndefined()
-
-    // click the Save button
-    await button.trigger('click')
-
-    // wait for save to complete and everything to update
-    await flushPromises()
+    // call the save action (since clicking button doesn't work)
+    await vm.onClickSave()
```
severinbeauvais, line 194: This was another bug: the save action was actually never called.

tests/unit/ConsentContinuationOut.spec.ts
```diff
@@ -194,14 +190,8 @@ describe('Consent to Continuation Out view', () => {
...
 
     // verify new Filing ID
     expect(vm.filingId).toBe(456)
```
severinbeauvais, line 197: This looked like it worked because it was using the filing that was (incorrectly) on the route. When I fixed this, the filing ID became 0 (not 456), which is how I discovered that the save action was never called.

tests/unit/StoreActions.spec.ts
```diff
@@ -6,7 +6,7 @@ describe('Business Actions', () => {
 
   it('loads dissolution state filing', async () => {
     // init store properties we need
-    store.commit('stateFiling', 'dummy_url')
+    store.commit('setStateFiling', 'dummy_url')
```
severinbeauvais, line 9: This and the line below were also bugs: they set the wrong store property. Because of this, the `loadStateFiling()` action resolved to null, but since there was no return statement, the call to fetch a filing was still made.

Anyway, it's all ironed out now.

seeker25 (approved):

src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
```diff
@@ -35,17 +34,17 @@ export default class ConsentContinuationOut extends Vue {
...
   get fileNumber (): string {
     return this.filing.data?.order?.fileNumber
   }
 
-  /** The court order plan of arrangement. */
-  get planOfArrangement (): string {
-    return this.filing.data?.order?.effectOfOrder ? 'Pursuant to a Plan of Arrangement' : null
+  /** Whether the court order has an effect of order. */
+  get hasEffectOfOrder (): boolean {
+    return Boolean(this.filing.data?.order?.effectOfOrder)
```
JazzarKarim, line 47: Can we do `!!` instead of using `Boolean()` here to convert to boolean?

ketaki-deodhar (approved):
LGTM!

JazzarKarim (approved):
LGTM! Learned a bunch of new stuff regarding unit tests 😃

src/store/actions.ts
severinbeauvais: Before:

image

After:

image

src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
severinbeauvais: Because of all the conditional chaining, I prefer this syntax.

I definitely like the `!!` syntax when you're looking at a single variable.

My rule: if it's simple, use `!!`, other use `Boolean(<not simple stuff here>)`.

JazzarKarim: Got it. Makes a lot of sense. Thanks Sev!

PR #496, opened 2023-05-03
16182 Fixed some incomplete string replacements

severinbeauvais:
*Issue #:* bcgov/entity#16182

*Description of changes:*
- app version = 6.3.9
- replaced some replace() with replaceAll()
- added a unit test

### DIFF
--- a/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
+++ b/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
@@ -34,7 +34,7 @@ export default class ConsentContinuationOut extends Vue {
   }
 
   get orderDetails (): string {
-    return this.filing.data?.order?.orderDetails?.replace('\n', '<br/>')
+    return this.filing.data?.order?.orderDetails?.replaceAll('\n', '<br/>')
   }
 
   /** The court order file number. */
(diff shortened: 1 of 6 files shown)


src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
```diff
@@ -34,7 +34,7 @@ export default class ConsentContinuationOut extends Vue {
   }
 
   get orderDetails (): string {
-    return this.filing.data?.order?.orderDetails?.replace('\n', '<br/>')
+    return this.filing.data?.order?.orderDetails?.replaceAll('\n', '<br/>')
```
severinbeauvais, line 37: Before (only first `\n` was replaced):

image

After:

image

ketaki-deodhar (approved):
Looks good!

tests/unit/date-mixin.spec.ts
```diff
@@ -81,6 +81,7 @@ describe('Date Mixin', () => {
   xit('returns correct values for yyyyMmDdToApi()', () => {
     expect(vm.yyyyMmDdToApi('2021-01-01')).toBe('2021-01-01T08:00:00+00:00') // PST
     expect(vm.yyyyMmDdToApi('2021-07-01')).toBe('2021-07-01T07:00:00+00:00') // PDT
+    expect(vm.yyyyMmDdToApi('2023/05/03')).toBe('2023-05-03T07:00:00+00:00')
```
ketaki-deodhar, line 84: Curious here: you added a test case for the change you did (which is great) but the test itself is excluded. Do we keep the excluded tests as is and not include in the test suite?

severinbeauvais: I ran the unit tests locally:

image

This unit test previously failed in the GH check (due to a different Node version or something). I don't know if it would still fail in GH but that's more than I want to get into right now.

However, I think this unit test is still valuable to have for the future when we can un-exclude it. (I think of this test as being "blocked", not "useless".) Also I wouldn't want to just delete the test because having it shows we thought about that the method under test does.

ketaki-deodhar: Totally makes sense to not delete. Good reference for me while adding unit tests in future :)

PR #497, opened 2023-05-03
15241 minio upgrade fix

piahmed:
*Issue #:* [/bcgov/entity15241](https://github.com/bcgov/entity/issues/15241)

*Description of changes:*
Axios defaults to sending an `Authorization` header for any request it makes - newer version of Minio does not like this header, and throws 400 errors. This change will not auto-add the header for urls with `minio` in them.


seeker25:
Update package + package.lock please

PR #499, opened 2023-05-09
16153 - Use ccoExpiryDate for Consent ledger badge

chenhongjing:
*Issue #:* /bcgov/entity#16153

*Description of changes:*

- Badge is shown when Consent to Continuation Out is valid

image

### DIFF
--- a/src/store/FilingHistoryList/getters.ts
+++ b/src/store/FilingHistoryList/getters.ts
@@ -90,10 +90,8 @@ export default {
     })
     if (ccoFiling) {
       const exp = ccoFiling.data?.consentContinuationOut?.expiry
-      const ccoExpiryDate = DateUtilities.dateToYyyyMmDd(new Date(exp))
-      const currentDate = rootGetters.getCurrentDate
-      const diff = DateUtilities.daysFromToday(new Date(currentDate), new Date(ccoExpiryDate))
-      return diff >= 0
+      const ccoExpiryDate = DateUtilities.apiToDate(exp)
+      return ccoExpiryDate >= new Date()
     }
     return false
   }
--- a/src/store/FilingHistoryList/getters.ts
+++ b/src/store/FilingHistoryList/getters.ts
@@ -81,15 +81,15 @@ export default {
 
   /** Whether the business is authorized to continue out, i.e. true if cco expiry date is present or in the future. */
   isAuthorizedToContinueOut (state: FilingHistoryListStateIF, rootGetters: any): boolean {
-    const ccoFilings = state.filings.filter(val => {
+    const ccoFiling = state.filings.find(val => {
       const exp = val.data?.consentContinuationOut?.expiry
       if (exp) {
         return true
       }
       return false
     })
-    if (ccoFilings && ccoFilings.length > 0) {
-      const exp = ccoFilings[0].data?.consentContinuationOut?.expiry
+    if (ccoFiling) {
+      const exp = ccoFiling.data?.consentContinuationOut?.expiry
       const ccoExpiryDate = DateUtilities.dateToYyyyMmDd(new Date(exp))
       const currentDate = rootGetters.getCurrentDate
       const diff = DateUtilities.daysFromToday(new Date(currentDate), new Date(ccoExpiryDate))
--- a/src/store/FilingHistoryList/getters.ts
+++ b/src/store/FilingHistoryList/getters.ts
@@ -79,7 +79,7 @@ export default {
     return state.loadingOne
   },
 
-  isCcoExpired (state: FilingHistoryListStateIF, rootGetters: any): boolean {
+  isAuthorizedToContinueOut (state: FilingHistoryListStateIF, rootGetters: any): boolean {
     const ccoFilings = state.filings.filter(val => {
       const exp = val.data?.consentContinuationOut?.expiry
       if (exp) {
--- a/src/store/FilingHistoryList/getters.ts
+++ b/src/store/FilingHistoryList/getters.ts
@@ -79,6 +79,7 @@ export default {
     return state.loadingOne
   },
 
+  /** Whether the business is authorized to continue out, i.e. true if cco expiry date is present or in the future. */
   isAuthorizedToContinueOut (state: FilingHistoryListStateIF, rootGetters: any): boolean {
     const ccoFilings = state.filings.filter(val => {
       const exp = val.data?.consentContinuationOut?.expiry
--- a/src/store/FilingHistoryList/getters.ts
+++ b/src/store/FilingHistoryList/getters.ts
@@ -1,5 +1,7 @@
 import { ApiFilingIF, FilingHistoryListStateIF } from '@/interfaces'
 import { DateUtilities, EnumUtilities } from '@/services'
+import getters from '../getters'
+import { Store } from 'vuex'
 
 export default {
   getCurrentFiling (state: FilingHistoryListStateIF): ApiFilingIF {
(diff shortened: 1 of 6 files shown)


severinbeauvais (commented):
I have comments, but letting Ketaki and Karim review first :)

severinbeauvais:
@chenhongjing Are you able to self-assign this PR? That will allow the CI checks to run.

severinbeauvais:
@chenhongjing Are you able to self-assign this PR? That will allow the CI checks to run.

Also, are you able to assign a couple of reviewers?

severinbeauvais:
Please check the code smells above. I think you can fix all 3.

chenhongjing:
> @chenhongjing Are you able to self-assign this PR? That will allow the CI checks to run.
> 
> Also, are you able to assign a couple of other reviewers?

I'm not able to do that. Seems I have no access.

src/store/FilingHistoryList/getters.ts
```diff
@@ -77,5 +77,23 @@ export default {
   /** Whether one document is downloading. */
   isLoadingOne (state: FilingHistoryListStateIF): boolean {
     return state.loadingOne
+  },
+
+  isCcoExpired (state: FilingHistoryListStateIF, rootGetters: any): boolean {
```
ketaki-deodhar, line 82: Can you rename this getter to something like "isAuthorizedToCco" or any name which suggests what the getter is actually doing? :)
Current name indicates that if CcoExpired do something which is a bit confusing name for actual logic

ketaki-deodhar (commented):
Please add unit tests for new getter

src/store/FilingHistoryList/getters.ts
chenhongjing: Is it possible to keep the getter name as "isAuthorizedToContinueOut" if we plan not to use the old getter? There's a getter named "isAuthorizedToContinueOut" in src/store/getters.ts. I replace it with a new getter in src/store/FilingHistoryList/getters.ts

severinbeauvais: If the old getter is no longer used, please delete it.

src/store/FilingHistoryList/getters.ts
```diff
@@ -77,5 +77,23 @@ export default {
   /** Whether one document is downloading. */
   isLoadingOne (state: FilingHistoryListStateIF): boolean {
     return state.loadingOne
+  },
+
```
JazzarKarim, line 81: Can you please add a comment here that explains what the getter is doing?

src/store/getters.ts
```diff
@@ -209,15 +209,5 @@ export default {
         EnumUtilities.isTypeRestorationLimitedExtension(stateFiling)
       )
     )
-  },
-
-  /**
-   * True if the business is authorized to continue out, ie, the state filing is a
-   * Consent to Continuation Out filing.
-   */
-  isAuthorizedToContinueOut (_state: StateIF, getters: any): boolean {
```
JazzarKarim, line 218: I was thinking to maybe leave this getter here and implement the logic of the getter above here. @severinbeauvais @ketaki-deodhar thoughts?

tests/unit/EntityHeader.spec.ts
```diff
@@ -185,8 +211,10 @@ describe('Entity Header - AUTHORIZED TO CONTINUE OUT badge', () => {
       }
 
       // cleanup
-      store.state.stateFiling = null
+      store.state.currentDate = null
+      store.state.filingHistoryList.filings.pop()
```
severinbeauvais, line 215: 👍 This is correct. The reason you need it is that the store is the same for all the tests in this file.

src/store/FilingHistoryList/getters.ts
```diff
@@ -77,5 +77,23 @@ export default {
...
+  isAuthorizedToContinueOut (state: FilingHistoryListStateIF, rootGetters: any): boolean {
+    const ccoFilings = state.filings.filter(val => {
```
severinbeauvais, line 83: Consider using [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead -- it will return the first match (or undefined, which is falsy).

src/store/FilingHistoryList/getters.ts
```diff
@@ -77,5 +77,23 @@ export default {
...
+      const exp = val.data?.consentContinuationOut?.expiry
+      if (exp) {
+        return true
+      }
+      return false
+    })
+    if (ccoFilings && ccoFilings.length > 0) {
+      const exp = ccoFilings[0].data?.consentContinuationOut?.expiry
+      const ccoExpiryDate = DateUtilities.dateToYyyyMmDd(new Date(exp))
```
severinbeauvais, line 92: Please use the correct date utility to convert expiry into a Date object. That will ensure proper conversion and you won't need to create an object for it on line 94.

Out of curiousity, what is the expiry date format coming from Legal API?

src/store/getters.ts
severinbeauvais: I have no strong opinion, but it does make sense to have it in the FHL store since that's where the filings array is.

src/store/FilingHistoryList/getters.ts
chenhongjing: The expiry date format is something like "2023-11-09T08:00:00+00:00".

severinbeauvais: OK, that's what we call "API format". See [apiToDate()](https://github.com/bcgov/business-filings-ui/blob/532caa1dc0ef4458abdfc723198642d8e6d3c7c8/src/services/date-utilities.ts#L197).

Also, can you confirm (with Argus or Vysakh or Kevin) that the date refers to a DAY (ie, the time is always 0700 or 0800)?

severinbeauvais: (Background: sometimes a date is a DAY, and sometimes it's a date-time. In this case it looks like a day-only because the offset is exactly the Pacific-to-Greenwich offset. I prefer that day-only dates not have the time component, eg,  simply "2023-11-09", but the UI can handle it either way if it know what to expect.)

argush3: it's always 0800

severinbeauvais: Thanks.
It had better be 0700 for half the year 😉

src/store/getters.ts
ketaki-deodhar: Leaning towards keeping in filing history store

JazzarKarim: Makes sense! I'm resolving this conversation 👍

severinbeauvais: All, please never stop asking questions. This is great! And we all learn some technical stuff and some wisdom 😃

src/store/FilingHistoryList/getters.ts
chenhongjing: `ccoExpiryDate` and `currentDate` are now day-only in this getter. For example, "2023-11-09". Do you think I had better switch to another date utility?

severinbeauvais: I think your use of `DateUtilities.daysFromToday()` is fine, but it needs 2 JS Date objects. I think you only need to convert `ccoExpiryDate` efficiently.

src/store/FilingHistoryList/getters.ts
severinbeauvais: Looks good.

src/store/FilingHistoryList/getters.ts
chenhongjing: I was thinking to use `DateUtilities.isDateFuture()`. In this case, only `ccoExpiryDate` is converted. But if the date is today, it will return false (expected true).

src/store/FilingHistoryList/getters.ts
```diff
@@ -77,5 +77,24 @@ export default {
...
+
+  /** Whether the business is authorized to continue out, i.e. true if cco expiry date is present or in the future. */
+  isAuthorizedToContinueOut (state: FilingHistoryListStateIF, rootGetters: any): boolean {
+    const ccoFiling = state.filings.find(val => {
+      const exp = val.data?.consentContinuationOut?.expiry
+      if (exp) {
+        return true
+      }
+      return false
+    })
```
severinbeauvais, line 90: No change needed, but if you wanted to simplify this a bunch then you could write:
```
const ccoFiling = state.filings.find(f => f.data?.consentContinuationOut?.expiry)
```

Actually I think that ^^ is a bit too terse. My guidance is, "is the code easy to read and understand?". Your code is; mine isn't 😁

src/store/FilingHistoryList/getters.ts
severinbeauvais: You have a few options.

Once expiry is converted to a Date, you could also just compare against "now" right here.

chenhongjing: Yeah, I think that's what `DateUtilities.isDateFuture()` does, except it's not for day-only dates. So I choose to use `DateUtilities.daysFromToday()` for day-only dates, though it seems not very efficient.

severinbeauvais: Sure.

And you can simplify the conversion of expiry (API date) to a JS Date object.

chenhongjing: I've modified the file and updated the unit test. 😃 Is it better?

PR #500, opened 2023-05-09
Pinia Upgrade

JazzarKarim:
*Issue #:* /bcgov/entity#15929

*Description of changes:*
Following in Travis' footsteps, here's a little poem:

Sleep, sleep, my beloved Pinia
without worry, without fear
although my soul does not sleep,
although I do not rest.

Sleep, Pinia, sleep
may your whispers be bleak
than a leaf of grass,
or the silken fleece of lambs.


tests/unit/TodoList1.spec.ts
```diff
@@ -1766,20 +1776,21 @@ describe('TodoList - Click Tests', () => {
     wrapper.destroy()
   })
 
-  it('redirects to Pay URL when \'Retry Payment\' is clicked', async () => {
+  xit('redirects to Pay URL when \'Retry Payment\' is clicked', async () => {
```
JazzarKarim, line 1779: I'm not sure why, but many of the tests that has line 1824 across multiple files are failing. I would love some guidance on how I might fix the tests (if fixable). 

The line that's failing is:
```javascript
expect(window.location.assign).toHaveBeenCalledWith(payURL + '?accountid=' + accountId)
```
This line is found in other files so I also skipped the tests there. Not all of them though, since it does not always fail.

src/App.vue
```diff
@@ -100,7 +100,7 @@
 </template>
 
 <script lang="ts">
-import { mapActions, mapGetters, mapMutations } from 'vuex'
+import { mapActions, mapState } from 'pinia'
```
JazzarKarim, line 103: `mapGetters` is substituted by `mapState` in Pinia

JazzarKarim:
There's a Bug and there's a code smell also left. These were pre-existent in the code though (not as a result of Pinia work).

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+        }
+      })
+      this.mutateFilings(filings)
+    },
+
+    /** Closes current panel or opens new panel and loads comments and documents. */
+    toggleFilingHistoryItem (index: number): Promise<void> {
+      const rootStore = useRootStore()
+      // need to return a promise because action is called via dispatch
+      return new Promise(async resolve => {
```
severinbeauvais, line 177: [Sonarcloud reports a bug on this line.](https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=500&id=bcgov_business-filings-ui&open=AYgC5YULUDaGZf7ln5iV)

Should this action actually return a `Promise<any>` ?

severinbeauvais: Also, what part of this promise function is async?

package.json
```diff
@@ -30,11 +30,13 @@
     "@sentry/integrations": "^7.50.0",
     "@sentry/vue": "^7.50.0",
     "axios": "^0.27.2",
-    "core-js": "^3.25.2",
+    "core-js": "^3.8.3",
```
severinbeauvais, line 33: This is a step backwards but that's OK -- the next ticket will take this to ^3.30.1.

src/App.vue
```diff
@@ -143,6 +143,10 @@ import {
   Routes
 } from '@/enums'
 import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
+import { useBusinessStore } from '@/stores/businessStore'
+import { useConfigurationStore } from '@/stores/configurationStore'
+import { useFilingHistoryListStore } from '@/stores/filingHistoryListStore'
+import { useRootStore } from '@/stores/rootStore'
```
severinbeauvais, line 149: Would it be possible to have an index.ts file in `@/stores/` and then `import * from '@/stores/` here?

src/App.vue
```diff
@@ -198,17 +202,26 @@ export default {
...
+        'getEntityName',
+        'getLegalType',
+        'getIdentifier'
+      ]),
+
+    ...mapState(useRootStore,
+      [
+        'isRoleStaff',
+        'showFetchingDataSpinner'
+      ]),
```
severinbeauvais, line 224: OK, I really like this as it makes clear where the getters are coming from.

src/App.vue
```diff
@@ -309,37 +322,43 @@ export default {
...
-      'setIdentifier'
-    ]),
+    ...mapActions(useBusinessStore,
+      [
+        'loadBusinessInfo',
+        'setLegalName',
+        'setLegalType',
+        'setGoodStanding',
+        'setIdentifier'
+      ]),
```
severinbeauvais, line 332: Could you take a pass at alphabetizing all the getters and setters, please? (Easier to find when coding or reviewing code.)

src/components/Dashboard/FilingHistoryList/HeaderActions.vue
```diff
@@ -73,15 +76,14 @@ export default class HeaderActions extends Vue {
...
 
-  @Action showCommentDialog!: ActionBindingIF
-  @Action toggleFilingHistoryItem!: ActionBindingIF
-
-  @Mutation mutateCurrentFiling!: (x: ApiFilingIF) => void
-  @Mutation mutateFileCorrectionDialog!: (x: boolean) => void
+  @Action(useFilingHistoryListStore) showCommentDialog!: ActionBindingIF
+  @Action(useFilingHistoryListStore) toggleFilingHistoryItem!: ActionBindingIF
+  @Action(useFilingHistoryListStore) setCurrentFiling!: (x: ApiFilingIF) => void
+  @Action(useFilingHistoryListStore) setFileCorrectionDialog!: (x: boolean) => void
```
severinbeauvais, line 86: What do you think about changing all actions to have this kind of "type", in a future ticket? I think it works better for typings, but of course it will be some work to change them all.

src/App.vue
severinbeauvais: The index file should work but I think the import won't.

I was thinking of this: https://github.com/bcgov/business-create-ui/blob/4fdd677f468632477e027e8da840ca539158aa39/src/App.vue#L207

Which can then be used like this: https://github.com/bcgov/business-create-ui/blob/4fdd677f468632477e027e8da840ca539158aa39/src/App.vue#L235

I think "import *" doesn't buy us anything, but this would be cleaner at least:
```
import { useBusinessStore, useConfigurationStore, useFilingHistoryListStore, useRootStore } from `@/stores/`
```

src/components/Dashboard/StaffNotation.vue
```diff
@@ -239,7 +229,17 @@ export default class StaffNotation extends Vue {
...
-  @Mutation mutateFetchingDataSpinner!: ActionBindingIF
+  @Getter(useConfigurationStore) getCreateUrl!: string
+  @Getter(useConfigurationStore) getEditUrl!: string
+  @Getter(useBusinessStore) getIdentifier!: string
+  @Getter(useBusinessStore) isAdminFrozen!: boolean
+  @Getter(useBusinessStore) isBenBcCccUlc!: boolean
+  @Getter(useBusinessStore) isCoop!: boolean
+  @Getter(useBusinessStore) isFirm!: boolean
+  @Getter(useBusinessStore) isHistorical!: boolean
+  @Action(useRootStore) isInLimitedRestoration!: boolean
```
severinbeauvais, line 240: 👍 I prefer this as we get typings.

src/components/dialogs/BusinessAuthErrorDialog.vue
```diff
@@ -37,13 +37,14 @@
...
-import { mapGetters } from 'vuex'
+import { mapState } from 'pinia'
 import { ContactInfo } from '@/components/common'
+import { useRootStore } from '@/stores/rootStore'
 
 @Component({
   computed: {
     // Property definition for runtime environment.
-    ...mapGetters(['isRoleStaff'])
+    ...mapState(useRootStore, ['isRoleStaff'])
```
severinbeauvais, line 47: Please replace this one with a `@Getter`. This also replaces line 53.

src/components/dialogs/CancelPaymentErrorDialog.vue
```diff
@@ -36,13 +36,14 @@
...
-import { mapGetters } from 'vuex'
+import { mapState } from 'pinia'
 import { ContactInfo } from '@/components/common'
+import { useRootStore } from '@/stores/rootStore'
 
 @Component({
   computed: {
     // Property definition for runtime environment.
-    ...mapGetters(['isRoleStaff'])
+    ...mapState(useRootStore, ['isRoleStaff'])
```
severinbeauvais, line 46: Oh no, there's more of these. Sigh.

severinbeauvais: History lesson: This is left over from early days where most of the components in this repo used "options" API to declare themselves. What we have here is partially options API and partially class API (as per line 50).

Part 2: New components in Vue3 seem to prefer options API (or composition API), but I believe there's lots of emerging libraries to keep using class API (which I prefer due their cleaner layout IMHO), so we don't need to change anything over at the moment.

src/App.vue
JazzarKarim: I like this a lot! I'll be doing this.

src/App.vue
ketaki-deodhar: Yes. This is nice!

src/main.ts
```diff
@@ -29,6 +30,11 @@ Vue.use(Affix)
 Vue.use(Vue2Filters) // needed by SbcFeeSummary
 
 const store = getVuexStore()
+const pinia = getPiniaStore()
+
+// Needed to fix "getActivePinia was called with no active Pinia" error
+// Type assertion to turn off type checking
+Vue.use(pinia as any)
```
severinbeauvais, line 37: Tell me more about the error.

src/components/dialogs/CancelPaymentErrorDialog.vue
JazzarKarim: That makes sense. I had a feeling this was the case since this mainly appears in Filings UI which is the oldest one between our 3 UIs. I'll be replacing them with `@Getter`.

src/stores/authenticationStore.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+
+    /** Whether the user is ServiceBC Staff (which is not the same as Staff). */
+    isSbcStaff (rootState: RootStateIF): boolean {
+      return (rootState.account.currentAccount.accountType === AccountTypes.SBC_STAFF)
+    },
+
+    /** The user's Keycloak GUID. */
+    getKeycloakGuid (rootGetters): string {
+      return rootGetters['auth/keycloakGuid']
```
severinbeauvais, line 19: Can you use dot notation here for the properties (like line 24) ?

Also on line 29.

src/main.ts
JazzarKarim: So, I spent a good amount of time on this.

For Pinia, if you try to access the store before having an active Pinia instance, you will get an error. Now what makes Filings UI different is that we use the `loadConfiguration()` method from the `configurationStore` in `main.ts`

At that instance, we don't have an active instance with Vue. So, in order to fix that in Vue 2, I had to use `Vue.use(pinia). Now, if I don't type assert as any, typescript was complaining. 

Ref: https://stackoverflow.com/questions/74865884/cant-access-pinia-store-in-beforeenter-in-vue-2-app

src/stores/authenticationStore.ts
```diff
@@ -0,0 +1,55 @@
+import { AccountTypes } from '@bcrs-shared-components/enums'
+import { RootStateIF } from '@/interfaces'
+import { defineStore } from 'pinia'
+
+export const useAuthenticationStore = defineStore('authentication', {
+  getters: {
+    /** Whether the current account is a premium account. */
+    isPremiumAccount (rootState: RootStateIF): boolean {
```
severinbeauvais, line 8: Is `rootState` actually the only formal parameter here? In the old code, it was the 3rd.

src/stores/businessStore.ts
```diff
@@ -0,0 +1,290 @@
...
+      nextAnnualReport: null,
+      state: null,
+      stateFiling: null,
+      warnings: []
+    }
+  }),
+
+  getters: {
+    /** The allowed actions object. */
+    getAllowedActions (state): AllowedActionsIF {
```
severinbeauvais, line 36: In your IDE, what is the type of `state` when you hover on it here? If it's "any", please specify the actual type ("BusinessStateIF" I think).

That way we get typing on "businessInfo" and "allowedActions" etc below.

src/stores/businessStore.ts
```diff
@@ -0,0 +1,290 @@
...
+    /** The legal type. */
+    getLegalType (state): CorpTypeCd {
+      return state.businessInfo.legalType
+    },
+
+    /** The entity name, or numbered description, or empty string. */
+    getEntityName (): string {
+      return (
+        this.getLegalName ||
+      GetCorpNumberedDescription(this.getLegalType)
```
severinbeauvais, line 94: Indent.

src/stores/configurationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+import Vue from 'vue'
+import axios from '@/axios-auth'
+
+export const useConfigurationStore = defineStore('configuration', {
+  state: (): ConfigurationStateIF => ({
+    configuration: null
+  }),
+
+  getters: {
+    getAuthWebUrl (state): string {
```
severinbeauvais, line 12: Same comment as above re: type of "state".

src/stores/configurationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+      return state.configuration.VUE_APP_AUTH_WEB_URL
+    },
+
+    getMyBusinessRegistryUrl (): string {
+      return this.getAuthWebUrl + 'business'
```
severinbeauvais, line 17: Did you test this somehow? I'm not sure if "this" works here.

src/stores/configurationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+    getCreateUrl (state): string {
+      return state.configuration.VUE_APP_BUSINESS_CREATE_URL
+    },
+
+    getEditUrl (state): string {
+      return state.configuration.VUE_APP_BUSINESS_EDIT_URL
+    },
+
+    getLegalApiUrl (state): string {
+      const root = state.configuration
```
severinbeauvais, line 45: Please rename this local variable to "configuration" :)

src/stores/configurationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+
+    getSentryDsn (state): string {
+      return state.configuration.VUE_APP_SENTRY_DSN
+    }
+  },
+
+  actions: {
+    setConfiguration (data: any) {
+      this.configuration = data
```
severinbeauvais, line 91: Darn, `configuration` doesn't have a type. Oh well, leave it for now.

src/stores/configurationStore.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * Use this mutator to set a specific attribute in unit testing. Eg,
+     * store.commit('setTestConfiguration', { key: 'VUE_APP_PAY_API_URL', value: 'https://auth.web.url/' })
+     */
+    setTestConfiguration (state: ConfigurationStateIF, { key, value }: KeyValueIF) {
+      if (!state.configuration) {
+        Vue.set(state, 'configuration', {})
+      }
+      Vue.set(state.configuration, key, value)
```
severinbeauvais, line 102: Looks like this line should be in an "else" block, or we might be setting a falsy (null/undefined) value in Vue.set().

Please fix and test (per existing unit tests).

severinbeauvais:
@chenhongjing , I added you as a reviewer FYI. This PR is a pretty good crash course into most of the Filings UI architecture, etc. Ask if you have questions.

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+          console.log('Invalid filing =', filing) // eslint-disable-line no-console
+          return false
+        }
+        return true
+      })
+    },
+
+    /** The count of filings in the Filing History List. */
+    getHistoryCount (): number {
+      const filings = this.getFilings as ApiFilingIF[]
```
severinbeauvais, line 42: Is the "as" cast needed, or is the type of filings correct without it?

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+    getPendingCoa (): ApiFilingIF {
+      const businessStore = useBusinessStore()
+      const filings = this.getFilings as ApiFilingIF[]
+      return filings.find(filing => {
+        return (
+          businessStore.isBenBcCccUlc &&
+            EnumUtilities.isTypeChangeOfAddress(filing) &&
+            filing.isFutureEffective &&
+            EnumUtilities.isStatusPaid(filing) &&
+            DateUtilities.isDateFuture(filing.effectiveDate)
```
severinbeauvais, line 66: Indent (line 63-66).

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+    },
+
+    /**
+     * Fetches the filings list for the specified id from the Legal API and,
+     * if successful, saves it in the store.
+     * @param context the Vuex context (passed in automatically)
+     * @param id a business id or temporary registration number
+     */
+    loadFilings (id: string): Promise<any> {
+    // need to return a promise
```
severinbeauvais, line 146: Indent.

src/stores/filingHistoryListStore.ts
severinbeauvais: PS This is OK:
`new Promise(resolve => ...`

And this is OK:
`new Promise((resolve) => ...`

The reason for the second form is in case there is zero of more than 1 function parameter, eg:
`new Promise(() => ...`
or
`new Promise((resolve, reject) => ...`

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+            // iterate over legalFiling properties
+              for (const prop in legalFiling) {
+              // this is a legal filing output
+                let title: string
+                // use display name for primary document's title
+                if (prop === filing.name) title = filing.displayName
+                else title = EnumUtilities.filingTypeToName(prop as FilingTypes, null)
+
+                const date = DateUtilities.dateToYyyyMmDd(new Date(filing.submittedDate))
+                const filename = `${businessStore.getIdentifier} ${title} - ${date}.pdf`
```
severinbeauvais, line 261: Instead of getting the identifier here each time through this for loop, I suggest using a local variable outside the loop.

Ditto for staff role on line 268.

severinbeauvais: Eg,
`const identifier = useBusinessStore().getIdentifier`

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+          // this is a submission level output
+            const title = EnumUtilities.camelCaseToWords(prop)
+            const date = DateUtilities.dateToYyyyMmDd(new Date(filing.submittedDate))
+            const filename = `${businessStore.getIdentifier} ${title} - ${date}.pdf`
+            const link = documents[prop] as string
+            pushDocument(title, filename, link)
+          }
+        }
+      } catch (error) {
+      // set property to null to retry next time
```
severinbeauvais, line 282: Indent. Also line 286.

src/App.vue
seeker25: You already have split up stores, why not just mapStore for some of these? the strings are fairly sensitive I believe - what happens if you mapState (useRootStore, ['badfunction']), I don't think it's detected by typescript? There's probably a better way of doing it - but don't know how yet

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,322 @@
...
+      const rootStore = useRootStore()
+      // need to return a promise
+      return new Promise(resolve => {
+        this.addCommentDialog = false
+
+        // if needed, reload comments for current filing
+        if (needReload) {
+          if (this.getCurrentFiling?.commentsLink) { // safety check
+            rootStore.setFetchingDataSpinner(true)
+            this.loadComments(this.getCurrentFiling)
```
severinbeauvais, line 314: I think you need to await this, as it's async.

src/stores/authenticationStore.ts
seeker25: Better TS support using dot

src/App.vue
JazzarKarim: If I use mapStore, I'll have to take a pass at each getter/action and call the respective store that has it.

For example, for `isRoleStaff`, I'll have to call it in the code via `rootSore.isRoleStaff`. What do you think?

src/stores/rootStore.ts
```diff
@@ -0,0 +1,389 @@
...
+      return state.stateFiling
+    },
+
+    /**
+     * A formatted concatenation of the name and the effective date of the filing.
+     * Only used when entity is historical.
+     */
+    getReasonText (): string {
+      const businessStore = useBusinessStore()
+      const stateFiling = this.getStateFiling as StateFilingIF // may be null
```
severinbeauvais, line 197: Is the "as" needed or does getStateFiling already have a type?

seeker25 (approved):
High level looks good to me. Nice work.

src/views/auth/Signin.vue
```diff
@@ -6,11 +6,12 @@
...
 import { Component } from 'vue-property-decorator'
 import { navigate } from '@/utils'
-import { mapGetters } from 'vuex'
+import { mapState } from 'pinia'
+import { useConfigurationStore } from '@/stores/configurationStore'
 
 @Component({
   computed: {
-    ...mapGetters(['getLoginUrl'])
+    ...mapState(useConfigurationStore, ['getLoginUrl'])
```
severinbeauvais, line 14: Consider changing to a getter.

tests/unit/AGMDate.spec.ts
```diff
@@ -13,21 +16,23 @@ Vue.use(Vuetify)
...
   beforeEach(() => {
     // init store
-    store.state.ARFilingYear = 2019
-    store.state.arMinDate = '2019-01-01'
-    store.state.arMaxDate = '2019-12-31'
-    store.commit('setLegalType', 'CP')
-    store.commit('setLastAnnualReportDate', '2018-07-15')
+    rootStore.ARFilingYear = 2019
+    rootStore.arMinDate = '2019-01-01'
+    rootStore.arMaxDate = '2019-12-31'
```
severinbeauvais, line 31: FUTURE: if there are setters (actions) for these, we should use them instead.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -114,13 +117,13 @@ import { navigate } from '@/utils'
...
 
-  @Getter getEditUrl!: string
-  @Getter getIdentifier!: string
-  @Getter getReasonText!: string
-  @Getter isFirm!: boolean
+  @Getter(useConfigurationStore) getEditUrl!: string
+  @Getter(useBusinessStore) getIdentifier!: string
+  @Getter(useRootStore) getReasonText!: string
+  @Getter(useBusinessStore) isFirm!: boolean
   @Getter isGoodStanding!: boolean
```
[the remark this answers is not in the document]
severinbeauvais: OK. I'll resolve this conversation when you commit the change.

src/App.vue
severinbeauvais: I'd like to encourage us to use the `@Getter` decorator in class API components so that get full typings.

Some components are still using options API so this is the only syntax available to them. Oh well.

src/main.ts
severinbeauvais: Oh, OK. Please just add that reference to the code comment, and I'm OK with this.

src/stores/filingHistoryListStore.ts
severinbeauvais: Or maybe Sonarcloud is complaining about "async resolve" and it would prefer "async (resolve)" ?

severinbeauvais:
> There's a Bug and there are 2 code smells. These were pre-existent in the code though (not as a result of Pinia work).
> 
> Shall I fix them here in this PR @severinbeauvais ?

You must have already fixed the code smells as I don't see them in the Sonarcloud check.

I left you a comment about the "bug". Would be nice to fix it but don't get stuck on it.

src/App.vue
JazzarKarim: A question regarding this, I already have an index.ts file which exports both the Vuex and Pinia stores. Is it OK to add exports for all the individual Pinia stores in that file as well?

tests/unit/TodoList1.spec.ts
JazzarKarim: error
error 2

I believe the line that's causing this error is this:
```javascript
    configurationStore.setTestConfiguration({ configuration: null },
      { key: 'VUE_APP_AUTH_WEB_URL', value: 'https://auth.web.url/' })
```
I can't figure out what `setTestConfiguration` is actually doing and how to fix it. I'll look into more though and try again.

src/App.vue
severinbeauvais: I don't see why not. The "importer" can specific which exported things it imports, so it won't import the things it doesn't need.

tests/unit/TodoList1.spec.ts
severinbeauvais: The error says that `window.location.assign` wasn't called. Check the code to see what leads to this. Maybe there's some test initialization missing.

severinbeauvais:
I'm still chugging through this review but it looks great so far!

Karim, do you want to merge first and then help Hongjing rebase, or vice-versa?

JazzarKarim:
> I'm still chugging through this review but it looks great so far!
> 
> Karim, do you want to merge first and then help Hongjing rebase, or vice-versa?

Thank you so much Sev for bearing with me with this PR 😟. I'm working on fixing all that you mentioned.

I think I'll let Hongjing merge then I'll rebase. I don't want to overwhelm her with rebasing this PR since she's going to have to change her getters for Pinia.

tests/unit/AnnualReport.spec.ts
```diff
@@ -28,23 +32,27 @@ Vue.use(Vuetify)
...
-    store.state.currentFilingStatus = 'NEW'
-    store.commit('setLegalType', 'CP')
-    store.commit('setTestConfiguration', { key: 'VUE_APP_PAY_API_URL', value: 'https://auth.web.url/' })
+    businessStore.setIdentifier('CP0001191')
+    rootStore.ARFilingYear = 2017
+    rootStore.arMinDate = '2017-01-01'
+    rootStore.arMaxDate = '2018-04-30'
+    rootStore.currentFilingStatus = FilingStatus.NEW
+    businessStore.setLegalType(CorpTypeCd.COOP)
+    configurationStore.setTestConfiguration({ configuration: null },
```
severinbeauvais, line 49: Yes, this does look strange to me. Does it work in localhost, just not in the tests?

tests/unit/AnnualReport.spec.ts
```diff
@@ -871,11 +871,12 @@ describe('Annual Report - Part 3 - Submitting', () => {
     sinon.restore()
   })
 
-  it('saves a new filing and redirects to Pay URL when this is a new AR and the File & Pay button ' +
+  xit('saves a new filing and redirects to Pay URL when this is a new AR and the File & Pay button ' +
```
severinbeauvais, line 874: This needs to be fixed.

JazzarKarim: Yes. You're going to see this in a couple of different places also. They're all the same error that I mentioned above. Once I fix one, it'll fix all of them.

tests/unit/App.spec.ts
```diff
@@ -882,8 +886,8 @@ describe('App as a Draft IA with approved NR', () => {
 
   beforeAll(() => {
     // clear store
-    store.commit('tasks', [])
-    store.commit('mutateFilings', [])
+    rootStore.setTasks([])
+    filingHistoryListStore.mutateFilings([])
```
severinbeauvais, line 890: Oh, you didn't rename mutateFilings -> setFilings ?

severinbeauvais: I see setFilings() calls mutateFilings(). Otherwise it's just the unit tests using mutateFilings(). Could the unit tests call setFilings() instead (and delete mutateFilings())?

JazzarKarim: So I tested this. It works fine with setFilings!

Just for a bit of backstory: the reason why mutateFilings exists is because setFilings doesn't simply call it. Moreover, mutateFilings is called multiple times in the code. This is why both exist and are separate. 

At one point I thought of changing the name of mutateFilings to something like setCurrentFilings. However, setCurrentFiling without the s also already exists. I didn't want to cause any confusion with the code, so I just left it as is.

JazzarKarim: Nevermind. I'm looking at it again and I see that mutateFilings hasn't been called anywhere. My memory is betraying me. I'm just going to remove it 😁

src/stores/configurationStore.ts
JazzarKarim: Taken from Pinia Documentation (https://pinia.vuejs.org/core-concepts/getters.html):

pinia ref

Getters can call each other with `this`.

severinbeauvais: I never doubted you 😁 

Nice reference and good job 👍

tests/unit/DetailComment.spec.ts
```diff
@@ -1,20 +1,19 @@
...
-import { getVuexStore } from '@/store'
+import { createPinia, setActivePinia } from 'pinia'
 import { DetailComment } from '@/components/common'
 import { sleep } from '@/utils/sleep'
 
 Vue.use(Vuetify)
 
 const vuetify = new Vuetify({})
-const store = getVuexStore() as any // remove typings for unit tests
+setActivePinia(createPinia())
```
severinbeauvais, line 11: Does this test/component even need the store?

tests/unit/DocumentsList.spec.ts
```diff
@@ -116,8 +116,8 @@ describe('Documents List', () => {
     expect(wrapper.find('.download-all-btn').classes('v-btn--loading')).toBe(false)
 
     // set the flags
-    vm.mutateLoadingOne(true)
-    vm.mutateLoadingOneIndex(1)
+    vm.setLoadingOne(true)
+    vm.setLoadingOneIndex(1)
```
severinbeauvais, line 120: Probably should call the store actions instead.

Also a few places below.

tests/unit/EntityMenu.spec.ts
```diff
@@ -201,16 +203,16 @@ describe('Entity Menu - HISTORICAL badge', () => {
...
 
-      const wrapper = mount(EntityMenu, { vuetify, store, router, propsData: { businessId: null } })
+      const wrapper = mount(EntityMenu, { vuetify, router, propsData: { businessId: null } })
       await Vue.nextTick()
 
       expect(wrapper.find('#historical-chip').exists()).toBe(_.exists)
       if (_.exists) {
         expect(wrapper.find('#historical-chip').text()).toBe('HISTORICAL')
-        expect(wrapper.find('#historical-chip + span').text()).toBe(_.text)
+        // expect(wrapper.find('#historical-chip + span').text()).toBe(_.text)
```
severinbeauvais, line 215: Needs a fix?

tests/unit/TodoList1.spec.ts
```diff
@@ -1091,27 +1100,27 @@ describe('TodoList - UI - BCOMPs', () => {
...
             }
-          }
+          } as any
         },
         'enabled': true,
         'order': 1
       }
     ]
-    store.state.isCoaPending = true // normally set by FilingHistoryList
+    // store.state.isCoaPending = true // normally set by FilingHistoryList
```
severinbeauvais, line 1121: What's this?

tests/unit/TodoList1.spec.ts
severinbeauvais: If this is bogging you down, and you have tested locally and the redirects work correctly, then I propose leaving these "redirect" tests disabled, merging this, and creating a new ticket to fix the tests. Raise this at Standup tomorrow.

tests/unit/StoreCurrentAccount.spec.ts
```diff
@@ -1,47 +1,32 @@
...
       const SbcHeader = require('sbc-common-components/src/components/SbcHeader.vue').default
-      const store = await getVuexStore() as any // remove typings for unit tests
-      shallowMount(SbcHeader, { store, localVue })
+      shallowMount(SbcHeader, { localVue })
 
-      store.commit('account/setCurrentAccount', dummyAccount('PREMIUM'), { root: true })
-      expect(store.getters.isPremiumAccount).toBe(true)
-      expect(store.getters.isSbcStaff).toBe(false)
+      expect(authenticationStore.isPremiumAccount).toBe(true)
+      expect(authenticationStore.isSbcStaff).toBe(false)
```
severinbeauvais, line 19: I don't understand how this is working, since you don't set the account type ...?

tests/unit/StandaloneOfficeAddressFiling.spec.ts
```diff
@@ -204,7 +210,7 @@ describe('Standalone Office Address Filing - Part 1 - UI', () => {
       certifyFormValid: false,
       addressesFormValid: false
     })
-    await vm.$store.commit('filingData', []) // no data
+    await rootStore.setFilingData([]) // no data
```
severinbeauvais, line 213: This doesn't need to be awaited. Ditto all the other calls to this in the unit tests.

tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -745,15 +757,17 @@ describe('Standalone Directors Filing - Part 3A - Submitting filing that needs t
     // verify redirection
     const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id
     const payURL = 'https://auth.web.url/makepayment/321/' + encodeURIComponent('https://base.url/?filing_id=123')
+
+    // Need to fix this for Pinia upgrade
```
severinbeauvais, line 761: Remember to remove this comment.

tests/unit/IncorporationApplication.ts
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-import { getVuexStore } from '@/store'
+import { getVuexStore } from '@/stores'
```
severinbeauvais, line 2: This doesn't look right on a couple of levels.

Please change this import, and rename this file to `IncorporationApplication.spec.ts`, and that's it. It can be fixed later when it's actually implemented.

tests/unit/LedgerConsentContinuationOut.spec.ts
```diff
@@ -2,16 +2,15 @@ import Vue from 'vue'
 import Vuetify from 'vuetify'
 import { mount } from '@vue/test-utils'
 import ConsentContinuationOut from '@/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue'
-import { getVuexStore } from '@/store'
+import { createPinia, setActivePinia } from 'pinia'
 
 Vue.use(Vuetify)
 const vuetify = new Vuetify({})
-const store = getVuexStore() as any // remove typings for unit tests
+setActivePinia(createPinia())
```
severinbeauvais, line 9: I think this component doesn't need the store -- this can be removed.

tests/unit/MissingInformation.spec.ts
```diff
@@ -1,19 +1,19 @@
...
+import { createPinia, setActivePinia } from 'pinia'
 import MissingInformation from '@/components/Dashboard/Alerts/MissingInformation.vue'
 import { ContactInfo } from '@/components/common'
 import flushPromises from 'flush-promises'
 
 Vue.use(Vuetify)
 
 const vuetify = new Vuetify({})
-const store = getVuexStore() as any // remove typings for unit tests
+setActivePinia(createPinia())
```
severinbeauvais, line 12: Please add a comment to this test suite (say, after line 42):
```
// FUTURE: add a text to verify hidePhoneNumbers()
```

severinbeauvais (commented):
Wow, wow, wow! So much work and so many great changes. Amazing work. What a deep dive for you -- you must know everything now 😁

src/components/AnnualReport/ARDate.vue
```diff
@@ -12,18 +12,19 @@
...
-import { State, Getter } from 'vuex-class'
+import { Getter } from 'pinia-class'
 import { DateMixin } from '@/mixins'
+import { useBusinessStore } from '@/stores/businessStore'
+import { useRootStore } from '@/stores/rootStore'
 
 @Component({
   mixins: [DateMixin]
 })
 export default class ArDate extends Vue {
```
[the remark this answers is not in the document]
JazzarKarim: Sev this works! I was thinking to make this change in the other ticket since there is a checkbox that mention it? https://github.com/bcgov/entity/issues/16207

src/components/Dashboard/FilingHistoryList/HeaderActions.vue
JazzarKarim: I do agree with you. I really like this instead of having `ActionBindingIF` for everything. Shall I create a ticket for that?

src/stores/authenticationStore.ts
JazzarKarim: Yes, I believe so. In the old code, the other parameters like `state` and `getters` weren't being used. So, I took them out.

src/stores/authenticationStore.ts
JazzarKarim: I'm having trouble changing this. So, `auth` doesn't have a `keycloakGuid` field. Hence, I can't access it via the dot notation. Same thing applies for line 29.
https://github.com/JazzarKarim/business-filings-ui/blob/fbb7af423f285df019befaacfd0f3fa7acf8c6e6/src/interfaces/root-state-interface.ts#L9

src/stores/businessStore.ts
JazzarKarim: Great great catch Sev. I tried this out and I can see the type of `businessInfo`and `allowedActions` now 👍

src/stores/filingHistoryListStore.ts
JazzarKarim: When we do not cast it, my IDE says that filings has a type of any:
Screenshot (65)

For line 59, typescript was complaining although `getFilings` has a return type of `ApiFilingIF[]`:
Screenshot (66)

src/components/AnnualReport/ARDate.vue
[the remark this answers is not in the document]
severinbeauvais: Sure.

src/stores/authenticationStore.ts
severinbeauvais: I think what you have is OK (but by the way, you can't just remove formal parameters that aren't used -- the 3rd still has to be the 3rd one).

src/stores/authenticationStore.ts
severinbeauvais: That must be why Jonathan did it that way.

OK, just leave it.

src/stores/businessStore.ts
severinbeauvais: Excellent. Are you fixing them all?

src/stores/rootStore.ts
JazzarKarim: It's the same scenario as the `ApiFilingIF[]` cast above. If we do not use the as, my IDE is saying that stateFiling has a type of any if I hover over it.

src/stores/filingHistoryListStore.ts
severinbeauvais: OK, I confirm that you need the `as` here. I think it's because this object is written in "options API" (methods are separated by commas) instead of "class API".

Darn, I was hoping local types would be known.

src/stores/rootStore.ts
severinbeauvais: OK, leave it.

src/stores/businessStore.ts
JazzarKarim: Yes. I've fixed all of them across all stores.

tests/unit/AGMDate.spec.ts
JazzarKarim: Oh man. I should've done it that way while I was doing the work. I think I'll do this in the other ticket since it makes sense to use the setters instead of changing the state directly.

tests/unit/AnnualReport.spec.ts
JazzarKarim: Yes. I went ahead and tried to file an Annual Report as a regular user. It was taking me to the credit card payment page. So, hopefully this should be OK although the test is still not working 😞

tests/unit/AGMDate.spec.ts
severinbeauvais: It might not be worth fixing all the existing code, but for new code, yes!

(Note that some state doesn't have setters.)

tests/unit/TodoList1.spec.ts
JazzarKarim: I searched the entire store and I couldn't find anything called `isCoaPending`. I wanted to remove it but thought let me comment it first. Any idea what this is or why was it added Sev?

tests/unit/AnnualReport.spec.ts
severinbeauvais: That's a relief.

OK, lmk the new ticket number and I'll try to grab it soon.

tests/unit/TodoList1.spec.ts
severinbeauvais: Oh I see -- it's not a store state variables; it's a local getter in the component.

image

If the test fails then you'll need to overwrite the computed value or twiddle the store so there is a pending COA filing.

tests/unit/EntityMenu.spec.ts
JazzarKarim: Yes. I'm getting "Unknown Reason" instead of "Involuntary Dissolution – January 1, 2020 at 12:01 am Pacific time". Looking into it.

tests/unit/StoreCurrentAccount.spec.ts
JazzarKarim: I'm not really sure tbh but the tests are passing 🤔

tests/unit/StandaloneOfficeAddressFiling.spec.ts
JazzarKarim: Some of the tests are failing without the await which is so weird since I thought exactly the same thing. I'm not really sure why.

For example in this file, 3 tests are failing without the await.

tests/unit/LedgerConsentContinuationOut.spec.ts
JazzarKarim: I was wondering why I had this here. I tried running the test without the component. I was getting the no active Pinia error that I mentioned in main.ts. So, I had to add it in order for the tests to run.

JazzarKarim:
@severinbeauvais As always, thank you so much for all your thorough comments. I fixed most of the stuff you mentioned but I still have a couple of unit tests that I need to fix. Almost there! 😁

tests/unit/StandaloneOfficeAddressFiling.spec.ts
severinbeauvais: If it works with an `await Vue.nextTick()` after line 213 then I prefer that instead -- it suggests we need to wait for something to happen in the component, not that setFilingData() is async.

src/stores/filingHistoryListStore.ts
JazzarKarim: I'd like your thoughts on the new version of this Sev. SonarCloud was complaining a lot about this and labelling it as a bug ("Promise returned in function argument where a void return was expected").

I made some research and ended up not returning a Promise anymore. Let me know what do you think of the change.

severinbeauvais: You know, I never really liked the returned promise -- I thought it was redundant. It looks like you figured it out! I ran your code locally and the filing history item opens, and the comments and documents load just fine, and no console messages.

LGTM!

src/App.vue
severinbeauvais: Fixed!

src/App.vue
severinbeauvais: I added this as a "todo" in bcgov/entity#16207.

src/components/Dashboard/FilingHistoryList/HeaderActions.vue
```diff
@@ -73,15 +74,14 @@ export default class HeaderActions extends Vue {
...
-
-  @Mutation mutateCurrentFiling!: (x: ApiFilingIF) => void
-  @Mutation mutateFileCorrectionDialog!: (x: boolean) => void
+  @Action(useFilingHistoryListStore) showCommentDialog!: ActionBindingIF
+  @Action(useFilingHistoryListStore) setCurrentFiling!: (x: ApiFilingIF) => void
+  @Action(useFilingHistoryListStore) setFileCorrectionDialog!: (x: boolean) => void
+  @Action(useFilingHistoryListStore) toggleFilingHistoryItem!: ActionBindingIF
 
   /** Whether this entity is a business (and not a temporary registration). */
   get isBusiness (): string {
```
severinbeauvais, line 87: I get a error that this is the wrong return type. Quick fix, please.

src/components/dialogs/BusinessAuthErrorDialog.vue
severinbeauvais: Nice, thanks!

src/components/dialogs/CancelPaymentErrorDialog.vue
severinbeauvais: I see you changed all except the 2 components that were writte in Options API -- perfect, thanks!

src/main.ts
severinbeauvais: Thanks.

src/stores/configurationStore.ts
severinbeauvais: Thanks!

src/stores/configurationStore.ts
severinbeauvais: Thanks!

src/stores/configurationStore.ts
severinbeauvais: This works, right?

src/stores/filingHistoryListStore.ts
severinbeauvais: Looks good.

src/stores/filingHistoryListStore.ts
severinbeauvais: Great, thanks.

src/stores/filingHistoryListStore.ts
severinbeauvais: Thanks.

src/stores/filingHistoryListStore.ts
severinbeauvais: Thanks.

src/stores/rootStore.ts
```diff
@@ -0,0 +1,389 @@
...
+    },
+
+    /**
+     * This is used to show Legal Obligations only for a new business
+     * that has no tasks and hasn't filed anything yet (except their IA).
+     **/
+    isBusinessWithNoMaintenanceFilings (state): boolean {
+      const filingHistoryListStore = useFilingHistoryListStore()
+      return (
+      // no todo items
```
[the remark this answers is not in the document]
severinbeauvais: Thanks.

src/views/auth/Signin.vue
severinbeauvais: Looks good.

tests/unit/AnnualReport.spec.ts
severinbeauvais: Is there a new ticket for this?

tests/unit/TodoList1.spec.ts
severinbeauvais: Is there a new ticket for this?

tests/unit/AnnualReport.spec.ts
severinbeauvais: Two other comments already mention a new ticket for this, so I'm resolving this conversation.

tests/unit/FilingHistoryList1.spec.ts
```diff
@@ -492,11 +496,10 @@ describe('Filing History List - misc functionality', () => {
 
   xit('disables corrections when "disable changes" prop is set', async () => {
     // init store
-    store.commit('setIdentifier', 'CP0001191')
-    store.commit('mutateFilings', [])
+    businessStore.setIdentifier('CP0001191')
+    filingHistoryListStore.setFilings([])
```
severinbeauvais, line 500: Please fix lines 479-480 just like this.

tests/unit/DetailComment.spec.ts
severinbeauvais: Removed, thanks!

tests/unit/DocumentsList.spec.ts
severinbeauvais: Thanks.

tests/unit/EntityMenu.spec.ts
severinbeauvais: Something in `getReasonText` is causing it to return null or throw an error...?

tests/unit/StoreCurrentAccount.spec.ts
severinbeauvais: I played a bit with this and this test isn't working at all -- the getters aren't being called.

I looked at where else the authentication store is used: only the Add Staff Notation dialog -- and it's broken (staff filings dropdown doesn't display because `getKeycloakGuid` returns undefined.

This needs to be fixed in this PR (we can't leave it broken for any time in Dev).

tests/unit/StandaloneOfficeAddressFiling.spec.ts
severinbeauvais: I tested this and it works fine without the await in all tests except 3 -- and awaiting nextTick after setFilingData works for those. That's my preferred implementation.

tests/unit/StandaloneDirectorsFiling.spec.ts
severinbeauvais: Thanks.

tests/unit/MissingInformation.spec.ts
severinbeauvais: Thanks.

tests/unit/AnnualReport.spec.ts
JazzarKarim: New ticket: https://github.com/bcgov/entity/issues/16392

tests/unit/TodoList1.spec.ts
JazzarKarim: New ticket: https://github.com/bcgov/entity/issues/16392

tests/unit/StoreCurrentAccount.spec.ts
JazzarKarim: So, I've spent my entire afternoon trying to fix this as I made this my top priority since it's super important.

The problem wasn't with the `getKeycloakGuid` as I fixed it but it was still broken. 

The way I found the problem was by cloning my repo in a different folder and checking out commit by commit. I was sure that the staff filings dropdown was working at a certain point and one of the commits actually broke it. Each time I checked out a commit, I ran the project and checked if the dropdown was broken.

The problem was in the 'Update part 2 - lots of code cleanup" commit. Specifically, in the `Signin.vue` file.
https://github.com/bcgov/business-filings-ui/commit/a217fa1240a19ab87db6a8c97bff174e5fd9f4d1#diff-e1ac6403d23a844d2b27c3d914115fcdd0a72be080e4ddf9c2d699c6e9e37793

I removed the `@Component` since I thought we don't need it since there's no computed property anymore. It turns out, we need it although there isn't anything in it. 

Sorry for the long comment and explanation. Just felt like I should share my findings :)

severinbeauvais: Nice find. Good job finding it. Sorry I didn't notice it either.

tests/unit/TodoList1.spec.ts
JazzarKarim: Sev, can I fix this test in the other ticket (https://github.com/bcgov/entity/issues/16392)? This is the only test that's left to fix. I'm trying to set the store so that this line returns `true` but I can't seem to get it to work: https://github.com/JazzarKarim/business-filings-ui/blob/96ead3bd0a620130d82795501935ab5c1d20d359/src/stores/filingHistoryListStore.ts#L61

I can take a deeper look at it while working on the other ticket.

severinbeauvais: Sure, add this one as a todo checkbox in the other ticket, and then you can resolve this conversation.

src/components/Dashboard/FilingHistoryList/filings/IncorporationApplication.vue
```diff
@@ -77,8 +78,8 @@ export default class IncorporationApplication extends Vue {
   @Prop({ required: true }) readonly filing!: ApiFilingIF
   @Prop({ required: true }) readonly index!: number
 
-  @Getter readonly getLegalName!: string
-  @Getter readonly getMyBusinessRegistryUrl!: string
+  @Getter(useBusinessStore) readonly getLegalName!: string
+  @Getter(useConfigurationStore) readonly getMyBusinessRegistryUrl!: string
```
severinbeauvais, line 82: That's weird -- we don't usually mark getters as readonly. Please add a todo checkbox to the other ticket to fix this (here and elsewhere in this project).

JazzarKarim: Added it to the cleanup ticket (https://github.com/bcgov/entity/issues/16207).

src/stores/filingHistoryListStore.ts
```diff
@@ -0,0 +1,338 @@
...
+
+    /** Loads the comments for this history item. */
+    async loadComments (filing: ApiFilingIF): Promise<void> {
+      try {
+      // fetch comments array from API
+        const comments = await LegalServices.fetchComments(filing.commentsLink)
+        // flatten and sort the comments
+        filing.comments = flattenAndSortComments(comments)
+      } catch (error) {
+      // set property to null to retry next time
```
severinbeauvais, line 228: indent

severinbeauvais: also line 223

severinbeauvais: also line 240

severinbeauvais: also line 256

tests/unit/FilingHistoryList1.spec.ts
severinbeauvais: fixed

tests/unit/StandaloneOfficeAddressFiling.spec.ts
severinbeauvais: looks fixed

tests/unit/StoreCurrentAccount.spec.ts
```diff
@@ -1,47 +1,32 @@
...
       const SbcHeader = require('sbc-common-components/src/components/SbcHeader.vue').default
-      const store = await getVuexStore() as any // remove typings for unit tests
-      shallowMount(SbcHeader, { store, localVue })
+      shallowMount(SbcHeader, { localVue })
 
-      store.commit('account/setCurrentAccount', dummyAccount('SBC_STAFF'), { root: true })
-      expect(store.getters.isPremiumAccount).toBe(false)
-      expect(store.getters.isSbcStaff).toBe(true)
+      expect(authenticationStore.isPremiumAccount).toBe(false)
+      expect(authenticationStore.isSbcStaff).toBe(true)
```
severinbeauvais, line 29: OK, wait. These tests should still not be passing since they have the same setup but test opposite things.

I think this is a blocker (it points to something wrong, somewhere).

tests/unit/StandaloneDirectorsFiling.spec.ts
```diff
@@ -104,7 +111,7 @@ describe('Standalone Directors Filing - Part 1 - UI', () => {
       directorFormValid: true,
       certifyFormValid: true
     })
-    await vm.$store.commit('filingData', [{}]) // dummy data
```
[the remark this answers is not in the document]
JazzarKarim: On it. I'll also be removing the await and instead have `Vue.nextTick()` like you mentioned in an another comment.

tests/unit/StoreCurrentAccount.spec.ts
JazzarKarim: So, regarding the Authentication Store, between all the getters, I only see one being used and that's the `getKeycloakGuid` getter. Why are the other getters there?

Also, there's an another getter that checks if user is staff that's being used and it's in the rootStore (`isRoleStaff`). Should I instead use it in this test file then?

severinbeauvais: Jonathan probably added the other getters for completeness / future use.

It would be ideal if the unit tests ran the same code as the user (runtime). I don't know what's needed to initialize the auth store correctly in the tests.

JazzarKarim: Ohh OK that makes sense. I was wondering why were these getters there. I'll look into this Sev and try to change it to test the correct stuff.

JazzarKarim: I decided to skip this test and returned Jonathan's code since I found out that the `SbcHeader` shared component still uses Vuex. Let me know what you think Sev.

severinbeauvais: I think that we can leave this until later as long as the code works correctly locally.

package.json
```diff
@@ -52,10 +54,11 @@
...
-    "@types/jest": "^29.0.3",
+    "@pinia/testing": "^0.0.15",
+    "@types/jest": "^27.0.1",
     "@types/lodash": "^4.14.185",
     "@types/sinon": "^10.0.13",
-    "@types/vuelidate": "^0.7.15",
+    "@types/vuelidate": "^0.7.16",
     "@typescript-eslint/eslint-plugin": "^5.38.0",
     "@typescript-eslint/parser": "^5.38.0",
     "@volar-plugins/vetur": "latest",
```
seeker25, line 64: I'd recommend upgrading Vue-cli to version 5 - it gets rid of a ton of dependabot vulnerabilities, updates JEST. The version of JEST you're using isn't compatible with Typescript 4.5.5 (the warnings indicate it). Maybe for another ticket?

severinbeauvais: It was decided to upgrade to Vue CLI v5 in a separate ticket. This is what I did for Ketaki's Create UI upgrades and that separate ticket took me the better part of a week. That work is now better understood but I still think it's better to try to keep pieces as small as possible.

tests/unit/StoreCurrentAccount.spec.ts
JazzarKarim: I totally agree. The code runs locally well. I was thinking of removing the authentication store and moving the only getter we use from it to the root store. Or, we can handle this later in a later ticket. Thoughts?

severinbeauvais: My thoughts:
- change no more than needed in this ticket
- ok to disable unit tests temporarily if the code runs fine (local or dev)
- keep track of things we need to go back and fix/update

So, I have a slight preference for keeping Jonathan's authentication store code.

JazzarKarim: Alright, noted! Makes total sense. I'll add a checkbox for this in the fix unit tests ticket.

package.json
JazzarKarim: Travis, just as a reference, the ticket number is: https://github.com/bcgov/entity/issues/16207.

JazzarKarim:
@severinbeauvais Please let me know if there are any outstanding issues Sev! Looking forward to closing the curtain here 😎

tests/unit/StoreCurrentAccount.spec.ts
severinbeauvais: OK, so these tests are partly updated for Pinia but disabled as they still don't pass.

Given that the code works locally and only the unit test fails (when enabled), and we've planned the follow-on ticket for the new sprint, then I'm OK with this. Resolving.

severinbeauvais (approved):
I approve to merge this.

JazzarKarim:
@ketaki-deodhar @chenhongjing Is it OK with you to merge?

ketaki-deodhar (approved):

chenhongjing (approved):

PR #501, opened 2023-05-18
16444 Pass keycloak roles to LaunchDarkly

severinbeauvais:
*Issue #:* bcgov/entity#16444

*Description of changes:*
- app version = 6.4.1
- added getKeycloakRoles (store getter)
- renamed local getKeycloakRoles -> fetchKeycloakRoles
- now pass Keycloak roles to LaunchDarkly
- misc cleanup


severinbeauvais:
Before:

image

After:

image

src/App.vue
```diff
@@ -374,7 +375,7 @@ export default {
       try {
         // get Keycloak roles
         const jwt = this.getJWT()
-        const keycloakRoles = this.getKeycloakRoles(jwt)
```
severinbeauvais, line 377: Renamed to avoid conflict with store getter.

src/App.vue
```diff
@@ -678,7 +679,7 @@ export default {
           applicationDate: this.dateToYyyyMmDd(this.apiToDate(header.date)),
           legalFilings: [header.name]
         }
-      }
+      } as ApiFilingIF
```
severinbeauvais, line 682: I did this to cast the incomplete object into the desired type.

vysakh-menon-aot (approved):

argush3 (approved):

ketaki-deodhar (approved):

JazzarKarim (approved):
LGTM!

src/App.vue
```diff
@@ -541,8 +542,8 @@ export default {
       const email: string = userInfo.contacts[0]?.email || userInfo.email
       const firstName: string = userInfo?.firstname
       const lastName: string = userInfo?.lastname
-      // remove leading { and trailing } and tokenize string
-      const custom: any = { roles: userInfo.roles?.slice(1, -1).split(',') }
```
kialj876, line 545: Hey Severin, would it be better to update the backend call to return the roles here? That way you wouldn't need to make this change across the other UIs and anything else using the same flow would automatically get this change in their ldarkly users

kialj876: Or maybe you can remove calling the api for the userInfo altogether? I think all of the information is within the JWT which you are already parsing within the UI. Maybe there's another reason I'm not aware of

severinbeauvais: The UI gets the feature flags in the `user` response from LD.

Eg, payload:
image

Eg, response:
image

I'm afraid both the UI and the API will need to "identify" the user. I don't know whether the user's attributes is retained between `user` calls. For example, if the API provides a custom object but the UI doesn't, will LD use the custom fields from the first call?

severinbeauvais: Yes, it looks like the UI can obtain most, if not all, the user info from the JWT, so fetching the user info from the Auth API (`@me`) may not be needed.

severinbeauvais: LMK if OK to merge or if you want some changes.

PR #503, opened 2023-05-19
Package updates + lots of cleanup

JazzarKarim:
*Issue #:* /bcgov/entity#16207

*Description of changes:*
- Please check the ticket for a precise description of all the changes made.
- Service worker removed (https://github.com/bcgov/entity/issues/16170)


src/components/Dashboard/TodoList.vue
```diff
@@ -1216,16 +1209,16 @@ export default class TodoList extends Vue {
         paymentToken: header.paymentToken || null,
         payErrorObj,
         // FUTURE: ideally, this would come from the filing:
-        warnings: this.getBusinessWarnings.map(warning => warning.message)
-      }
+        warnings: (this.getBusinessWarnings as any).map(warning => warning.message)
```
JazzarKarim, line 1212: This is done to fix the `Property 'map' does not exist on type 'BusinessWarningIF'` eslint error.

tests/unit/TodoList2.spec.ts
```diff
@@ -411,11 +411,11 @@ xdescribe('TodoList - tests specific to corporations (BEN, etc)', () => {
 
 // FUTURE: implement this
 xdescribe('TodoList - tests specific to Incorporation Applications', () => {
-  beforeAll(() => {
-    sessionStorage.clear()
-    sessionStorage.setItem('TEMP_REG_NUMBER', 'TaAbBcC123')
-    businessStore.setLegalType(CorpTypeCd.BENEFIT_COMPANY)
-  })
+  // beforeAll(() => {
```
JazzarKarim, line 414: This is done to fix the `Invalid: beforeAll() may not be used in a describe block containing no tests` eslint error.

src/components/Dashboard/StaffNotation.vue
```diff
@@ -335,7 +327,7 @@ export default class StaffNotation extends Vue {
 
   @Emit('close')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  private close (needReload: boolean): void {}
+  close (needReload: boolean): void { /* no empty function */ }
```
JazzarKarim, line 330: There are a lot of these in the code. This is done to fix the `Unexpected empty method` SonarCloud code smell.

For reference: https://eslint.org/docs/latest/rules/no-empty-function

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -82,16 +81,14 @@ import { useBusinessStore, useConfigurationStore, useFilingHistoryListStore, use
...
+  }
 })
-export default class FilingHistoryList extends Vue {
+export default class FilingHistoryList extends Mixins(FilingMixin) {
   @Prop({ default: null }) readonly highlightId!: number
 
   @Getter(useFilingHistoryListStore) getCurrentFiling!: ApiFilingIF
   @Getter(useConfigurationStore) getEditUrl!: string
   @Getter(useFilingHistoryListStore) getFilings!: Array<ApiFilingIF>
-  @Getter(useBusinessStore) getIdentifier!: string
```
JazzarKarim, line 94: Eslint was complaining since this getter was also in the mixin so I removed it. This applies to many places in the code.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -186,12 +183,12 @@ export default class FilingHistoryList extends Vue {
...
    */
   @Watch('getFilings', { immediate: true })
-  private onFilingsChange (): void {
+  onFilingsChange (): void {
     // if needed, highlight a specific filing
     if (this.highlightId) {
       const index = this.getFilings.findIndex(f => (f.filingId === this.highlightId))
       if (index >= 0) { // safety check
-        this.toggleFilingHistoryItem(index, this.getFilings[index])
+        this.toggleFilingHistoryItem(index)
```
JazzarKarim, line 191: Eslint complained that this function should only have one argument but it got 2. So, I removed the second argument.

src/components/Dashboard/TodoList.vue
```diff
@@ -1027,9 +1020,9 @@ export default class TodoList extends Vue {
         correctedFilingId: correction.correctedFilingId,
         // this is only used for external corrections (IA):
         // Corrections on Alterations always have a correctedFilingType of Alteration -> alterationRequired = true
-        correctedFilingType: EnumUtilities.filingTypeToName(correction.correctedFilingType as FilingTypes, null, true),
+        correctedFilingType: EnumUtilities.filingTypeToName(correction.correctedFilingType as FilingTypes, null, null),
```
JazzarKarim, line 1023: Eslint was complaining about this since last argument was supposed to be of type `FilingSubTypes`.

src/components/Dashboard/TodoList.vue
```diff
@@ -508,15 +502,14 @@ export default class TodoList extends Vue {
...
-  @Getter(useBusinessStore) getLegalType!: CorpTypeCd
   @Getter(useBusinessStore) getIdentifier!: string
   @Getter(useRootStore) getNameRequest!: any
   @Getter(useConfigurationStore) getPayApiUrl!: string
   @Getter(useFilingHistoryListStore) getPendingCoa!: ApiFilingIF
   @Getter(useRootStore) getTasks!: Array<ApiTaskIF>
   @Getter(useRootStore) getTodoListResource!: TodoListResourceIF
   @Getter(useBusinessStore) isBenBcCccUlc!: boolean
-  @Getter(useBusinessStore) isGoodStanding!: boolean
+  @Getter(useBusinessStore) isSoleProp!: boolean
```
JazzarKarim, line 512: This getter was being used in the code but it wasn't here 😅

vue.config.js
```diff
@@ -14,10 +13,6 @@ const aboutText2 = (sbcName && sbcVersion) ? `${sbcName} v${sbcVersion}` : ''
 module.exports = {
   configureWebpack: {
     plugins: [
-      // this is needed to remove existing service workers on users' systems
-      // ref: https://www.npmjs.com/package/webpack-remove-serviceworker-plugin
-      // ref: https://github.com/NekR/self-destroying-sw/tree/master/packages/webpack-remove-serviceworker-plugin
-      new RemoveServiceWorkerPlugin({ filename: 'service-worker.js' }),
```
JazzarKarim, line 20: Reason for removing this: https://github.com/bcgov/entity/issues/16170

tsconfig.json
```diff
@@ -3,6 +3,8 @@
     "target": "esnext",
     "module": "esnext",
     "strict": false,
+    "noImplicitAny": false,
+    "strictNullChecks": false,
```
JazzarKarim, line 7: Added the two rules to stay consistent between all the 3 UIs.

tests/unit/StoreActions.spec.ts
```diff
@@ -63,7 +63,6 @@ describe('Business Actions', () => {
     await rootStore.loadStateFiling()
     expect(LegalServices.fetchFiling).toHaveBeenCalled()
     expect(rootStore.stateFiling.business.identifier).toBe('BC0871273')
-    expect(rootStore.stateFiling.consentContinuationOut.expiry).toBe('2023-12-31')
```
JazzarKarim, line 66: There is no `consentContinuationOut` field in `stateFiling`. Eslint was complaining about this.

tests/unit/allowable-actions-mixin.spec.ts
```diff
@@ -105,9 +105,11 @@ describe('Allowable Actions Mixin', () => {
     expect(vm.isAllowed(AllowableActions.BUSINESS_INFORMATION)).toBe(false)
 
     // verify both feature flag and allowed filing type
-    setFeatureFlag(true)
+    setFeatureFlag('special-resolution-ui-enabled')
     setAllowedFilingType({ name: FilingTypes.SPECIAL_RESOLUTION })
-    expect(vm.isAllowed(AllowableActions.BUSINESS_INFORMATION)).toBe(true)
```
JazzarKarim, line 110: This test was returning false for some reason. I've spent some time trying to fix it. Ditto lines 155, 165, 184, and 369.

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
```diff
@@ -15,18 +15,16 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue'
-import { Component, Prop } from 'vue-property-decorator'
+import { Component, Prop, Vue } from 'vue-property-decorator'
 import { ContactInfo } from '@/components/common'
-import { HistoryItemIF } from '@/interfaces'
```
JazzarKarim, line 21: There doesn't exist an interface called `HistoryItemIF`.

src/components/Dashboard/FilingHistoryList/filings/LimitedRestoration.vue
```diff
@@ -36,7 +35,7 @@ export default class LimitedRestoration extends Vue {
 
   /** The legal name from the restoration filing. */
   get legalName (): string {
-    return this.filing.data?.restoration?.toLegalName || '[unknown]'
+    return this.filing.data?.restoration?.legalName || '[unknown]'
```
JazzarKarim, line 38: The property is `legalName` not `toLegalName`.

Line: https://github.com/JazzarKarim/business-filings-ui/blob/d0b5ec751455746f56bb5043f4794a558ede173b/src/interfaces/api-filing-interface.ts#L99

package.json
```diff
@@ -15,75 +15,85 @@
...
-    "pdfjs-dist": "2.0.943",
     "provinces": "^1.11.0",
     "qrcode.vue": "1.7.0",
-    "regenerator-runtime": "^0.13.9",
+    "regenerator-runtime": "^0.13.11",
     "register-service-worker": "^1.7.2",
-    "sbc-common-components": "3.0.5",
-    "vue": "^2.7.10",
+    "sbc-common-components": "3.0.8",
+    "vue": "^2.7.14",
```
seeker25, line 47: Use 2.7.10 for now - Severin found some issues with 2.7.14

JazzarKarim: Deal, on it 👍

src/components/common/Directors.vue
```diff
@@ -795,7 +797,7 @@ export default class Directors extends Vue {
    * See also LegalServices.fetchParties().
    */
   // FUTURE: this API call should be in the parent component or some mixin/service
-  private async fetchDirectors (): Promise<void> {
+  async fetchDirectors (): Promise<void> {
```
JazzarKarim, line 800: There are 3 code smells that are related to the Cognitive Complexity. One of them is here. SonarCloud is complaining about them now after the package updates have been done.

I was wondering, shall I refactor these 3 methods to reduce their cognitive complexity or leave them as if for now? I'm open to suggestions.

package.json
```diff
@@ -44,7 +44,7 @@
     "regenerator-runtime": "^0.13.11",
     "register-service-worker": "^1.7.2",
     "sbc-common-components": "3.0.8",
-    "vue": "^2.7.14",
+    "vue": "^2.7.10",
```
severinbeauvais, line 47: This won't work -- npm will see the `^` and automatically use the latest "2.x.x" version. See the package-lock.json file to see what version is actually used.

Also, after you change this, `rm -rf node_modules package-lock.json`, then `npm i`.

JazzarKarim: Great catch. On it.

src/components/common/Directors.vue
```diff
@@ -588,51 +581,60 @@ export default class Directors extends Vue {
...
-  private activeIndex = -1
-  private activeIndexCustomCease = -1
-  private activeDirectorToDelete: DirectorIF = null
-  private cessationDateTemp: string = null
-  private isEditingDirector = false
-  private messageIndex = -1
+  allDirectors: DirectorIF[] = []
+
+  /** Initialize dir to use in the template's for loop. */
+  dir: DirectorIF = {
```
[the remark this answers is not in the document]
severinbeauvais: This is a decoy fix, which calms the tests perhaps, but doesn't solve the problem.

I see on line 191 that `dir` is a loop variable. Maybe try renaming it in case it's a reserved word? You could also try moving the v-for to a template element around the `<li>`.

src/components/common/Directors.vue
severinbeauvais: Leave them for now. I'll tell SC that these aren't issues.

We have to be careful to limit scope in tickets.

severinbeauvais: > Leave them for now. I'll tell SC that these aren't issues.

Done (marked as "won't fix").

JazzarKarim: Awesome, thanks Sev!

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: That's fine. I saw this (and changed it) too.

severinbeauvais: (It's because the mixins are now properly "mixed into" `this`, so their method signatures are now recognized.)

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Must have been left over from my refactor. Looks good 👍 

Do a global search for `toggleFilingHistoryItem` -- there are 4 calls to this with a second parameter -- please fix those, too. Thx.

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
```diff
@@ -15,18 +15,16 @@
...
 import { DateUtilities } from '@/services'
 
 @Component({
   components: { ContactInfo }
 })
 export default class LimitedRestorationExtensionFiling extends Vue {
   /** The subject filing. */
-  @Prop({ required: true }) readonly filing!: HistoryItemIF
+  @Prop({ required: true }) readonly filing!: any
```
severinbeauvais, line 27: Please change to `ApiFilingIF` instead.

src/components/Dashboard/FilingHistoryList/filings/LimitedRestoration.vue
severinbeauvais: Please attach a screenshot of this in action.

tests/unit/StoreActions.spec.ts
severinbeauvais: Actually, it looks like this unit test caught an error -- `StateFilingIF` should have a "consentContinuationOut" object (just as it can have a dissolution, resolution, etc objects).

Line 51 above shows the object.

Also please restore this test :)

tests/unit/TodoList2.spec.ts
severinbeauvais: 🤦‍♂️

src/mixins/allowable-actions-mixin.ts
```diff
@@ -154,7 +153,7 @@ export default class AllowableActionsMixin extends Vue {
    * @param type the filing subtype to check (optional)
    * @returns True if the specified filing is allowed, else False
    */
-  private isAllowedFiling (name: FilingTypes, type = ''): boolean {
```
severinbeauvais, line 157: No, this should be a local helped function only.

vue.config.js
severinbeauvais: You may be able to delete src/regitsterServiceWorker.ts.

See https://github.com/bcgov/business-filings-ui/pull/266 for the changes that were done to implement this feature that you are now removing.

tests/unit/FilingHistoryList2.spec.ts
```diff
@@ -108,7 +108,7 @@ filings.forEach((filing: any, index: number) => {
       // expect(item.isFutureEffectiveIaPending).toBeDefined() // FUTURE: test this more specifically
     })
 
-    itIf(isBcompCoa(filing))('BCOMP change of address filing', () => {
```
severinbeauvais, line 111: Well, it _should_ taking filing as a parameter, but `isBcompCoa` was updated in a previous commit to not take a parameter because it's not implemented yet.

I'm thinking it would be better to restore `isBcompCoa` and this line and comment them out instead.

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -491,7 +482,7 @@ export default class StandaloneDirectorsFiling extends Vue {
   }
 
   /** Called just before this component is destroyed. */
-  protected beforeDestroy (): void {
```
severinbeauvais, line 494: I think you didn't need to remove these visibility modifiers (since these methods are not called from the template, which is the problem we were trying to solve).

Anyway, no big deal because the modifiers do nothing at the moment except make developer a bit harder. No change needed, but now you know.

src/components/Dashboard/TodoList.vue
severinbeauvais: As the second and third parameters of this method have defaults, you could just omit them here (and below).

src/components/Dashboard/TodoList.vue
severinbeauvais: The root problem here is that `getBusinessWarnings` is incorrectly typed in the `@Getter`... it should be an array, and then the previous code here should be fine.

Please fix and revert this line and line 1242 below.

src/components/Dashboard/TodoList.vue
```diff
@@ -1487,19 +1480,24 @@ export default class TodoList extends Vue {
...
+         * Type assertion is done to fix TypeScript error.
+         * "This condition will always return 'false' since the types
+         * 'FilingSubTypes' and 'RestorationTypes' have no overlap."
+         */
+        if ((item.filingSubType as any) === (RestorationTypes.FULL as any)) {
           restorationType = RestorationTypes.FULL
         }
 
-        if (item.filingSubType === RestorationTypes.LIMITED) {
+        if ((item.filingSubType as any) === (RestorationTypes.LIMITED as any)) {
```
severinbeauvais, line 1492: Instead of going type-less here, I'd prefer to see the enum changed to `FilingSubTypes.FULL_RESTORATION`, etc.

severinbeauvais: Also below in `buildRestorationUrl()`.

src/components/Dashboard/FilingHistoryList/filings/LimitedRestoration.vue
JazzarKarim: Screenshot: 
limited unklnown


I've reverted this change and it's working well. I did a global search and toLegalName doesn't exist. Not sure why does it work.

src/components/common/FileUploadPdf.vue
```diff
@@ -43,10 +38,20 @@ export default class FileUploadPdf extends Vue {
...
+
+  private pdfjsLib: any
+
+  created () {
+    // NB: we load the lib and worker this way to avoid a memory leak (esp in unit tests)
+    // NB: must use require instead of import or this doesn't work
+    // NB: must use legacy build for unit tests not running in Node 18+
+    this.pdfjsLib = require('pdfjs-dist/legacy/build/pdf.js')
+    this.pdfjsLib.GlobalWorkerOptions.workerSrc = require('pdfjs-dist/legacy/build/pdf.worker.entry')
+  }
```
severinbeauvais, line 54: Nice! Does it still work? Please call this out to Riyaz in a "Testing Notes" comment.

src/components/Dashboard/TodoList.vue
JazzarKarim: I'm reverting this change. It seems to me that there were a lot of stuff that Vue was complaining about but not anymore.

I'm not really sure if it's related to the Vue version or to the rules that I've added at a certain point. I'll be going through the code again and see what changes I can revert back since they might not be a problem anymore.

JazzarKarim: If I change to `FilingSubTypes.FULL_RESTORATION`, I'm getting the same complaint as before:
"This comparison appears to be unintentional because the types 'FilingSubTypes' and 'RestorationTypes' have no overlap."

src/components/Dashboard/FilingHistoryList/filings/LimitedRestoration.vue
severinbeauvais: Unfortunately the Legal API doesn't follow a schema when providing data, so what you see here is a mismatch between ApiFilingIF and what we're actually getting back from the API.

Please fix the IF. Thanks :)

src/components/common/FileUploadPdf.vue
JazzarKarim: Yup! I've tested upload a PDF when filing a court order. It looks OK. Not sure if I missed anything though. I'll add a note for Riyaz.

src/components/Dashboard/TodoList.vue
severinbeauvais: Weird. It works for me.

image

`restorationType` should also be FilingSubTypes, not a string (and also please fix buildRestorationUrl()).

severinbeauvais: Yes, it's related to the Vue version -- the later versions do more thorough type-checking.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "6.4.2",
+  "version": "6.4.3",
```
severinbeauvais, line 3: This is a significant cleanup/update - go for 6.5.0 !

src/views/ConsentContinuationOut.vue
```diff
@@ -576,11 +569,11 @@ export default class ConsentContinuationOut extends Vue {
       this.showErrors = true
       if (!this.detailCommentValid) {
         // Show error message of detail comment text area if invalid
-        this.$refs.detailCommentRef.$refs.textarea.error = true
+        (this.$refs.detailCommentRef.$refs.textarea as any).error = true
```
severinbeauvais, line 572: You could also fix this by declaring textarea in DetailComment.vue:
```
  // Refs
  $refs!: {
    textarea: any
  }
```

severinbeauvais: Similar for certifyTextfieldRef below.

severinbeauvais (commented):
I've completed my initial review. Looks really good!

src/components/Dashboard/TodoList.vue
JazzarKarim: Screenshot (71)


I could leave it as is since the project is building successfully. What do you think?

package.json
JazzarKarim: On it! 🚀

src/components/common/Stepper.vue
```diff
@@ -38,20 +38,19 @@
```
severinbeauvais, line 38: You can fix the lint warning below by moving the v-for to the div (and then you can remove the template).

tests/unit/allowable-actions-mixin.spec.ts
JazzarKarim: Yes, makes sense. I was struggling with finding a fix for this test. I'll create a ticket and mention there all the tests that I'll be disabling here.

src/views/ConsentContinuationOut.vue
JazzarKarim: Tried this and tested, it works! There are always new stuff/ways to learn 👍

tests/unit/allowable-actions-mixin.spec.ts
JazzarKarim: New ticket: https://github.com/bcgov/entity/issues/16525

src/components/Dashboard/TodoList.vue
severinbeauvais: I forgot to remove the `as` in my screenshot above. Try this:

image

tests/unit/LedgerConsentContinuationOut.spec.ts
```diff
@@ -20,13 +20,15 @@ describe('Display ConsentContinuationOut component', () => {
         }
```
severinbeauvais, line 20: There's a console warning on the mount commands in this test suite, "[Vuetify] The v-expansion-panel component must be used inside a v-expansion-panels".

I tried a few things to fix this test issue (is not a component issue), using `parentComponent`, but it didn't quite work and I've spend enough time on it for now.

So just ignore these test warnings.

package.json
severinbeauvais: This is resolved.

src/components/common/Directors.vue
severinbeauvais: This is resolved.

package.json
severinbeauvais: This is resolved.

src/views/ConsentContinuationOut.vue
severinbeauvais: (Ideally we would declare textarea with a proper type/interface, but that's getting into Vuetify code so I'm OK with this.)

src/components/Dashboard/TodoList.vue
severinbeauvais: > `restorationType` should also be FilingSubTypes, not a string (and also please fix buildRestorationUrl()).

I'd prefer to use actual types than string :)

JazzarKarim: Oh oh sorry. My mind was thinking of something else. `restorationType` being of type RestorationTypes. OK Sev, on it. Will do that!

tests/unit/LedgerConsentContinuationOut.spec.ts
JazzarKarim: Alright, noted Sev!

src/components/common/Directors.vue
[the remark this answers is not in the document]
JazzarKarim: Both didn't work Sev. I was thinking of adding optional chaining to dir in the template. Apparently, it's not supported in Vue 2. It is in Vue 3: https://github.com/vuejs/vue/issues/11088

I'm thinking of disabling the tests for the time being and adding them to the new ticket. I'll look into this more and try to fix before doing so.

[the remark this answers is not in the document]
severinbeauvais: Does the code work correctly right now?

If it does then create a follow-up ticket with the comments above and move on.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Fixed, thanks.

src/components/common/Directors.vue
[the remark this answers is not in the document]
JazzarKarim: I've been messing around with the Directors and I see everything is working correctly with no errors. So, I guess the code is OK.

Alright, got it.

src/components/Dashboard/FilingHistoryList/filings/LimitedRestoration.vue
severinbeauvais: Fixed, thanks.

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
severinbeauvais: Fixed, thanks.

src/components/common/FileUploadPdf.vue
severinbeauvais: Resolving.

package.json
severinbeauvais: Fixed, thanks.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed, thanks.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed, thanks.

src/components/Dashboard/TodoList.vue
severinbeauvais: Fixed, thanks!

src/views/ConsentContinuationOut.vue
severinbeauvais: Fixed, thanks.

tests/unit/allowable-actions-mixin.spec.ts
severinbeauvais: Reverted, thanks.

tests/unit/StoreActions.spec.ts
severinbeauvais: Fixed, thanks.

src/mixins/allowable-actions-mixin.ts
severinbeauvais: Reverted, thanks.

tests/unit/FilingHistoryList2.spec.ts
severinbeauvais: Reverted, thanks.

src/views/StandaloneDirectorsFiling.vue
severinbeauvais: Resolving.

src/components/common/Stepper.vue
severinbeauvais: Thanks for the fix.

severinbeauvais (commented):
LGTM, but please get some other review approvals.

severinbeauvais (approved):

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: I'm OK to resolve this conversation.

tsconfig.json
seeker25: Might want to add in https://www.typescriptlang.org/tsconfig/skipLibCheck.html

What do you think Severin?

vue.config.js
seeker25: If the service worker is stuck - There's a certain header you can setup on NGINX that gets it unstuck. We had to do this when we loaded on a bad version of the service worker that turfed some clients.

tsconfig.json
severinbeauvais: Do we skip type checking of declaration files anywhere else? I'd rather my build take a few extra seconds than give up type-checking :)

seeker25: Maybe not such a good idea considering `bcrs-shared-components` isn't precompiled

seeker25 (approved):

vue.config.js
severinbeauvais: That's nice to know. Instead, we used this plugin for over a year to remove the SW.

tsconfig.json
severinbeauvais: Hahaha

vue.config.js
seeker25: Yeah the plugin wouldn't work if the service worker is toast and errors in javascript - Kinda last ditch effort we had to do with Director Search

ketaki-deodhar (approved):
I did go through all the changes. Lots of learning from the conversations. Great work @JazzarKarim ! 👍

vue.config.js
JazzarKarim: Sev and Travis, I removed the `registerServiceWorker.ts` file with it's respective import in `main.ts` in my latest commit. Please let me know if you think that's OK to do.

severinbeauvais: I think so... ? Try `npm run build`. Maybe we won't know for sure until it's deployed to Dev.

JazzarKarim: I did try a build, everything looks OK. I also tested in local. Everything seems to be running normally. Okay then, hopefully all goes smoothly when I merge. Thanks!

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-503-8u5hy650.web.app

PR #511, opened 2023-05-31
Created the Foreign Jurisdiction Component

JazzarKarim:
*Issue #:* /bcgov/entity#16540

*Description of changes:*
- Created the new Foreign Jurisdiction Component
- When country is not Canada or USA:
non canada us

- When country is Canada or USA:
canada

### DIFF
--- a/src/components/common/ForeignJurisdiction.vue
+++ b/src/components/common/ForeignJurisdiction.vue
@@ -6,6 +6,7 @@
       </v-col>
       <v-col cols="12" sm="9">
         <v-select
+          id="country-selector"
           :items="countryNames"
           v-model="selectedCountry"
           filled
@@ -15,6 +16,7 @@
         />
         <v-select
           v-if="canadaUsaRegions.length > 0"
+          id="region-selector"
           :items="canadaUsaRegions"
           v-model="selectedRegion"
           filled
@@ -36,10 +38,12 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
   selectedCountry = ''
   selectedRegion = ''
 
+  /** Get country names as an array of strings. */
   get countryNames (): string[] {
     return this.getCountries().map(country => country.name)
   }
 
+  /** Get the respective regions of the country selected as an array of strings. */
   get canadaUsaRegions (): string[] {
     if (this.selectedCountry === 'Canada') {
       let regions = this.getCountryRegions('CA').map(region => region.name)
@@ -66,6 +70,7 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
     ]
   }
 
+  /** Emit event whenever a country is selected. */
   @Emit('update:country')
   emitChangedCountry (): void {
     this.selectedRegion = ''
@@ -80,6 +85,7 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
     }
   }
 
+  /** Helper function to validate if a region is selected when applicable. */
   private validateRegions (): void {
     if (this.canadaUsaRegions.length > 0) {
       if (this.selectedRegion !== '') {
@@ -90,6 +96,7 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
     }
   }
 
+  /** Emit event whenever a region is selected. */
   @Emit('update:region')
   emitChangedRegion (): void {
     if (this.canadaUsaRegions.length > 0) {
--- a/src/components/common/ForeignJurisdiction.vue
+++ b/src/components/common/ForeignJurisdiction.vue
@@ -94,6 +94,23 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
     return countryCode?.code
   }
 
+  /** Helper function to get a country's name when given the code.
+   * @example ('CA') -> 'Canada'
+   */
+  private getCountryNameFromCode (code: string): string {
+    const country = this.getCountries().find(country => country.code === code)
+    return country?.name
+  }
+
+  /** Helper function to get a region's name when given the short.
+   * @example ('AB') -> 'Alberta'
+   * @example ('NY') -> 'New York'
+   */
+  private getRegionNameFromCode (short: string): string {
+    const region = this.canadaUsaRegions?.find(region => region.short === short)
+    return region?.name
+  }
+
   @Watch('draftCountry')
   onDraftCountryChanged (val: string): void {
     this.selectedCountryName = val
@@ -115,23 +132,6 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
       this.emitValid(true)
     }
     return this.selectedCountryCode
-  }
-
-  /** Helper function to get a country's name when given the code.
-   * @example ('CA') -> 'Canada'
-   */
-  private getCountryNameFromCode (code: string): string {
-    const country = this.getCountries().find(country => country.code === code)
-    return country?.name
-  }
-
-  /** Helper function to get a region's name when given the short.
-   * @example ('AB') -> 'Alberta'
-   * @example ('NY') -> 'New York'
-   */
-  private getRegionNameFromCode (short: string): string {
-    const region = this.canadaUsaRegions?.find(region => region.short === short)
-    return region?.name
   }
 
   /** Emit the selected country's region short whenever a new region is selected. */
--- a/src/components/common/ForeignJurisdiction.vue
+++ b/src/components/common/ForeignJurisdiction.vue
@@ -15,7 +15,7 @@
           placeholder="Jurisdiction Country"
           :rules="countryRules"
           menu-props="auto"
-          @change="emitChangedCountry($event)"
+          @input="emitChangedCountry($event)"
         />
         <v-select
           v-if="canadaUsaRegions.length > 0"
@@ -27,7 +27,7 @@
           return-object
           placeholder="Jurisdiction Region"
           :rules="regionRules"
-          @change="emitChangedRegion($event)"
+          @input="emitChangedRegion($event)"
         />
       </v-col>
     </v-row>
@@ -59,14 +59,14 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
   /** Validation rules for the Jurisdiction Country dropdown. */
   get countryRules (): Array<(val) => boolean | string> {
     return [
-      val => (val.length !== 0) || 'Jurisdiction Country is required.'
+      val => !!(val.code) || 'Jurisdiction Country is required.'
     ]
   }
 
   /** Validation rules for the Jurisdiction Region dropdown. */
   get regionRules (): Array<(val) => boolean | string> {
     return [
-      val => (val.length !== 0) || 'Jurisdiction Region is required.'
+      val => !!(val.short) || 'Jurisdiction Region is required.'
     ]
   }
 
@@ -89,7 +89,7 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
   /** Helper function to validate if a region is selected when applicable. */
   private validateRegions (): void {
     if (this.canadaUsaRegions.length > 0) {
-      if (this.selectedRegion.name !== '') {
+      if (!this.selectedRegion) {
         this.emitValid(true)
       } else {
         this.emitValid(false)
--- a/src/components/common/ForeignJurisdiction.vue
+++ b/src/components/common/ForeignJurisdiction.vue
@@ -98,15 +98,7 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
 
   /** Emit event whenever a region is selected. */
   @Emit('update:region')
-  emitChangedRegion (): void {
-    if (this.canadaUsaRegions.length > 0) {
-      if (this.selectedRegion !== '') {
-        this.emitValid(true)
-      } else {
-        this.emitValid(false)
-      }
-    }
-  }
+  emitChangedRegion (): void { /* no empty function */ }
 
   /** Emits an event indicating whether or not the form is valid. */
   @Emit('valid')
--- a/src/components/common/ForeignJurisdiction.vue
+++ b/src/components/common/ForeignJurisdiction.vue
@@ -1,5 +1,5 @@
 <template>
-  <div id="foreign-jurisdiction">
+  <v-card flat id="foreign-jurisdiction">
     <v-row no-gutters>
       <v-col cols="12" sm="3">
         <label class="title-label">Jurisdiction</label>
@@ -31,7 +31,7 @@
         />
       </v-col>
     </v-row>
-  </div>
+  </v-card>
 </template>
 
 <script lang="ts">
@@ -45,12 +45,12 @@ export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin)
 
   /** Get the respective regions of the country selected as an array of objects. */
   get canadaUsaRegions (): Array<object> {
-    if (this.selectedCountry.name === 'Canada') {
+    if (this.selectedCountry.code === 'CA') {
       let regions = this.getCountryRegions('CA')
-      regions = regions.filter(province => province.name !== 'British Columbia')
-      regions.push({ name: 'Federal', code: 'FD' })
+      regions = regions.filter(province => province.short !== 'BC')
+      regions.push({ name: 'Federal', short: 'FD' })
       return regions
-    } else if (this.selectedCountry.name === 'United States of America') {
+    } else if (this.selectedCountry.code === 'US') {
       return this.getCountryRegions('US')
     }
     return []
--- a/src/components/common/ForeignJurisdiction.vue
+++ b/src/components/common/ForeignJurisdiction.vue
@@ -1,0 +1,117 @@
+<template>
+  <v-card flat class="pt-6 px-4">
+    <v-row no-gutters>
+      <v-col cols="12" sm="3">
+        <label class="title-label">Jurisdiction</label>
+      </v-col>
+      <v-col cols="12" sm="9">
+        <v-select
+          :items="countryNames"
+          v-model="selectedCountry"
+          filled
+          placeholder="Jurisdiction Country"
+          :rules="countryRules"
+          @change="emitChangedCountry($event)"
+        />
+        <v-select
+          v-if="canadaUsaRegions.length > 0"
+          :items="canadaUsaRegions"
+          v-model="selectedRegion"
+          filled
+          placeholder="Jurisdiction Province"
+          :rules="regionRules"
+          @change="emitChangedRegion($event)"
+        />
+      </v-col>
+    </v-row>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Mixins } from 'vue-property-decorator'
+import { CountriesProvincesMixin } from '@/mixins'
+
+@Component({})
+export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin) {
+  selectedCountry = ''
+  selectedRegion = ''
+
+  get countryNames (): string[] {
+    return this.getCountries().map(country => country.name)
+  }
+
+  get canadaUsaRegions (): string[] {
+    if (this.selectedCountry === 'Canada') {
+      let regions = this.getCountryRegions('CA').map(region => region.name)
+      regions = regions.filter(province => province !== 'British Columbia')
+      regions.push('Federal')
+      return regions
+    } else if (this.selectedCountry === 'United States of America') {
+      return this.getCountryRegions('US').map(region => region.name)
+    }
+    return []
+  }
+
+  /** Validation rules for the Jurisdiction Country dropdown. */
+  get countryRules (): Array<(val) => boolean | string> {
+    return [
+      val => (val && val.trim().length > 0) || 'Jurisdiction Country is required.'
+    ]
+  }
+
+  /** Validation rules for the Jurisdiction Region dropdown. */
+  get regionRules (): Array<(val) => boolean | string> {
+    return [
+      val => (val && val.trim().length > 0) || 'Jurisdiction Region is required.'
+    ]
+  }
+
+  @Emit('update:country')
+  emitChangedCountry (): void {
+    this.selectedRegion = ''
+    if (this.selectedCountry !== '') {
+      if (this.selectedCountry === 'Canada' || this.selectedCountry === 'United States of America') {
+        this.validateRegions()
+      } else {
+        this.emitValid(true)
+      }
+    } else {
+      this.emitValid(false)
+    }
+  }
+
+  private validateRegions (): void {
+    if (this.canadaUsaRegions.length > 0) {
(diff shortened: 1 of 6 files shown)


src/views/ConsentContinuationOut.vue
```diff
@@ -88,6 +88,15 @@
               </div>
             </section>
 
+            <section>
```
JazzarKarim, line 91: I'll be removing this section before I merge. I just have it here so that you can play around with it and try it out in the temporary URL.

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-511-plcasw8e.web.app

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-511-plcasw8e.web.app

src/views/ConsentContinuationOut.vue
severinbeauvais: Glad you have that in there for temporary testing (which looks good btw).

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+  <v-card flat class="pt-6 px-4">
```
severinbeauvais, line 2: You should let the parent specify the whitespace around this component.

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
...
+      <v-col cols="12" sm="9">
+        <v-select
+          id="country-selector"
+          :items="countryNames"
+          v-model="selectedCountry"
+          filled
+          placeholder="Jurisdiction Country"
+          :rules="countryRules"
+          @change="emitChangedCountry($event)"
+        />
```
severinbeauvais, line 16: You'll have to tweak a style so the "g" shows correctly in this drop-down menu. (Probably ditto for jurisdiction province.)

I did this somewhere, some time. I can't remember. Just look at the various elements that Vuetify renders for this component and you'll find the one that is too short or whatever.

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
...
+        <v-select
+          v-if="canadaUsaRegions.length > 0"
+          id="region-selector"
+          :items="canadaUsaRegions"
+          v-model="selectedRegion"
+          filled
+          placeholder="Jurisdiction Province"
```
severinbeauvais, line 23: How about making this dynamic so it says "Jurisdiction State" for USA?

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
...
+import { CountriesProvincesMixin } from '@/mixins'
+
+@Component({})
+export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin) {
+  selectedCountry = ''
+  selectedRegion = ''
+
+  /** Get country names as an array of strings. */
+  get countryNames (): string[] {
+    return this.getCountries().map(country => country.name)
```
severinbeauvais, line 43: Very nice.

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /** Get the respective regions of the country selected as an array of strings. */
+  get canadaUsaRegions (): string[] {
+    if (this.selectedCountry === 'Canada') {
+      let regions = this.getCountryRegions('CA').map(region => region.name)
+      regions = regions.filter(province => province !== 'British Columbia')
+      regions.push('Federal')
```
severinbeauvais, line 51: Beautiful.

src/components/common/ForeignJurisdiction.vue
severinbeauvais: Or call it "Jurisdiction Region" per the rules?

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
...
+    } else if (this.selectedCountry === 'United States of America') {
+      return this.getCountryRegions('US').map(region => region.name)
+    }
+    return []
+  }
+
+  /** Validation rules for the Jurisdiction Country dropdown. */
+  get countryRules (): Array<(val) => boolean | string> {
+    return [
+      val => (val && val.trim().length > 0) || 'Jurisdiction Country is required.'
```
severinbeauvais, line 62: Why trim it -- are there countries (or regions below) with only spaces? We control what displays in there, right?

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,116 @@
...
+    if (this.selectedCountry !== '') {
+      if (this.selectedCountry === 'Canada' || this.selectedCountry === 'United States of America') {
+        this.validateRegions()
+      } else {
+        this.emitValid(true)
+      }
+    } else {
+      this.emitValid(false)
+    }
+  }
```
severinbeauvais, line 86: Does this method actually emit anything? I think you have to return a value.

Ditto for emitChangedRegion() below.

src/views/ConsentContinuationOut.vue
severinbeauvais: Yes, remove when done because you don't do anything with the data and we don't want to leave in code that looks like it's doing something but it really isn't (decoy pattern).

src/components/common/ForeignJurisdiction.vue
severinbeauvais, line 2: You should assign an id or a class to name this component. (Sort of a general practise for all components, if not used for styling then maybe used for testing.)

src/components/common/ForeignJurisdiction.vue
severinbeauvais: Just one thing: I think the country code (and region code) needs to be provided to the Legal API, so you'll have to figure that out.

It may be possible to use an array of objects for a drop-down menu. Then you'd have access to all object properties when one is selected.

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  it('initializes correctly', () => {
+    const wrapper = mount(ForeignJurisdiction,
+      {
+        vuetify
+      })
+    const vm: any = wrapper.vm
+
+    expect(vm.selectedCountry).toBe('')
+    expect(vm.selectedRegion).toBe('')
+    expect(vm.countryNames.length).toBe(249)
```
severinbeauvais, line 20: This may change over time since we're importing a package and the world is changing.

However, I say leave it. Some future developer may have to update this but it serves the purpose of verifying that the country array is loaded.

Add a comment that this may change, or similar.

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const wrapper = mount(ForeignJurisdiction,
+      {
+        vuetify
+      })
+    const vm: any = wrapper.vm
+
+    const countriesSelector = wrapper.find('#country-selector')
+    await countriesSelector.setValue('Lebanon')
+
+    expect(vm.selectedCountry).toBe('Lebanon')
```
severinbeauvais, line 35: This should expect "LB" (abbreviation for Lebanon).

severinbeauvais: Also you should check that an update event was emitted. (Ditto all unit tests where you can verify this.)

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+      })
+    const vm: any = wrapper.vm
+
+    const countriesSelector = wrapper.find('#country-selector')
+    await countriesSelector.setValue('Canada')
+    const regionsSelector = wrapper.find('#region-selector')
+    await regionsSelector.setValue('Ontario')
+
+    expect(vm.selectedCountry).toBe('Canada')
+    expect(vm.selectedRegion).toBe('Ontario')
```
severinbeauvais, line 53: Should be "ON".

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const wrapper = mount(ForeignJurisdiction,
+      {
+        vuetify
+      })
+    const vm: any = wrapper.vm
+
+    const countriesSelector = wrapper.find('#country-selector')
+    await countriesSelector.setValue('Lebanon')
+
+    vm.emitChangedCountry()
```
severinbeauvais, line 67: I think you can actually emit a change event on the component, so that you don't have to call this directly.

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    const regionsSelector = wrapper.find('#region-selector')
+    await regionsSelector.setValue('Ontario')
+
+    vm.emitChangedCountry()
+    vm.emitChangedRegion()
+
+    expect(wrapper.emitted('valid')).toBeTruthy()
+    wrapper.destroy()
+  })
+})
```
severinbeauvais, line 104: Some suggestions:
- a test for region displaying when country is USA
- a test for region not displaying when country is not Canada or USA
- a test for no BC when Canada
- a test for Federal when Canada

tests/unit/ForeignJurisdiction.spec.ts
severinbeauvais: Example:
https://github.com/bcgov/business-edit-ui/blob/e37407b44e3233773cfb913151f0bc3812bde0fc/tests/unit/OrgPerson.spec.ts#L279

---

To do this on some components, you have to select them by either id or classname -- Vuetify uses one of those on the component's div and the other on the embedded HTML input element.

Example showing the id on the input element, but if there was a class then it would be on the top div:

image

severinbeauvais: Example:
https://github.com/bcgov/business-edit-ui/blob/e37407b44e3233773cfb913151f0bc3812bde0fc/tests/unit/OrgPerson.spec.ts#L279

---

To do this on some components, you have to select them by either id or classname -- Vuetify uses one of those on the component's div and the other on the embedded HTML input element.

Example showing the id on the input element, but if there was a class then it would be on the top div:

image

src/components/common/ForeignJurisdiction.vue
ketaki-deodhar: I think same is id for "y". Fix for this might fix both

ketaki-deodhar (commented):
Great work Karim!

JazzarKarim:
I just disabled my unit tests for now as it's taking me some time to fix them (figuring things out...). I'll have another commit later to fix them.

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,124 @@
...
+      val => (val.length !== 0) || 'Jurisdiction Region is required.'
+    ]
+  }
+
+  /** Emit event whenever a country is selected. */
+  @Emit('update:country')
+  emitChangedCountry (): object {
+    this.selectedRegion = {}
+    if (this.selectedCountry.name !== '') {
+      if (this.selectedCountry.name === 'Canada' || this.selectedCountry.name === 'United States of America') {
```
severinbeauvais, line 78: It would be simpler to compare country codes ;)

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <div id="foreign-jurisdiction">
```
severinbeauvais, line 2: This can still be a v-card (as you had it originally) -- any styling you put in the parent on this component will be applied to the top element in this component (inside the template block).

tests/unit/ForeignJurisdiction.spec.ts
severinbeauvais: Fix looks good:

image

image

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,124 @@
...
+  selectedRegion = {}
+
+  /** Get the respective regions of the country selected as an array of objects. */
+  get canadaUsaRegions (): Array<object> {
+    if (this.selectedCountry.name === 'Canada') {
+      let regions = this.getCountryRegions('CA')
+      regions = regions.filter(province => province.name !== 'British Columbia')
+      regions.push({ name: 'Federal', code: 'FD' })
+      return regions
+    } else if (this.selectedCountry.name === 'United States of America') {
```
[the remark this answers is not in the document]
JazzarKarim: Oh shoot, I missed them. I'll be doing that Sev.

src/components/common/ForeignJurisdiction.vue
JazzarKarim: I never knew that actually. OK Sev, good to know. Thanks!

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-511-plcasw8e.web.app

src/components/common/ForeignJurisdiction.vue
severinbeauvais: Looks good.

src/components/common/ForeignJurisdiction.vue
severinbeauvais: Fix looks good:

image

image

tests/unit/ForeignJurisdiction.spec.ts
severinbeauvais: Please check that country and region are correctly emitted.

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -29,80 +29,124 @@ describe('ForeignJurisdiction', () => {
       })
     const vm: any = wrapper.vm
 
-    wrapper.find('#country-selector').trigger('click')
-    await Vue.nextTick()
-    wrapper.find('.menuable__content__active').findAll('.v-list-item').at(123).trigger('click')
+    await wrapper.find('#country-selector').trigger('click')
+    wrapper.findAll('.v-list-item').at(123).trigger('click')
```
severinbeauvais, line 33: You should await this one, too, since it's async.

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -29,80 +29,124 @@ describe('ForeignJurisdiction', () => {
...
 
     expect(vm.selectedCountry).toEqual({code: 'LB', name: 'Lebanon'})
     expect(vm.selectedRegion).toEqual({})
```
severinbeauvais, line 36: I'm wondering if this should expect null? (I know it depends on how you initialized this variable.)

I'm thinking ahead to what this value will look like in the store and how we will save it to the filing object. Maybe this will work?
```
filing.consentContinuationOut.foreignJurisdiction.region = this.getSelectedRegion.code || undefined
```

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -29,80 +29,124 @@ describe('ForeignJurisdiction', () => {
...
-    await Vue.nextTick()
-    wrapper.findAll('.v-list-item').at(39).trigger('click')
+    await wrapper.find('#country-selector').trigger('click')
+    await wrapper.findAll('.v-list-item').at(39).trigger('click')
+
+    const region = wrapper.find('#region-selector')
+    await region.setValue(JSON.stringify({code: 'ON', name: 'Ontario'}))
+
     expect(vm.selectedCountry).toEqual({code: 'CA', name: 'Canada'})
+    expect(vm.selectedRegion).toEqual(JSON.stringify({code: 'ON', name: 'Ontario'}))
```
severinbeauvais, line 54: Why does this need to be stringified but the check on line 54 doesn't?

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -29,80 +29,124 @@ describe('ForeignJurisdiction', () => {
...
-    wrapper.find('#region-selector').trigger('click')
-    wrapper.find('.menuable__content__active').findAll('.v-list-item').at(7).trigger('click')
-    await Vue.nextTick()
-    expect(vm.selectedRegion).toEqual({name: 'Ontario', short: 'ON'})
+    await wrapper.find('#country-selector').trigger('click')
+    await wrapper.findAll('.v-list-item').at(39).trigger('click')
+
+    const provinces = wrapper.findAll('.v-select').at(1).props('items')
+
+    expect(!provinces.includes({'name': 'British Columbia', 'short': 'BC'}))
```
severinbeauvais, line 69: Can this be written as `expect(provinces).includes(...` ?

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -29,80 +29,124 @@ describe('ForeignJurisdiction', () => {
...
       })
 
-    const countriesSelector = wrapper.find('#country-selector')
-    await countriesSelector.setValue('Canada')
+    await wrapper.find('#country-selector').trigger('click')
+    wrapper.findAll('.v-list-item').at(123).trigger('click')
 
-    expect(wrapper.emitted('valid')).toBeFalsy()
+    expect(wrapper.emitted('update:country').pop()[0]).toEqual({code: 'LB', name: 'Lebanon'})
+    expect(wrapper.emitted('valid')).toBeTruthy()
```
severinbeauvais, line 116: I recommend using `toBe(true)` (and false in test below) so that we check explicity values. This prevents errors where, for instance, `{}` is truthy. And so is `123`. But neither of those are correct for "valid".

tests/unit/ForeignJurisdiction.spec.ts
JazzarKarim: This is exactly what I had in mind. It should work!

Shall I change the test? The test would fail since it's expecting `{}`. I'd have to change the initialization of the object. I think initializing to empty object should be OK though.

src/components/common/ForeignJurisdiction.vue
```diff
@@ -0,0 +1,124 @@
...
+</template>
+
+<script lang="ts">
+import { Component, Emit, Mixins } from 'vue-property-decorator'
+import { CountriesProvincesMixin } from '@/mixins'
+
+@Component({})
+export default class ForeignJurisdiction extends Mixins(CountriesProvincesMixin) {
+  selectedCountry = {}
+  selectedRegion = {}
```
severinbeauvais, line 44: This component needs the ability to restore foreign jurisdiction from a draft filing. You'll need to accept a couple of props, and you'll need to watch them and update your local variables when they change. (You'll probably also want to use the `immediate: true` option to set your local variables from the props when this component is created.)

severinbeauvais: ^^ Also add a unit test(s) for this.

tests/unit/ForeignJurisdiction.spec.ts
JazzarKarim: I was running into all sorts of errors and problems since `v-select` works in mysterious ways. Basically, for line 53, it's OK since I'm setting the object via a click. For the region, when I try do that, I'm running into a problem where it's not setting it correctly. By using the `stringify` method, it fixed the problem.

I can show you more details if you want Sev regarding this. It's just I spent way too much time here than I would like 😟

tests/unit/ForeignJurisdiction.spec.ts
JazzarKarim: Hmm, this doesn't seem to work.

tests/unit/ForeignJurisdiction.spec.ts
severinbeauvais: Not sure. Thinking ahead is risky and especially on a Friday.

It may help to try implementing parts of 16528 (ie, save/resume) and see what works best.

tests/unit/ForeignJurisdiction.spec.ts
severinbeauvais: Are you able to "click" the region same as country?

tests/unit/ForeignJurisdiction.spec.ts
JazzarKarim: Agreed. Alright, sounds good.

tests/unit/ForeignJurisdiction.spec.ts
severinbeauvais: Try `toContain()`.

tests/unit/ForeignJurisdiction.spec.ts
JazzarKarim: No! I've tried every single possible way ever. It doesn't work. Pretty sure since there are two v-selects. It's not getting the input as I had hoped for. I've searched everywhere but there isn't much out there. This is the solution I came up with.

severinbeauvais: OK, leave it then. It's fine, just different.

tests/unit/ForeignJurisdiction.spec.ts
JazzarKarim: There's a code smell regarding this. I've tried  `toContain()` but it doesn't also seem to work. I'll be looking into this to fix it 👍

severinbeauvais: There has to be some way to check if an array has a particular element...

https://jestjs.io/docs/expect

tests/unit/ForeignJurisdiction.spec.ts
```diff
@@ -66,7 +66,7 @@ describe('ForeignJurisdiction', () => {
 
     const provinces = wrapper.findAll('.v-select').at(1).props('items')
 
-    expect(!provinces.includes({'name': 'British Columbia', 'short': 'BC'}))
+    expect(provinces).toEqual(expect.not.arrayContaining([{name: 'British Columbia', short: 'BC'}]))
```
severinbeauvais, line 69: Thanks for figuring this out. Can it be simplified?
`expect(provinces).not.arrayContaining([{name: 'British Columbia', short: 'BC'}])`

JazzarKarim: This didn't work. Sev, for reference: 
not containing

severinbeauvais: I have to stop doubting that you've already tried everything, lol.

Fair enough. This is resolved.

severinbeauvais (approved):
Good from me when you sort out the final things

JazzarKarim:
@severinbeauvais I've made some changes to the component. I, of course, would've loved for you to take a look at the changes before I merge. However, since you're away, I'll be merging soon. My follow-up ticket will use this component so I'm ready to make any changes if you see the need once you come back!

Also, I tested every single possible scenario in local before committing (by printing results and unit tests). All seems to be working well. So, it should be fine hopefully!

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-511-plcasw8e.web.app

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-511-plcasw8e.web.app

ketaki-deodhar (approved):
Looks good!

chenhongjing (approved):
LGTM 🐱

PR #513, opened 2023-06-02
16344 Fix pending consent ledger item blurb

leodube-aot:
*Issue #:* /bcgov/entity#16344

*Description of changes:*

- Ensure that the filing object is not pending before showing the item details. If it is pending or it does not contain a `data` object, use the default template.

*Reproducing the issue manually in the code:*

1. Add the following code into the `getFilings` return filter in the file `filingHistoryListStore.ts`:
```
filing.status = FilingStatus.PENDING
delete filing.data
```

2. Make sure `isStatusPaid` in the file `FilingTemplate.vue` always returns `true`.
3. Expand the details on the web app.

### DIFF
--- a/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
+++ b/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
@@ -50,7 +50,7 @@ export default class ConsentContinuationOut extends Vue {
 
   /** Whether the filing is complete. */
   get isFilingComplete (): boolean {
-    return !EnumUtilities.isStatusPending(this.filing) && Boolean(this.filing.data)
+    return EnumUtilities.isStatusCompleted(this.filing)
   }
 }
 </script>
--- a/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
+++ b/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
@@ -48,7 +48,7 @@ export default class ConsentContinuationOut extends Vue {
     return Boolean(this.filing.data?.order?.effectOfOrder)
   }
 
-  /** Whether the filing is pending. */
+  /** Whether the filing is complete. */
   get isFilingComplete (): boolean {
     return !EnumUtilities.isStatusPending(this.filing) && Boolean(this.filing.data)
   }
--- a/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
+++ b/src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
@@ -1,12 +1,14 @@
 <template>
   <FilingTemplate class="consent-continuation-out" :filing="filing" :index="index">
     <template #body>
-      <p class="mt-4">
-        This consent is valid <strong>until {{ expiry }} at 12:01 am Pacific time</strong>.
-      </p>
-      <p v-if="orderDetails" class="mt-4" v-html="orderDetails" />
-      <p v-if="fileNumber" class="mt-4 mb-0">Court Order Number: {{ fileNumber }}</p>
-      <p v-if="hasEffectOfOrder" class="mt-0">Pursuant to a Plan of Arrangement</p>
+      <div v-if="isFilingComplete">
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
```diff
@@ -45,6 +47,11 @@ export default class ConsentContinuationOut extends Vue {
   get hasEffectOfOrder (): boolean {
     return Boolean(this.filing.data?.order?.effectOfOrder)
   }
+
+  /** Whether the filing is complete. */
+  get isFilingComplete (): boolean {
+    return !EnumUtilities.isStatusPending(this.filing) && Boolean(this.filing.data)
```
severinbeauvais, line 53: I don't think we check whether the data object exists anywhere else, so I think you should leave it out here also.

The numerous places in this component where we use conditional chaining (`this.filing.data?.something`) handles de-reference errors in case data is absent. I think that's enough.

What do you think?

severinbeauvais (commented):
Thanks for the reproduction steps in the PR description.

Please update the app version (package.json), run `npm i`, and commit both package files.

severinbeauvais:
Is it the same fix for 16376? If you can fix that and also test it then let's do it at the same time. (I'll take any heat in doing it even though Dave didn't prioritize it.)

src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
leodube-aot: Yeah I wasn't too sure whether there were any other cases where a `filing` object might not have a `data` object, other than when the `status` is pending.

If that can't happen then it should be okay to remove the check on whether the data object exists.

leodube-aot:
> Is it the same fix for 16376? If you can fix that and also test it then let's do it at the same time. (I'll take any heat in doing it even though Dave didn't prioritize it.)

At a glance it appears to be the same, or similar enough issue. Looks like it's not related to a missing `data` object but to other properties that are not included in the `filing` object for that given status, so it defaults to [unknown]. I can fix and test that one as well.

src/components/Dashboard/FilingHistoryList/filings/ConsentContinuationOut.vue
severinbeauvais: There may be other cases where a data object is absent -- that's why conditional chaining is used here.

leodube-aot: It looks like conditional chaining in the component was already catching the case where the `data` object is absent, and using "[unknown]" as the fallback for the expiry, leading to this bug. I'm not 100% sure if just removing the explicit check will work correctly, since the bug was already occurring with the conditional chaining working as expected. 

Although, perhaps there is a way to conditionally show the consent for continuation out `body` slot that does not require the explicit check to see if the `body` object exists, I am just not sure what that would look like.

severinbeauvais: You now show body content only if the filing is complete, and I think we can assume the data is there in that case, so I think the bug is solved ...? (And you don't need to explicit check for the data object,,,?)

leodube-aot: I see, so would it make more sense then for me to use `EnumUtilities.IsStatusCompleted` rather than `!EnumUtilities.IsStatusPending`?

severinbeauvais: Whoops, I didn't notice that. Yes, for sure, good catch! See how the other filings do it.

leodube-aot: Latest commit removes the explicit check of `data` and uses `isStatusCompleted`. Also includes both package files with updated version.

severinbeauvais (approved):
LGTM

JazzarKarim (approved):
Looks good to me! 👍

PR #522, opened 2023-06-15
16048 Fix broken styling for detail component

leodube-aot:
*Issue #:* /bcgov/entity#16048

*Description of changes:*

I set the padding of the `v-textarea` to match the padding of the content inside the src/components/common/Certify.vue component (included in the screenshots). In that component, the content is wrapped with a `v-row` with `class` attributes to add padding. I did the same for the `v-textarea`, however it also works if it is wrapped in a plain old `div` with the same `class` attributes. I went with the `v-row` for consistency, however I also get if it doesn't make the most sense semantically and would be open to using a `div` instead.

I'll get UX to have a look after reviews to see if they have any preferences on the padding amount.

*Screenshot of changes:*

fix

### DIFF
--- a/src/components/common/DetailComment.vue
+++ b/src/components/common/DetailComment.vue
@@ -1,17 +1,19 @@
 <template>
   <v-card flat id="detail-comment-container">
-    <v-textarea
-      ref="textarea"
-      auto-grow
-      rows="5"
-      id="detail-comment-textarea"
-      :counter="maxLength"
-      :rules="rules"
-      :value="value"
-      :placeholder="placeholder"
-      :autofocus="autofocus"
-      @input="emitInput($event)"
-    />
+    <v-row no-gutters class="pl-4 pr-4 pt-2 pb-2">
+      <v-textarea
+        ref="textarea"
+        auto-grow
+        rows="5"
+        id="detail-comment-textarea"
+        :counter="maxLength"
+        :rules="rules"
+        :value="value"
+        :placeholder="placeholder"
+        :autofocus="autofocus"
+        @input="emitInput($event)"
+      />
+    </v-row>
   </v-card>
 </template>
 
--- a/src/components/common/DetailComment.vue
+++ b/src/components/common/DetailComment.vue
@@ -1,19 +1,17 @@
 <template>
   <v-card flat id="detail-comment-container">
-    <v-row no-gutters class="pl-4 pr-4 pt-2 pb-2">
-      <v-textarea
-        ref="textarea"
-        auto-grow
-        rows="5"
-        id="detail-comment-textarea"
-        :counter="maxLength"
-        :rules="rules"
-        :value="value"
-        :placeholder="placeholder"
-        :autofocus="autofocus"
-        @input="emitInput($event)"
-      />
-    </v-row>
+    <v-textarea
+      ref="textarea"
+      auto-grow
+      rows="5"
+      id="detail-comment-textarea"
+      :counter="maxLength"
+      :rules="rules"
+      :value="value"
+      :placeholder="placeholder"
+      :autofocus="autofocus"
+      @input="emitInput($event)"
+    />
   </v-card>
 </template>
 
--- a/src/views/Correction.vue
+++ b/src/views/Correction.vue
@@ -75,12 +75,16 @@
                 <p>Enter a detail that will appear on the ledger for this entity.</p>
                 <p class="black--text mb-0">{{defaultComment}}</p>
               </header>
-              <DetailComment
-                v-model="detailComment"
-                placeholder="Add a Detail that will appear on the ledger for this entity."
-                :maxLength="maxDetailCommentLength"
-                @valid="detailCommentValid=$event"
-              />
+              <v-card flat>
+                <div no-gutters class="px-4 py-2">
+                  <DetailComment
+                    v-model="detailComment"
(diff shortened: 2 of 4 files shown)


src/components/common/DetailComment.vue
```diff
@@ -1,17 +1,19 @@
...
+        rows="5"
+        id="detail-comment-textarea"
+        :counter="maxLength"
+        :rules="rules"
+        :value="value"
+        :placeholder="placeholder"
+        :autofocus="autofocus"
+        @input="emitInput($event)"
+      />
+    </v-row>
```
severinbeauvais, line 16: Thanks for the great description of changes in the PR 👍 

Vuetify renders a v-row as a div with some classes, so you're right -- underneath, they're equivalent.

In this case, I recommend a div instead, since you're not actually using any row functionality (such as columns within it).

With a div, you may have to explicitly add some Vuetify classes that "row" automatically adds, like flexbox. Ref: https://vuetifyjs.com/en/styles/flex/

severinbeauvais, line 16: This DetailComment component is used by various filings. You need to check all of them to ensure this extra padding doesn't mess those up.

PS Check out the [Vuetify spacing helper classes](https://vuetifyjs.com/en/styles/spacing/) -- you could use py-something here.

leodube-aot: That makes sense, I'll switch it to a `div`. I did a quick test comparing `v-row` and `div` and didn't notice any immediate differences but I'll double check to make sure that there are no Vuetify classes for `v-row` that need to be added.

leodube-aot: Good call, I only checked the `DetailComment` for corrections. I'll track down where else it is used and make sure the extra padding doesn't affect it. 

Also yeahI think it would look cleaner using `px-#` and `py-#` since the vertical/horizontal numbers are the same.

src/views/Correction.vue
```diff
@@ -75,12 +75,16 @@
...
                 <p class="black--text mb-0">{{defaultComment}}</p>
               </header>
-              <DetailComment
-                v-model="detailComment"
-                placeholder="Add a Detail that will appear on the ledger for this entity."
-                :maxLength="maxDetailCommentLength"
-                @valid="detailCommentValid=$event"
-              />
+              <v-card flat>
+                <div no-gutters class="px-4 py-2">
```
severinbeauvais, line 79: Do you need `no-gutters` here? I usually see it used for v-rows, which do something weird with left/right (negative) margins.

You might also be able to put the px/py classes right in the DetailComment element below.

severinbeauvais (approved):
The screenwhot and changes LGTM.

You'll need to rebase and update the app version again, sorry.

Wait for another review approval or 2 before merging.

leodube-aot:
So turns out it does affect the spacing in the other places where `DetailComment` is used. I brought the `div` with padding to the `Correction.vue` component to wrap `DetailComment` so that it does not affect the layout of the `DetailComment` in other places. This does not quite result in the same fix (see screenshot) and I needed to add a `v-card` around the entire thing to get it to work.

_Wrapped only in div_
wrap detail comment

_Wrapped in div and in v-card_
wrap in card

src/views/Correction.vue
leodube-aot: Just removed `no-gutter` and it didn't change anything. I'll add a new commit removing it.

I think the px/py classes have to be outside the DetailComment to not affect it in other places, see my comment below for more details.

leodube-aot: I misunderstood I see what you mean by adding it right in the DetailComment. I added it and it works so I should be good to remove the div.

JazzarKarim (approved):
LGTM! 👍

JazzarKarim:
Btw Leo, not really sure if you used it before, you can run `/gcbrun` to generate a Temporary URL so that you can test your changes there.

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-522-zw0l23dz.web.app

leodube-aot:
> Btw Leo, not really sure if you used it before, you can run `/gcbrun` to generate a Temporary URL so that you and other Devs can test your changes there.

Thanks for the tip! I've never used `/gcbrun` before but I'll have to start using it more often

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-522-zw0l23dz.web.app

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-522-zw0l23dz.web.app

severinbeauvais:
> > Btw Leo, not really sure if you used it before, you can run `/******` to generate a Temporary URL so that you and other Devs can test your changes there.
> 
> Thanks for the tip! I've never used `/******` before but I'll have to start using it more often

Careful with that keyword -- just having it in a comment runs it. "That which shall not be named."

JazzarKarim:
Looks Great!
image

I just used the temporary URL and appended the identifier to it, like this: https://business-filings-dev--pr-522-zw0l23dz.web.app/BC0870818/

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-522-zw0l23dz.web.app

leodube-aot:
> Wait for another review approval or 2 before merging.

I don't believe I have merge access yet so up to you on this.

Also I see what you mean about being carful with that keyword haha

PR #523, opened 2023-06-15
16763 fixed Corrected Filing Date format

severinbeauvais:
*Issue #:* bcgov/entity#16763

*Description of changes:*
    - app version = 6.5.16
    - set explicit correctedFilingDate type (ie, ISO date Pacific)
    - fixed local correction correctedFilingDate format
    - fixed Corrected Filing Date when building (draft) correction filing
    - deleted a console.log left over from testing
    - changed submittedDate in unit tests to GMT format

### DIFF
--- a/src/interfaces/correction-filing-interface.ts
+++ b/src/interfaces/correction-filing-interface.ts
@@ -1,4 +1,6 @@
 import { CorrectionTypes } from '@/enums'
+import { IsoDatePacific } from '@/interfaces'
+
 /** Incorporation Application filing loaded from / saved to the Legal API. */
 export interface CorrectionFilingIF {
   header: {
@@ -21,7 +23,7 @@ export interface CorrectionFilingIF {
   correction: {
     correctedFilingId: string
     correctedFilingType: string
-    correctedFilingDate: string
+    correctedFilingDate: IsoDatePacific
     type?: CorrectionTypes
     comment: string
   }
--- a/src/mixins/filing-mixin.ts
+++ b/src/mixins/filing-mixin.ts
@@ -91,6 +91,7 @@ export default class FilingMixin extends DateMixin {
    * @returns the filing body
    */
   buildCorrectionFiling (correctedFiling: any, correctionType: CorrectionTypes): CorrectionFilingIF {
+    const submittedDate = new Date(correctedFiling.submittedDate)
     const correctionFiling: CorrectionFilingIF = {
       header: {
         date: this.getCurrentDate,
@@ -103,7 +104,7 @@ export default class FilingMixin extends DateMixin {
       },
       correction: {
         comment: '',
-        correctedFilingDate: correctedFiling.submittedDate,
+        correctedFilingDate: this.dateToYyyyMmDd(submittedDate),
         correctedFilingId: correctedFiling.filingId,
         correctedFilingType: correctedFiling.name,
         type: correctionType
--- a/src/views/Correction.vue
+++ b/src/views/Correction.vue
@@ -669,7 +669,7 @@ export default class Correction extends Mixins(CommonMixin, DateMixin, EnumMixin
     }
 
     const data: any = {
-      correction: {
+      [FilingTypes.CORRECTION]: {
         correctedFilingId: this.correctedFilingId,
         correctedFilingType: this.origFiling.header.name,
         correctedFilingDate: this.dateToYyyyMmDd(this.apiToDate(this.origFiling.header.date)),
--- a/src/views/Correction.vue
+++ b/src/views/Correction.vue
@@ -669,7 +669,7 @@ export default class Correction extends Mixins(CommonMixin, DateMixin, EnumMixin
     }
 
     const data: any = {
-      [FilingTypes.CORRECTION]: {
+      correction: {
         correctedFilingId: this.correctedFilingId,
         correctedFilingType: this.origFiling.header.name,
         correctedFilingDate: this.dateToYyyyMmDd(this.apiToDate(this.origFiling.header.date)),
@@ -681,7 +681,7 @@ export default class Correction extends Mixins(CommonMixin, DateMixin, EnumMixin
     // NB: a correction to a correction is applied to the original data
     const filing = Object.assign({}, header, business, data)
     try {
-      let ret
+      let ret = null
       if (this.filingId > 0) {
         // we have a filing id, so update an existing filing
         ret = await LegalServices.updateFiling(this.getIdentifier, filing, this.filingId, isDraft)
--- a/tests/unit/DetailsList.spec.ts
+++ b/tests/unit/DetailsList.spec.ts
@@ -19,7 +19,7 @@ describe('Details List', () => {
     title: 'Annual Report (2018)',
     filingId: 63958,
     submitter: 'John Doe',
-    submittedDate: '2018-12-12',
+    submittedDate: 'Wed, 12 Dec 2018 12:00:00 GMT',
     paymentToken: null,
     isPaperOnly: true,
     comments: []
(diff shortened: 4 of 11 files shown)


src/interfaces/correction-filing-interface.ts
```diff
@@ -21,7 +23,7 @@ export interface CorrectionFilingIF {
   correction: {
     correctedFilingId: string
     correctedFilingType: string
-    correctedFilingDate: string
+    correctedFilingDate: FormattedDateTimeGmt
```
severinbeauvais, line 26: FormattedDateTimeGmt is an alias to a string, but the name gives us extra meaning.

In shared repo:
```
/**
  * Formatted date-time string with explicit GMT timezone (ie, UTC).
  * @example "Tue, 19 Apr 2022 16:40:18 GMT"
  */
export type FormattedDateTimeGmt = string
```

src/views/Correction.vue
```diff
@@ -669,10 +669,10 @@ export default class Correction extends Mixins(CommonMixin, DateMixin, EnumMixin
     }
 
     const data: any = {
-      [FilingTypes.CORRECTION]: {
```
severinbeauvais, line 672: There is only the one filing type here, so I decided to be explicit instead of generic.

JazzarKarim (approved):
LGTM

tests/unit/DetailsList.spec.ts
```diff
@@ -19,7 +19,7 @@ describe('Details List', () => {
     title: 'Annual Report (2018)',
     filingId: 63958,
     submitter: 'John Doe',
-    submittedDate: '2018-12-12',
+    submittedDate: 'Wed, 12 Dec 2018 12:00:00 GMT',
```
severinbeauvais, line 22: The old data is not a valid GMT string. The new data is.

src/interfaces/correction-filing-interface.ts
```diff
@@ -21,7 +23,7 @@ export interface CorrectionFilingIF {
   correction: {
     correctedFilingId: string
     correctedFilingType: string
-    correctedFilingDate: string
+    correctedFilingDate: IsoDatePacific
```
severinbeauvais, line 26: IsoDatePacific  is an alias to a string, but the name gives us extra meaning.

In shared repo:
```
/**
 * ISO date string (YYYY-MM-DD), assumed to be in Pacific timezone.
 * @example "2021-12-31"
 */
export type IsoDatePacific = string
```

tests/unit/FilingHistoryList1.spec.ts
```diff
@@ -772,7 +772,7 @@ describe('Filing History List - redirections', () => {
     business: { identifier: 'CP1002587', legalName: undefined, legalType: 'CP' },
     correction: {
       comment: '',
-      correctedFilingDate: 'Wed, 24 May 2023 18:02:16 GMT',
```
severinbeauvais, line 775: This and line 885 were the wrong format, but the tests didn't care.

severinbeauvais, line 775: This and line 885 were the wrong format, but the tests didn't catch this because we're mocking the API response (and it's the API that complains about the wrong format.

tests/unit/LegalObligation.spec.ts
```diff
@@ -23,12 +23,12 @@ const newIncorporationFiling = [
   {
     availableOnPaperOnly: false,
     displayName: 'Incorporation Application',
-    effectiveDate: new Date('2019-06-02T19:22:59.003777+00:00'),
```
severinbeauvais, line 26: This is wrong since effectiveDate should be a string, not a Date object.

Ditto for submittedDate below.

tests/unit/DetailsList.spec.ts
severinbeauvais: Explanation:

Submitted Date (and Effective Date), both in the filing data from the Legal API, are GMT strings (as above).

Corrected Filing Date is based on the submitted date of the filing being correct, but Corrected Filing Date must be a JSON "date", such as YYYY-MM-DD.

Correction schema:
https://github.com/bcgov/business-schemas/blob/1d00db747741857c3919e31f9ee37280a16295cb/src/registry_schemas/schemas/correction.json#L57

Defintion of JSON schema date:
Clipboard - June 15, 2023 10_05 AM

src/views/Correction.vue
severinbeauvais: Reverted (since the old code was consistent with the other local filings). Resolving.

src/mixins/filing-mixin.ts
```diff
@@ -103,7 +104,7 @@ export default class FilingMixin extends DateMixin {
       },
       correction: {
         comment: '',
-        correctedFilingDate: correctedFiling.submittedDate,
+        correctedFilingDate: this.dateToYyyyMmDd(submittedDate),
```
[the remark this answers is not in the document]
severinbeauvais: First, Vysakh confirmed that the Legal API does nothing with this value. We probably save it as part of the corrected filing's data for better traceability only.

Second, here are our time formats and their meanings:
https://github.com/bcgov/bcrs-shared-components/blob/main/src/interfaces/date-interfaces.ts

Additionally, we sometimes use `new Date('2023-06-15 12:34:56 GMT')`. Note that this is different from `new Date('2023-06-15 12:34:56 PDT')`. We use this format because the Date constructor understands this date-time format consistently. Other formats may or may not work depending on Node/browser/locale.

[ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

Just to make date-times more complicated, in addition to timezone offset, we also have Pacific Daylight Time (PDT) vs Pacific Standard Time (PST).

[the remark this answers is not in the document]
severinbeauvais: To actually answer your question, if we save 2023-06-15 then the Legal API should understand this to mean "June 15, 2023 Pacific".

If the API needs a time then it should append 00:00:00.

Example 1: "2023-06-15T00:00:00 PDT"
Example 2: "2023-06-15T07:00:00 GMT"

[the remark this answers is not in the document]
severinbeauvais: The same thing should apply here.

If you're seeing `2023-06-15:T00:00:00+00:00` in the database then the API did not save the date as a correct UTC/GMT value.

If you save `2023-06-15` then the API should return a value that means "June 15 Pacific". It is not up to the UI to convert `2023-06-15:T00:00:00+00:00` to "June 15 Pacific".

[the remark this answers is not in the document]
severinbeauvais: The same thing should apply here.

If you're seeing `2023-06-15:T00:00:00+00:00` in the database (or API response) then the API is in error.

If you save `2023-06-15` then the API should return a value that means "June 15 Pacific". It is not up to the UI to convert `2023-06-15:T00:00:00+00:00` to "June 15 Pacific".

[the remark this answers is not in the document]
severinbeauvais: If the API insists on returning a date-time (instead of just a date), then it should look like `2023-06-15:T70:00:00+00:00` (note the 7h offset to convert from GMT to Pacific).

vysakh-menon-aot (approved):

src/mixins/filing-mixin.ts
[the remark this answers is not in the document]
JazzarKarim: OK, noted Sev. Thanks for clearing that up!

JazzarKarim (approved):

src/mixins/filing-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: One more thing.

For some fields, the Legal API actually does return YYYY-MM-DD, which is understood to be Pacific. (No time.)

PR #525, opened 2023-06-19
16552 Added new eslint presets

JazzarKarim:
*Issue #:* /bcgov/entity#16552

*Description of changes:*
- Added new ESLint presets and implemented all the necessary changes accordingly.


.eslintrc.js
```diff
@@ -13,15 +23,18 @@ module.exports = {
     'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     'max-len': ['warn', { code: 120 }],
     'vue/attribute-hyphenation': 'off',
+    'vue/multi-word-component-names': 'off',
```
JazzarKarim, line 26: This is to fix line 258 in `Dashboard.vue`.

Reference: https://stackoverflow.com/questions/71205264/component-name-temp-should-always-be-multi-word-vue-multi-word-component-names

.eslintrc.js
```diff
@@ -13,15 +23,18 @@ module.exports = {
...
     'vue/no-deprecated-filter': 'warn',
     'vue/no-deprecated-slot-scope-attribute': 'warn',
     'vue/no-deprecated-v-on-native-modifier': 'warn',
     'vue/no-v-for-template-key-on-child': 'warn',
     'vue/no-v-html': 'off',
+    'vue/no-v-model-argument': 'off',
```
JazzarKarim, line 32: This is to fix lines 78 and 79 in `AddStaffNotationDialog.vue`. 
addstaffnotaiondialog v-model directive require no argument

Reference: https://stackoverflow.com/questions/72952665/vue-no-v-model-argument-v-model-directives-require-no-argument-eslint-plugin-v

.eslintrc.js
```diff
@@ -13,15 +23,18 @@ module.exports = {
...
     'vue/v-on-event-hyphenation': 'off',
     '@typescript-eslint/ban-types': 'warn',
     '@typescript-eslint/no-empty-function': 'off',
-    '@typescript-eslint/no-explicit-any': 'off'
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-var-requires': 'off'
```
JazzarKarim, line 37: This is to fix lines 29 and 41 in `StoreCurrentAccount.spec.ts`. It's also for line 1 in `volar.config.js`.
no var requires storecurrentacc test and volarconfig

Reference: https://dev.to/msoup/fixing-require-statement-not-part-of-import-statementeslinttypescript-eslintno-var-requires-25n3

src/components/Dashboard/TodoList.vue
```diff
@@ -1349,11 +1451,12 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
         this.setCurrentFilingStatus(FilingStatus.NEW)
         this.$router.push({ name: Routes.ANNUAL_REPORT, params: { filingId: '0' } }) // 0 means "new AR"
         break
-      case FilingTypes.CONVERSION:
+      case FilingTypes.CONVERSION: { // Curly braces to fix "Unexpected lexical declaration in case block"
```
JazzarKarim, line 1454: ESLint was complaining about "Unexpected lexical declaration in case block".
todolist error

Adding curly braces fixes it. Reference: https://eslint.org/docs/latest/rules/no-case-declarations

src/App.vue
```diff
@@ -297,6 +325,16 @@ export default {
     }
   },
 
+  watch: {
```
JazzarKarim, line 328: The `watch` block is moved to before the lifecycle hooks. ESLint was complaining about that.
watch should be above

src/components/dialogs/AddStaffNotationDialog.vue
```diff
@@ -53,16 +68,16 @@
...
             <li>Use a white background and a legible font with contrasting font colour</li>
-            <li>PDF file type (maximum {{MAX_FILE_SIZE}} MB file size)</li>
+            <li>PDF file type (maximum {{ MAX_FILE_SIZE }} MB file size)</li>
           </ul>
 
           <div class="d-flex mt-4">
             <span class="font-weight-bold">Upload File</span>
             <FileUploadPdf
               ref="fileUploadRef"
+              :file="file"
```
JazzarKarim, line 78: The `.sync` modifier on `v-bind` is deprecated. ESLint was complaining about that. I tested this by trying to upload a PDF. Seems to be working 👍 

addstaffnotationdialogue  sync is deprecated

JazzarKarim:
Final terminal after fixing all the issues: 
final

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-525-qukvvnj7.web.app

.eslintrc.js
severinbeauvais: I agree -- turn this one off (not even a warning).

Please make sure all 3 UIs have the same ESLint presets and rule overrides.

.eslintrc.js
severinbeauvais: Please change these to warnings, as we will have to fix this before going to Vue/Vuetify 3.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Could you change it as recommended in the error message instead?

severinbeauvais:
I'm clearing up some SonarCloud Security Hotspots...

Can you look into this one, please? I'm curious what the "built-in Vue sanitization" is.

image

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Thanks for the change. Make sure you test with values going into the sub-component and back out (via same prop).

.eslintrc.js
severinbeauvais: ^^ Create new tickets if needed.

src/App.vue
```diff
@@ -314,7 +352,7 @@ export default {
   },
 
   /** Called just before this component is destroyed. */
-  beforeDestroy (): void {
+  beforeUnmount (): void {
```
severinbeauvais, line 355: Please test this.

You can add a `debugger` command in here to force a breakpoint, or you could add an alert (which stops execution).

(In Edit UI, I left it as beforeDestroy(). Hopefully that's correct.)

JazzarKarim: So I tested this with both `beforeUnmount` and `beforeDestroy` by disabling the `vue/no-deprecated-destroyed-lifecycle` rule. Unless I'm missing something, they seem to work the exact same way. Shall I keep it `beforeUnmount` or change it to `beforeDestroy` with the rule so that we are consistent with Edit UI?

severinbeauvais: Cool. Thanks for looking into it.

Keep it as `beforeUnmount` since that appears to be the future way. Ref: https://vuejs.org/guide/essentials/lifecycle.html#lifecycle-diagram

We can deal with Edit UI later. Interesting that it doesn't complain about `beforeDestroy` even though it doesn't have the rule override...?

src/components/AnnualReport/AGMDate.vue
```diff
@@ -14,27 +21,27 @@
             transition="scale-transition"
             max-width="290"
           >
-            <template v-slot:activator="{ on, attrs }">
+            <template #activator="{ on, attrs }">
```
severinbeauvais, line 24: Did you change this or did ESLint? Could you test it, please?

JazzarKarim:
> I'm clearing up some SonarCloud Security Hotspots...
> 
> Can you look into this one, please? I'm curious what the "built-in Vue sanitization" is.
> 
> image

So Sev, I've been looking into this. Apparently there is some sort of security risk for binding the `href` attribute. It's something called URL Injection (REF: https://medium.com/@isaacwangethi30/vue-js-security-6e246a7613da).

URL injection
URL injection 2

HOWEVER, I don't think it applies in our situation since the links we're binding to in the project are not user-provided URLs. What do you think?

src/components/Dashboard/FilingHistoryList/DetailsList.vue
```diff
@@ -18,15 +20,21 @@
...
     <v-list class="pb-0">
-      <v-list-item class="pl-0 pr-0 detail-body" v-for="(comment, index) in filing.comments" :key="index">
+      <v-list-item
+        v-for="(comment, index) in filing.comments"
+        :key="index"
+        class="pl-0 pr-0 detail-body"
+      >
         <v-list-item-content>
-          <v-list-item-title class="body-2">
+          <v-list-item-title class="text-body-2">
```
severinbeauvais, line 29: What is this change?

JazzarKarim: ESLint decided to add this. So, I went with it and fixed the test for it.

severinbeauvais: OK. I see it got changed in a few places. So the class exists? And it looks the same as before on screen?

src/components/Dashboard/FilingHistoryList/filings/LimitedRestorationExtension.vue
```diff
@@ -1,11 +1,15 @@
...
+    :filing="filing"
+    :index="index"
+  >
     <template #body>
       <h4>Extension of Limited Restoration</h4>
 
       <p>
         The period of restoration was successfuly extended and is active <strong>until
-        {{ expiryDate }}</strong>. At the end of the extended limited restoration period, the
+          {{ expiryDate }}</strong>. At the end of the extended limited restoration period, the
```
severinbeauvais, line 12: Rule applied incorrectly. Oh well, leave it.

src/components/AnnualReport/AGMDate.vue
JazzarKarim: Tested, working fine!
Screenshot (78)
tested 2

src/components/Dashboard/FilingHistoryList/DetailsList.vue
severinbeauvais: Must be a Vuetify 3 vs Vuetify 2 thing.

We need a way to remember to change this when we do upgrade to Vuetify 3. Any ideas?

src/App.vue
severinbeauvais: Confused? -> I don't see this same lint issue in Edit UI, even though it uses the old lifecycle event but doesn't have the rule to shut up the message. I'm confused, too, and suggesting we don't worry about it for now.

src/components/Dashboard/FilingHistoryList/DetailsList.vue
JazzarKarim: Shall I add a comment in the template?
`<!-- FUTURE: Change body-2 (Vuetify 2) to text-body-2 (Vuetify 3) -->`

severinbeauvais: My thoughts;
1. change it to a warning instead
or
2. leave a code comment to remove the rule
or
3. create a tech debt ticket for this, or add to an existing Vue/Vuetify 3 ticket

severinbeauvais: > Shall I add a comment in the template?

Not in the template, as that would be in many places. In the config maybe?

JazzarKarim: I like the first one! This way it'll be in our face whenever we run. This'll make sure that we remember once we do the upgrade.

severinbeauvais: Haha, fine by me :)

src/App.vue
JazzarKarim: Exactly! I've trying to figure out why this might be the case. Could it be since Filings UI uses Options API while Edit UI uses Class API? I really don't know.

src/components/dialogs/AddStaffNotationDialog.vue
JazzarKarim: Tested. Looks OK. I've been getting the `invalid PDF` error non-stop though. I'm assuming it's my fault as I probably am trying to upload invalid PDFs.

src/components/EntityInfo.vue
```diff
@@ -82,6 +95,7 @@ export default class EntityInfo extends Vue {
 
   /** Emits an event to indicate business is not in good standing. */
   @Emit('notInGoodStanding')
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   emitNotInGoodStanding (message: NigsMessage): void { /* no empty function */ }
```
severinbeauvais, line 99: I wonder if you can have multiple rules to disable on 1 line?

src/components/common/ForeignJurisdiction.vue
```diff
@@ -1,34 +1,45 @@
...
+          v-model="selectedCountryName"
           :items="getCountries()"
           item-text="name"
-          v-model="selectedCountryName"
           filled
           placeholder="Jurisdiction Country"
           :rules="countryRules"
-          immediate: true
+          immediate:
+          true
```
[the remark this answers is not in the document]
severinbeauvais: Or if "immediate" is a boolean (like "filled") then just its presence here means it's True.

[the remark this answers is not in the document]
severinbeauvais: Also on lines 41-42 below.

src/components/common/NameRequestInfo.vue
```diff
@@ -19,26 +22,44 @@
             <span class="val">{{ expirationDate }}</span>
           </li>
           <li id="status">
-            <v-icon v-if="state === NameRequestStates.APPROVED ||
-                          state === NameRequestStates.CONDITIONAL"
-              color="green" class="nr-status-icon">mdi-check</v-icon>
+            <v-icon
+              v-if="state === NameRequestStates.APPROVED ||
+                state === NameRequestStates.CONDITIONAL"
```
severinbeauvais, line 27: This is a good example why I prefer parentheses around more complex if statements for ease of reading/reviewing/understanding.

severinbeauvais: This is actually not a bad one. You can leave it as-is.

src/components/common/SummaryDirectors.vue
```diff
@@ -67,17 +87,29 @@
...
-                  <div class="address same-address" v-if="isBenBcCccUlc">
+                  <div
+                    v-if="isBenBcCccUlc"
+                    class="address same-address"
+                  >
                     <span v-if="isSame(director.deliveryAddress, director.mailingAddress)">
                       Same as Delivery Address
                     </span>
-                    <base-address v-else :address="director.mailingAddress" />
+                    <base-address
```
severinbeauvais, line 97: Could you rename this to `<BaseAddress` please? Also search for others in this file and globally.

src/components/common/SummaryOfficeAddresses.vue
```diff
@@ -17,16 +17,20 @@
       <!-- Registered Mailing Address -->
       <li class="address-list-container">
         <div class="meta-container">
-          <label></label>
-          <div class="meta-container__inner"
+          <label />
```
severinbeauvais, line 20: FYI, this (and line 55) is required for the CSS selectors at the bottom of this file.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: Now Github Actions is giving us a warning. Hopefully it will go away when we upgrade to Vue/Vuetify 3.

As this tests OK (but please try with a valid PDF ;) ), I'm OK to resolve this conversation.

src/components/dialogs/StaffPaymentDialog.vue
```diff
@@ -105,6 +115,7 @@ export default class StaffPaymentDialog extends Vue {
 
   // Pass updated staff payment data to parent.
   @Emit('update:staffPaymentData')
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   emitStaffPaymentData (staffPaymentData: StaffPaymentIF): void { /* no empty function */ }
```
severinbeauvais, line 119: Same here -- can both disabled rules be on 1 line?

src/shims-tsx.d.ts
```diff
@@ -3,9 +3,9 @@ import Vue, { VNode } from 'vue'
 declare global {
   namespace JSX {
     // tslint:disable no-empty-interface
-    interface Element extends VNode {}
+    type Element = VNode
     // tslint:disable no-empty-interface
```
severinbeauvais, line 7: Probably no lonoger need the tslint disable here or above.

src/views/AnnualReport.vue
```diff
@@ -39,54 +39,89 @@
     />
 
     <StaffPaymentDialog
+      v-model:staffPaymentData="staffPaymentData"
```
severinbeauvais, line 42: Please test this change.

If this doesn't work then we could use a value (prop) and event to keep staffPaymentData synchronized between THIS component and the StaffPaymentDialog sub-component.

severinbeauvais: Unless I've already raised this issue? 😕 

If this _is_ a problem then there are several places that this substitution was made.

src/views/AnnualReport.vue
```diff
@@ -279,13 +367,24 @@
...
+    >
+      <div class="buttons-left" />
 
       <div class="buttons-right">
-        <v-tooltip top color="#3b6cff" content-class="top-tooltip">
-          <template v-slot:activator="{ on }">
-            <div v-on="on" class="d-inline">
+        <v-tooltip
+          top
+          color="#3b6cff"
```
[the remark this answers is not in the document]
severinbeauvais: I see hex colors in other views, too. Would be nice to fix all.

src/views/ContinuationOut.vue
```diff
@@ -621,7 +687,7 @@ export default class ContinuationOut extends Mixins(CommonMixin, DateMixin,
    * or when user retries from Save Error dialog
    * or when user submits from Staff Payment dialog.
    */
-  async onClickFilePay (fromStaffPayment = false): Promise<void> {
+  async onClickFilePay (staffPayment = false): Promise<void> { // eslint-disable-line @typescript-eslint/no-unused-vars
```
severinbeauvais, line 690: I prefer to see this eslint disable on the previous line, please and thanks.

src/views/AnnualReport.vue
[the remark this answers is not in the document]
severinbeauvais: If all else fails, we could declare a new color (tooltipColor?) in main.ts along with the others.

src/views/Correction.vue
```diff
@@ -39,46 +39,70 @@
...
             <!-- Detail -->
             <section>
               <header>
-                <h2 id="correction-step-1-header">1. Detail</h2>
+                <h2 id="correction-step-1-header">
+                  1. Detail
+                </h2>
                 <p>Enter a detail that will appear on the ledger for this entity.</p>
-                <p class="black--text mb-0">{{defaultComment}}</p>
+                <p class="black--text mb-0">
```
severinbeauvais, line 98: Here's yet another way to set a color (see https://v2.vuetifyjs.com/en/styles/colors/).

In fact, this is probably wrong and should be a `grayX`, but ignore that for now.

src/views/Dashboard.vue
```diff
@@ -118,21 +149,39 @@
...
+                    mdi-pencil
+                  </v-icon>
                   <span>Change</span>
                 </v-btn>
               </header>
 
-              <div class="scrollable-container" style="max-height: 49rem">
+              <div
+                class="scrollable-container"
+                style="max-height: 49rem"
```
severinbeauvais, line 183: This is also declared on line 85. Please create a class for this instead (or add to scrollable-container CSS?).

severinbeauvais: Also line 219.

src/views/Dashboard.vue
```diff
@@ -219,7 +271,14 @@ export default {
     ProprietorPartnersListSm,
     StaffNotation,
     TodoList
-  },
+  }, // eslint-disable-line vue/multi-word-component-names
```
severinbeauvais, line 274: Should this be a global rule override instead?

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: By the way, there's a difference between `:file="file"` and the sync or v-model syntax.

The first binds a value going "in" to the sub-component only.

The second and third update the variable HERE (in this component) when the sub-component emits an update event.

severinbeauvais:
> > HOWEVER, I don't think it applies in our situation since the links we're binding to in the project are not user-provided URLs. What do you think?
> 
> I agree. I'll mark the ones where we provide the URL as "won't fix".

On further thought, it may possible to inject a different href here (or v-html in other cases): the app code may be editable in the browser (inspecting sources). But I'm not going to worry about this right now.

src/App.vue
severinbeauvais: You're probably right 😃

src/components/Dashboard/FilingHistoryList/DetailsList.vue
JazzarKarim: Sev, circling back this. If I have the rule as 'warn', whenever I run `npm run lint`, ESLint is changing this class. It's determined to change it. I guess I'll be turning off the rule and adding a code comment in the eslint config file.

severinbeauvais: Haha, darn. I'm surprised there isn't a way to override that behaviour.

I'm OK with a comment in the config file.

src/views/AnnualReport.vue
JazzarKarim: OK, great great catch. 
missing required prop

I'm thinking of turning off the rule `vue/no-deprecated-v-bind-sync`. Moreover, I think this change was also done somewhere else. I'll be reverting this change everywhere.

src/views/Dashboard.vue
JazzarKarim: This rule only applies for this file so I opted to have it only in this file. However, I could add it to the eslint config file as a global rule. Your call Sev!

src/views/AnnualReport.vue
severinbeauvais: I think "deprecated" rules should be changed to warnings possible, since we'll eventually have to replace the deprecated feature by something new. (Or turned off with a comment in the config file.)

src/views/Dashboard.vue
severinbeauvais: Weird that this rule only applies to Options API components...?

However, I think that, globally, we do want to allow multi-word component names.

src/views/AnnualReport.vue
JazzarKarim: When I set the rule to 'warn', `npm run lint` is changing them. I have the rule set to off in my latest commit. However, I'll look into fixing this without having the rule.

JazzarKarim:
I totally forgot about the code smells that SonarCloud complain about whenever I have empty functions 😞 I removed the `empty function` comments in the code since there is a rule that's set to off in regards to that.

src/views/AnnualReport.vue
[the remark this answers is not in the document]
JazzarKarim: So, I had a look where actual hex values are being used in the code. Turns out, only for the tooltip color. So, I declared a new color like you proposed Sev since I think that's the nicest way of going on about this.

src/components/common/SummaryDirectors.vue
severinbeauvais: Thanks.

src/components/common/ForeignJurisdiction.vue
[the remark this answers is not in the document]
severinbeauvais: What does "immediate" do?

src/assets/styles/theme.scss
```diff
@@ -70,6 +70,7 @@ $app-dk-blue: #38598a; // same as the Vuetify theme appDkBlue
 $app-red: #d3272c;
 $app-green: #1a9031;
 $app-alert: #F8661A;
+$app-tooltip-color: #3b6cff;
```
severinbeauvais, line 73: Is this variable used anywhere?

Also, "color" in the name is redundant :) How about just app-tooltip?

src/shims-tsx.d.ts
severinbeauvais: Thanks.

src/views/ContinuationOut.vue
severinbeauvais: Thx

src/views/Dashboard.vue
severinbeauvais: Thanks.

src/assets/styles/theme.scss
JazzarKarim: No actually, it's not. I just wanted to be consistent with the other declarations in `main.ts`. Also, agreed.

I think I'll just remove the entire declaration. In case we need it down the line, we can re-add it.

src/views/Dashboard.vue
severinbeauvais: Oh, looks like this rule was already off ...?

JazzarKarim: Yup, not sure why I added it in the first place 🤣 Must be my brain imagining things.

src/views/AnnualReport.vue
severinbeauvais: I like the comment for the deprecated classes rule. Please add a similar comment for the deprecated sync that you turned off.

src/components/common/ForeignJurisdiction.vue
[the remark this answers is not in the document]
JazzarKarim: "Passing in immediate: true in the option will trigger the callback immediately with the current value of the expression". Ref: https://stackoverflow.com/questions/45354027/vue-js-how-to-fire-a-watcher-function-when-a-component-initializes

However, I think we don't actually need it for this component. I'll be testing it.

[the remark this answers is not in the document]
severinbeauvais: Yes, I use immediate for watchers sometimes, but this is in a component !? Maybe a typo.

src/components/EntityInfo.vue
severinbeauvais: Obsolete comment.

src/components/dialogs/StaffPaymentDialog.vue
severinbeauvais: Obsolete comment.

src/views/AnnualReport.vue
severinbeauvais: Thanks for both comments :)

src/components/common/ForeignJurisdiction.vue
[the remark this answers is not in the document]
JazzarKarim: Agreed. Removed it! Tested the component with out, works perfectly fine.

src/components/dialogs/AddStaffNotationDialog.vue
severinbeauvais: This issue is handled in another comment below.

severinbeauvais: @chenhongjing I think you'll have to do the same thing Karim did.

src/components/common/ForeignJurisdiction.vue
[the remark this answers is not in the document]
severinbeauvais: Thanks for removing.

severinbeauvais (approved):
I ran out of comments 😁 

Please add some comment for Riyaz in the ticket so he can focus his testing (not "please test everything").

Or, you could help Riyaz out by testing Hongjing's Create UI changes and she could test yours? Connect with R and H as needed.

chenhongjing (approved):
I don't have any more comment either. Great job, Karim! 👍 
And thanks Sev! I've learned more when I read through each comment. 😆

JazzarKarim:
@chenhongjing @bennypc Please let me know if there are still any outstanding issues. Thanks 😄

PR #531, opened 2023-07-21
#16281 adding todo items for affiliation invitation - access request

hfekete:
*Issue #:16281* [/bcgov/entity16281](https://github.com/bcgov/entity/issues/16281)

Adding todo items on the entity panel for open org affiliation invitations (request access type)


hfekete:
Pushing the PR first to get first feedback and answer any comments needed.
Will add TCs for added features in the next commits.

src/components/Dashboard/TodoList.vue
```diff
@@ -560,11 +593,13 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
   panel: number = null // currently expanded panel
   checkTimer: number = null
   inProcessFiling: number = null
+  FilingStatus = FilingStatus
```
severinbeauvais, line 596: Move this to line 621-622 (and make is readonly).

src/components/Dashboard/TodoList.vue
```diff
@@ -526,6 +558,7 @@ import { useBusinessStore, useConfigurationStore, useFilingHistoryListStore, use
 
 @Component({
   components: {
+    AffiliationInvitationDetails,
```
severinbeauvais, line 561: Move this to "components" on line 566-567.

src/components/Dashboard/TodoList.vue
```diff
@@ -796,6 +833,70 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
...
+        fromOrgName: affiliationInvitation.fromOrg.name,
+        additionalMessage: affiliationInvitation.additionalMessage
+      }
+    }
+
+    return newTodo
+  }
+
+  /** Loads Org Affiliations invitation todo **/
+  async loadAffiliationInvitationsTodo () {
```
severinbeauvais, line 876: Can this follow the existing design of the Legal API creating any todo tasks?

hfekete: For sure, but can you give me a bit more details what should be changed and/or refer me to the specific parts of the code you want my code to be similar to?

Are you possibly referring to the loadTodoItem method and adding "Affiliation Invite" as a task first ( and to create a new FilingType) or to something else ?

Thanks!

severinbeauvais: I believe all todo items come from the Legal API. The Legal API dynamically creates todo tasks as needed (eg, an Annual Report is due).

Here, the Legal API would add a todo task if an invitation needs to be approved. The TodoList would simply display it (may need small TodoList updates to handle display).

@argush3 @vysakh-menon-aot Could one of you link to the code in Legal API where todo tasks are created?

tests/unit/TodoList2.spec.ts
```diff
@@ -70,6 +70,11 @@ describe('TodoList - common expansion panel header tests', () => {
     const sandbox = sinon.createSandbox()
     const axiosGet = sandbox.stub(axios, 'get')
 
+    // feature flag override
+    sandbox.stub(utils, 'GetFeatureFlag')
+      .withArgs('enable-affiliation-invitation-request-access')
+      .returns(true)
```
severinbeauvais, line 76: Nice! 👍

tests/unit/TodoList2.spec.ts
```diff
@@ -57,6 +61,95 @@ describe('TodoList - common expansion panel header tests', () => {
...
+
+    const sandbox = sinon.createSandbox()
+    const axiosGet = sandbox.stub(axios, 'get')
+
+    // feature flag override
+    sandbox.stub(utils, 'GetFeatureFlag')
+      .withArgs('enable-affiliation-invitation-request-access')
+      .returns(true)
+
+    axiosGet.withArgs(affiliationInvitationsForEntityUrl).returns(new Promise(
```
severinbeauvais, line 78: Dunno where my previous comment went, but...

I recommend stubbing the auth service call directly instead of the axios call. (This test suite is not responsible for testing the services.)

hfekete: That is fair, replaced the stub from axios to auth-service.

src/components/Dashboard/TodoList.vue
hfekete: I am not fully familiar with all the APIs that exist as part of the system. I did work on Auth API, and currently all the data in the backend will be stored in the Auth API. 
I am assuming these two are not the same backend projects. And if so, I would prefer to avoid adding proxy backend calls.

I believe that avoiding proxy call would reduce complexity, improve maintenance and simplicity of understanding where data is coming from.

Also, I tried to put the call to load affiliation invitations at the place where data is getting loaded into the control, as my understanding is that they are not to be part of any filings or anything else other then todo reminder on that specific page.

That being said, if it is mandatory to go through Legal API to provide todo data I am not against it, just think that approach to getting data from the source API might be more straightforward.

Thank you!

hfekete: Maybe it would make sense to discuss this offline and see what is the desired approach for this specific case. 

I could get better understanding of the API relations to this frontend and their role and I can share why I chose this approach. 

I am open to advice and can update my approach/code to fit the current repo both in style and data flow.

Thanks!

severinbeauvais: I'm hoping Argus or Vysakh can provide a link for you to look at. We can discuss next week if needed.

The basic principle is that the UI should be as dumb as possible.

kialj876: hey Sev thats a good point, but I think from a backend perspective it really doesn't make sense to combine auth/lear data in the legal-api response. In general I think the apis should be as decoupled as possible and the UI can decide what it needs from each one.  I think that makes it simpler when designs change or new use cases come up and makes a lot more sense from a 3rd party api perspective too

severinbeauvais: OK, let's try to have the Todo List fetch data from both APIs...

1. do affiliation invitations have a status (eg, new)?
2. do affiliation invitations block other filings, ie, only allow 1 active filing active at a time?
3. do any other scenarios block affiliation invitations,ie, do we need `if (isAllowed(affilliation_invitation))`?

src/components/Dashboard/TodoList.vue
```diff
@@ -58,7 +58,7 @@
 
                 <!-- blue details button -->
                 <v-btn
-                  v-else-if="showDetailsBtnBlue(item)"
+                  v-else-if="showDetailsBtnBlue(item) || isAffiliationInvitation(item)"
```
severinbeauvais, line 61: Instead of parallel conditions throughout this (already-ugly) template, could the affiliation condition be added to `showDetailsBtnBlue`?

src/components/Dashboard/TodoList.vue
```diff
@@ -684,6 +719,8 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
       }
     }
 
+    await this.loadAffiliationInvitationsTodo()
```
severinbeauvais, line 722: Do you actually want to load affiliation invitations before the other tasks?

src/components/Dashboard/TodoList.vue
```diff
@@ -796,6 +833,70 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
...
+      })
+  }
+
+  _buildTodoItemIfFromAffiliationInvitation (affiliationInvitation) {
+    const newTodo: TodoItemIF = {
+      draftTitle: undefined,
+      enabled: true,
+      filingId: -1, // not a filing
+      name: undefined,
+      order: 0,
```
severinbeauvais, line 861: So all affiliation invitations have order=0, thus displaying above the other tasks maybe? This feels hacky,

src/components/Dashboard/TodoList.vue
```diff
@@ -796,6 +833,70 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
...
+      subtitle: `From: ${affiliationInvitation.fromOrg.name}`,
+      status: FilingStatus.AFFILIATION_INVITATION_PENDING,
+      title: 'Request for authorization to mange this business',
```
severinbeauvais, line 864: Typo

src/components/Dashboard/TodoList/AffiliationInvitationDetails.vue
```diff
@@ -0,0 +1,46 @@
...
+      The following account is requesting authorization to manage Registries activities for
+      {{ getEntityName || 'Unknown name' }}:
+    </p>
+    <p><strong>{{ affiliationInvitationTodo.affiliationInvitationDetails.fromOrgName }}</strong></p>
+    <p>
+      Only authorize this account if you recognize it.
+    </p>
+    <p>
+      Allowing {{ affiliationInvitationTodo.affiliationInvitationDetails.fromOrgName }} to manage all Registries
+      activities for {{ getEntityName || 'Unknown name' }}, will
```
severinbeauvais, line 15: Should not have comma before "will".

tests/unit/TodoList1.spec.ts
```diff
@@ -255,6 +255,7 @@ describe('TodoList - UI', () => {
     const wrapper = mount(TodoList, { vuetify, mixins: [AllowableActionsMixin] })
     const vm = wrapper.vm as any
     await Vue.nextTick()
+    await flushPromises()
```
severinbeauvais, line 258: Is this required because of the multiple asyncs in loadData()?

Theoretically this additional wait shouldn't be needed, but if you can't figure this out quickly then I recommend leaving it and remove the await nextTick() calls (you really shouldn't need both; if you do then something's wrong with the code or test).

tests/unit/TodoList2.spec.ts
```diff
@@ -57,6 +61,95 @@ describe('TodoList - common expansion panel header tests', () => {
...
+    expect(wrapper.findAll('.todo-item').length).toEqual(1)
+    // verify title starts with
+    expect(wrapper.find('.list-item__title').text())
+      .toContain('Request for authorization to mange this business')
+
+    // verify subtitle has information
+    expect(wrapper.find('.todo-subtitle').text()).toBe('From: Tree Frog Design Inc.')
+
+    // verify both buttons authorize and do not authorize are visible
+    const actionBtns = wrapper.findAll('.width-45-float-right')
```
severinbeauvais, line 140: It would be better to assign a classname to your buttons that refers to its purpose, instead of this styling classname. I think you can just rename the existing class -- it will serve as both identifying the type of button and setting some CSS for it.

(I'm trying to avoid confusion between styling classes and "functional" classes. Styling classes should be used for styling only. Functional classes can help identify what the button is for and also include styling/CSS if needed. Another way to look at it is that "all affiliation buttons have width 45% and float-right, but not all width-45-float-right buttons are necessarily affiliation buttons".)

src/components/Dashboard/TodoList.vue
```diff
@@ -796,6 +833,70 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
...
+  /** Loads Org Affiliations invitation todo **/
+  async loadAffiliationInvitationsTodo () {
+    // feature-flag-it
+    if (!GetFeatureFlag('enable-affiliation-invitation-request-access')) {
+      return
+    }
+
+    // load all the invitations here and push them into todo items
+    const affiliationInvitations =
+      await AuthServices.fetchAffiliationInvitations(this.getAuthApiUrl, this.getIdentifier)
```
seeker25, line 884: This should be moved into the service itself, using await and .then afterwards seems fairly weird

src/components/Dashboard/TodoList.vue
severinbeauvais: OK, thanks. Resolving. (I have made other comments to make this fit more cleanly with the existing code.)

hfekete: Agree with @kialj876 
1) not that it would matter (only pending status, for specific type of the affiliation invitation is displayed)
2) does not block any filling, it is just something owner has as a todo item while invite is 'pending'
3) no scenarios that would block affiliation invitations (affiliation invitation is basically a request from someone to manage the business; access to the managing org)

tests/unit/TodoList1.spec.ts
hfekete: Yes, it is required because of the await in loadData. 

I believe we could do without the nextTick() but not without flushPromises now.

What I understand is that nextTick waits for next updates of the DOM for vue (which can happen multiple times) while flushPromises waits for clearing of all the promises currently in queue (including the DOM update)

OK will remove nexTick() calls

src/components/Dashboard/TodoList.vue
hfekete: Updated, to set order as array size. And as you previously mentioned, I moved the code to load the todos in front of the other items.

src/components/Dashboard/TodoList/AffiliationInvitationDetails.vue
```diff
@@ -0,0 +1,46 @@
...
+      allow it to do the following (including but not limited to):
+    </p>
+    <ul>
+      <li>Authorization of accounts to manage the business,</li>
+      <li>Appointment and cessation of directors,</li>
+      <li>Dissolution of the business,</li>
+      <li>Filing of Annual Reports, and</li>
+      <li>Changing the records office of the business.</li>
```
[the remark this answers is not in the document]
hfekete: Updated with new wording from UX team.

src/components/Dashboard/TodoList.vue
severinbeauvais: If you set order as array size then these will appear last (if the list is sorted), but if you load these before the other todos then do you actually want them first?

src/enums/filingStatus.ts
```diff
@@ -11,4 +11,5 @@ export enum FilingStatus {
   PENDING = 'PENDING',
   PENDING_CORRECTION = 'PENDING_CORRECTION',
   WITHDRAWN = 'WITHDRAWN',
+  AFFILIATION_INVITATION_PENDING = 'AFFILIATION_INVITATION_PENDING',
```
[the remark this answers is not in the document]
hfekete: Updated, will check if the invitation details `id` is set in the todo item.

src/components/Dashboard/TodoList.vue
hfekete: will update code a bit to be more clear

src/enums/filingStatus.ts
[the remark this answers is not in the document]
severinbeauvais: But `id` may not be unique to affiliation invitations...

Do you want me to add a value to the Filing Types enum? (AFFILIATION_INVITATION)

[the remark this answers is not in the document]
hfekete: I added `affiliationInvitationDetails` in the todo item:
So I am checking this: `return !!item.affiliationInvitationDetails?.id`

I wanted to avoid adding AFFILIATOIN_INVITATION as filing type because it is not a filing, like you know. Though it is requested to be displayed along filing todos. 

Is the check I am doing good enough (that part should be unique to the affiliation invitation)

src/components/Dashboard/TodoList.vue
hfekete: Moved the call to be before adding other items.

src/components/Dashboard/TodoList.vue
hfekete: I'll double check with the UX on should they be ordered first, as I do not have the explicit requirement for this page (though on the other pages affiliation invitations were always on top.)

src/enums/filingStatus.ts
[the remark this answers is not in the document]
severinbeauvais: Oh I see -- `id` is a property in a sub-object.

Wouldn't it be easier/cleaner to check item.name? I'm OK with adding it to the filing type because it's kind of handled like a filing/task.

Or, why not just check if item.affiliationInvitationDetails is not falsy?

src/components/Dashboard/TodoList.vue
severinbeauvais: OK, I'll leave this with you. Resolving.

src/enums/filingStatus.ts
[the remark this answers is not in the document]
hfekete: Yeah, it makes sense, will update to check `affiliationInvitationDetails` is not falsy

severinbeauvais (approved):
LGTM but please wait for more reviews before merging.

src/components/Dashboard/TodoList.vue
```diff
@@ -796,6 +834,72 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
...
+      }
+
+      return newTodo
+    }
+
+    // load all the invitations here and push them into todo items
+    const response =
+      await AuthServices.fetchAffiliationInvitations(this.getAuthApiUrl, this.getIdentifier)
+        .catch((err) => {
+          console.log('Error fetching affiliation invitations for todo', err) // eslint-disable-line no-console
```
kialj876, line 888: Should this be a console.error or console.warn ?

kialj876 (approved):
LGTM too 👍

src/components/Dashboard/TodoList.vue
severinbeauvais: Or a dialog to let the user know?

hfekete: UX has no specific scenario for this, when I previously asked they said we could potentially display a generic modal for that page.

I could provide alertbox / modal saying something like following:

"Failed to fetch affiliation invite TODOs." 

Would that work ?
Or if someone can suggest example how failed API calls are handled right now ? 
I would prefer not to fail loading entire page for it though, as it might retrieve other TODO items.

severinbeauvais: The code has numerous examples of a dialog for API errors.

As this isn't a user error, I'd be ok with an alert() as a last resort, but I'd prefer to see a dialog. You probably need to create a new one as this is a specific error.

hfekete: Agreed, dialog looks much better and more user friendly.

Added preview of how the dialog looks like.
image

Also, I named it "Generic error dialog" so any error like this could reuse it by using template slots. Added title and message, and example inside the dialog component if no title and message is displayed.  
Hope that this is acceptable.

src/components/Dashboard/TodoList.vue
```diff
@@ -886,6 +911,7 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
       await AuthServices.fetchAffiliationInvitations(this.getAuthApiUrl, this.getIdentifier)
         .catch((err) => {
           console.log('Error fetching affiliation invitations for todo', err) // eslint-disable-line no-console
+          this.fetchAffiliationInvitationsErrorDialog = true
```
severinbeauvais, line 914: Do you ever have to reset this to false? (Example: see resetCancelPaymentErrors().)

Or does this dialog just display until the user closes it?

src/components/Dashboard/TodoList.vue
severinbeauvais: I think this looks great! Did you happen to check with UX folks if they like it too?

src/components/dialogs/GenericErrorDialog.vue
```diff
@@ -0,0 +1,59 @@
...
+    <v-card>
+      <v-card-title id="generic-error-dialog-title">
+        <slot name="errorTitle">
+          <v-icon
+            large
+            color="error"
+          >
+            mdi-alert-circle-outline
+          </v-icon>
+          <span class="ml-3">Error happened!</span>
```
severinbeauvais, line 17: Haha, but this is the default slot so why not? 😁

src/components/dialogs/GenericErrorDialog.vue
```diff
@@ -0,0 +1,59 @@
...
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Emit, Vue } from 'vue-property-decorator'
+import { ContactInfo } from '@/components/common'
+
+@Component({
+  components: { ContactInfo }
```
severinbeauvais, line 45: You don't use this so please remove it.

src/components/dialogs/GenericErrorDialog.vue
```diff
@@ -0,0 +1,59 @@
...
+  @Prop({ default: false }) readonly dialog!: boolean
+
+  /** Prop to provide attachment selector. */
+  @Prop({ default: '' }) readonly attach!: string
+
+  // Pass click event to parent.
+  @Emit() okay () {
+  }
+}
+</script>
```
severinbeauvais, line 58: I like this a lot 👍

src/components/dialogs/GenericErrorDialog.vue
```diff
@@ -4,7 +4,6 @@
     width="45rem"
     persistent
     :attach="attach"
-    content-class="delete-error-dialog"
```
severinbeauvais: Please add this back in as "generic-error-dialog" -- it can be used for unit testing.

src/components/Dashboard/TodoList.vue
hfekete: OK, apparently this is old design. I updated it to match new design. Also changed it a bit so that it should always conform to their style (e.g. can change the icon, but it is not a slot anymore, just icon mdi font, etc ...)

New design looks like this:
image

src/components/Dashboard/TodoList.vue
hfekete: I do this on line ~13 `@okay="fetchAffiliationInvitationsErrorDialog=false"`

There is no cleanup actions other then closing the dialog, so wanted to avoid adding unnecessary method. If need for method arises, it can be added at that time.

src/components/Dashboard/TodoList.vue
severinbeauvais: Looks good to me.

Kial, OK to resolve this conversation?

severinbeauvais:
LGTM.

Travis, Karim, any comments?

JazzarKarim (approved):
LGTM 👍

PR #536, opened 2023-07-27
16364 Changed store actions type

bennypc:
*Issue #:* /bcgov/entity#16364

*Description of changes:*
Changed action store type to match new requirements

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -84,7 +84,7 @@ import { Action, Getter } from 'pinia-class'
 import { AddCommentDialog, DownloadErrorDialog, FileCorrectionDialog, LoadCorrectionDialog }
   from '@/components/dialogs'
 import { CorrectionTypes } from '@/enums'
-import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF } from '@/interfaces'
+import { ApiFilingIF, CorrectionFilingIF } from '@/interfaces'
 import { FilingMixin } from '@/mixins'
 import { EnumUtilities, LegalServices } from '@/services/'
 import { navigate } from '@/utils'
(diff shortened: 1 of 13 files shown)


src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -84,7 +84,7 @@ import { Action, Getter } from 'pinia-class'
 import { AddCommentDialog, DownloadErrorDialog, FileCorrectionDialog, LoadCorrectionDialog }
   from '@/components/dialogs'
 import { CorrectionTypes } from '@/enums'
-import { ActionBindingIF, ApiFilingIF, CorrectionFilingIF } from '@/interfaces'
```
severinbeauvais, line 87: As this IF is no longer used (and the other one in that file is not used), please delete it.

severinbeauvais (approved):
Looks good. Remember to update the app version and commit the package files.

leodube-aot (approved):
LGTM, just the one comment from Sev about removing the unused interfaces. For my similar ticket I just had to remove ActionIF and ActionBindingIF from action-interface.ts.

JazzarKarim (approved):
LGTM 👍

src/components/Dashboard/FilingHistoryList.vue
bennypc: Which IF are you referring to?

ChrisGabel23:
When I did a similar ticket for Create-UI, there was a file with the action interfaces that was no longer needed. I don't see that file being deleted here: src/interfaces/action-interface.ts

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: The one you stopped using in this ticket.

src/interfaces/action-interface.ts
```diff
@@ -4,8 +4,3 @@ import { ActionContext } from 'vuex'
 export interface ActionIF {
```
ChrisGabel23, line 4: Is the "ActionIF" interface being used anywhere? This is the file @severinbeauvais was talking about deleting.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Please also remove the reference to this file from the interfaces/index.ts file. It's causing a failure of the build and unit tests checks.

Before committing, you should do these to check for issues:
```
npm run lint
npm run test:unit
```

src/interfaces/action-interface.ts
severinbeauvais: Please also remove the reference to this file from the interfaces/index.ts file. It's causing a failure of the build and unit tests checks.

Before committing, you should do these to check for issues:
```
npm run lint
npm run test:unit
```

bennypc: removed reference!

PR #540, opened 2023-08-22
17164 Vue CLI to Vite Upgrade

JazzarKarim:
*Issue #:* /bcgov/entity#17164

*Description of changes:*
- Migrating from Vue CLI to Vite
- Same as https://github.com/bcgov/entity/issues/17118 and https://github.com/bcgov/entity/issues/17165


package.json
```diff
@@ -24,6 +22,7 @@
     "@bcrs-shared-components/enums": "1.0.39",
     "@bcrs-shared-components/folio-number-input": "1.1.18",
     "@bcrs-shared-components/interfaces": "1.0.61",
+    "@bcrs-shared-components/mixins": "1.1.21",
```
JazzarKarim, line 25: I'm adding this package as I will be using the `CountriesProvincesMixin` from here now. It was previously a local mixin (I've deleted it from the project).

src/mixins/index.ts
```diff
@@ -11,11 +9,12 @@ import ResourceLookupMixin from './resource-lookup-mixin'
...
-  CountriesProvincesMixin,
-  DateMixin,
   DirectorMixin,
   EnumMixin,
   FilingMixin,
   NameRequestMixin,
   ResourceLookupMixin
 }
+
+export { default as DateMixin } from './date-mixin'
```
JazzarKarim, line 19: This mixin's way of import was cause problems:
filing-mixin error

It was a circular import. By importing it this way, it'll fix the problem.

src/mixins/index.ts
```diff
@@ -11,11 +9,12 @@ import ResourceLookupMixin from './resource-lookup-mixin'
...
+export { CountriesProvincesMixin } from '@bcrs-shared-components/mixins'
```
JazzarKarim, line 20: This is what I mentioned above in `package.json`. The mixin now is imported from the shared repo to fix the `require` problem.

tests/unit/StaffNotation.spec.ts
```diff
@@ -502,6 +504,8 @@ describe('StaffNotation', () => {
     // verify that build and create methods were called
     expect(buildRestorationFiling).toHaveBeenCalled()
     expect(createFiling).toHaveBeenCalled()
+    await Vue.nextTick()
+    await Vue.nextTick()
```
JazzarKarim, line 508: I was getting some weird behavior here. The test was returning the value of the previous one. The next test was returning the value of this test. This fixes that by waiting for the DOM to update.

Keep in mind I've tested this in local and it works fine!

tests/unit/TodoList2.spec.ts
```diff
@@ -67,11 +67,16 @@ describe('TodoList - common expansion panel header tests', () => {
 
     const sandbox = sinon.createSandbox()
     const fetchAffiliationInvites = sandbox.stub(AuthServices, 'fetchAffiliationInvitations')
-    // feature flag override
-    const featureFlag = sandbox.stub(utils, 'GetFeatureFlag')
 
-    featureFlag.withArgs('enable-affiliation-invitation-request-access')
-      .returns(true)
+    // feature flag override
+    vi.spyOn(utils, 'GetFeatureFlag').mockImplementation(
```
JazzarKarim, line 72: This test was failing:
failed test todolist

After spending some time on it, I found out the feature flag override wasn't working properly. Hence, the todo items list is always being 0. By mocking the feature flag like this, it'll fix the test.

tests/unit/StoreCurrentAccount.spec.ts
```diff
@@ -24,8 +25,8 @@ describe('testing current account module', () => {
     }
   }
 
-  it('fetches current account from session storage - PREMIUM', async () => {
-    jest.isolateModules(() => async () => {
+  it.skip('fetches current account from session storage - PREMIUM', async () => {
+    vi.isolateModules(() => async () => {
```
JazzarKarim, line 29: There's no function in Vitest that's the same as `isolateModules()`. I decided to skip the tests for now. Please let me know if that's OK.

postcss.config.js
JazzarKarim, line 1: No need for this file anymore.

tests/unit/StaffNotation.spec.ts
```diff
@@ -503,6 +505,10 @@ describe('StaffNotation', () => {
     expect(buildRestorationFiling).toHaveBeenCalled()
     expect(createFiling).toHaveBeenCalled()
 
+    // need to update DOM to not get previous test results
+    await Vue.nextTick()
+    await Vue.nextTick()
```
JazzarKarim, line 510: I was getting some weird behavior here. This test was returning the value of the previous one. The next test was returning the value of this test. This fixes that by waiting for the DOM to update. Ditto lines 563, 564, and 565.

Keep in mind I've tested this in local and it works fine!

index.html
```diff
@@ -6,10 +6,10 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <title>BC Business Registry</title>
     <link rel="icon" href="dummy-not-used-do-not-remove">
-    <link rel="icon" href="<%= BASE_URL %>favicon.png">
-    <link rel="preload" href="<%= BASE_URL %>fonts/materialdesignicons-webfont.927457ed.woff2" as="font" type="font/woff2" crossorigin>
-    <link rel="stylesheet" href="<%= BASE_URL %>css/addresscomplete-2.30.min.css">
-    <script type="text/javascript" src="<%= BASE_URL %>js/addresscomplete-2.30.min.js" defer></script>
+    <link rel="icon" href="/favicon.png">
+    <link rel="preload" href="/fonts/materialdesignicons-webfont.927457ed.woff2" as="font" type="font/woff2" crossorigin>
```
severinbeauvais, line 10: I wonder if we should preload some fonts in Edit UI and Create UI?

package.json
```diff
@@ -36,64 +35,53 @@
...
+    "vuelidate": "0.6.2",
     "vuetify": "^2.6.15",
     "vuex": "^3.6.2"
   },
   "devDependencies": {
-    "@babel/plugin-proposal-private-methods": "^7.18.6",
     "@pinia/testing": "^0.0.16",
-    "@types/jest": "^27.5.2",
     "@types/lodash": "^4.14.194",
-    "@types/pdfjs-dist": "^2.10.378",
```
severinbeauvais, line 61: You didn't see any use in this, or you couldn't find a version compatible with the package above?

package.json
```diff
@@ -36,64 +35,53 @@
...
-    "ts-jest": "^27.1.5",
     "typescript": "4.5.5",
+    "vite": "4.3.9",
+    "vite-plugin-environment": "^1.1.3",
+    "vitest": "0.33.0",
     "vue-class-component": "^7.2.6",
-    "vue-cli-plugin-vuetify": "^2.5.8",
     "vue-debounce-decorator": "^1.0.1",
     "vue-plugin-helper-decorator": "^0.0.12",
     "vue-property-decorator": "^9.1.2",
```
severinbeauvais, line 82: We don't need `vuelidate-property-decorators` here? Not even for the shared BaseAddress component?

package.json
JazzarKarim: No need for it as per latest package version:
no need for pdfjist types

src/mixins/filing-mixin.ts
```diff
@@ -1,5 +1,5 @@
 import { Component } from 'vue-property-decorator'
-import { DateMixin } from '@/mixins'
+import DateMixin from './date-mixin'
```
severinbeauvais, line 2: Why change this? Not that we would do it, but it makes THIS file less portable (say, to another folder).

In the old code, we could do a global search for "@/mixins" to see where they're used, but now we cannot.

src/views/StandaloneOfficeAddressFiling.vue
```diff
@@ -866,7 +866,7 @@ export default class StandaloneOfficeAddressFiling extends Mixins(CommonMixin, D
     // fetch original office addresses with new date + update working data
     // (this will overwrite the current data)
     this.isFetching = true
-    if (!this.isJestRunning) {
+    if (!this.isVitestRunning) {
```
severinbeauvais, line 869: Boy, some day we should just fix the unit tests so we can rid of this "test code" in Prod.

package.json
JazzarKarim: I agree but weirdly enough, it's working perfectly. I'm going to change that actually to stay consistent between all the UIs.

src/views/StandaloneOfficeAddressFiling.vue
JazzarKarim: Agreed.

index.html
JazzarKarim: As far as I know, I don't think so Sev?

tests/unit/StaffNotation.spec.ts
severinbeauvais: Yes, we've seen this elsewhere. There's a test function kicking around to do "await nextTick" and/or "await flushPromises" a few times for this purpose.

Why? Possibly some things that take more than 1 tick to complete, like validation maybe, or transitions.

(1 tick = 1 timeslice = 1 flow of code at a time = 1 message from event-driven architecture, which this is) Ref: [link](https://en.wikipedia.org/wiki/JavaScript#:~:text=event%2Ddriven%2C%20functional%2C%20and%20imperative%20programming%20styles)

tests/unit/StaffNotation.spec.ts
```diff
@@ -553,6 +559,11 @@ describe('StaffNotation', () => {
...
+    await Vue.nextTick()
```
severinbeauvais, line 565: Hahaha...

See https://github.com/bcgov/business-create-ui/blob/main/tests/wait-for-update.ts

src/mixins/filing-mixin.ts
JazzarKarim: Sev, as I mentioned in a comment above, many of the unit tests were failing "TypeError: Class extends value undefined is not a constructor or null"

This is a special case only in `filing-mixin.ts`. As I made some research online, I find out this type of import is what caused the issue.

tests/unit/StaffNotation.spec.ts
JazzarKarim: Ahhhh, I see! Never knew about this 🤣 

Sev, shall I create this file here too and make it the same as Create UI or keep as is?

vite.config.js
```diff
@@ -0,0 +1,60 @@
...
+      // simulate DOM with happy-dom
+      environment: 'happy-dom',
+      // enable jest-like global test APIs
+      globals: true,
+      setupFiles: ['./tests/setup.ts'],
+      // enable threads to speed up test running
+      threads: true,
+      // hide Vue Devtools message
+      onConsoleLog: function (log) {
+        if (log.includes('Download the Vue Devtools extension')) {
```
severinbeauvais, line 54: Weird that there's isn't a property override to hide this message, along the lines of this:
https://github.com/bcgov/business-filings-ui/blob/35248572ab7c39a5717ae7378c25ef7812ff89eb/src/main.ts#L24

package.json
JazzarKarim: I just noticed, I didn't update the BaseAddress component. It's still using the one from sbc-common-components. I'll fix that.

index.html
severinbeauvais: I remember playing with this (years ago), and this line preloads the MDI icons so the page can display then when it first loads -- probably for the initial spinner.

But we don't want to go crazy with preloading fonts as that delays the first render (ie, performance).

If Edit UI and Create UI don't show a spinner initially then they won't benefit from this. Otherwise, they might.

package.json
severinbeauvais: Excellent. Thanks!

src/mixins/filing-mixin.ts
severinbeauvais: So your change to mixins/index.ts didn't fix it?

tests/unit/StaffNotation.spec.ts
severinbeauvais: I'd copy the file over. Three "await nextTick" looks really hacky, but in a loop in an external function, no one will ever know! 😆

src/mixins/filing-mixin.ts
JazzarKarim: It did not. These two changes have to go together to fix the unit tests!

vite.config.js
JazzarKarim: Agreed.

tests/unit/StaffNotation.spec.ts
JazzarKarim: Thanks for pointing this out Sev! This is a pretty cool way to do it. Added the function in the latest commit.

src/mixins/filing-mixin.ts
severinbeauvais: OK. If it's a problem in the future then we'll fix it then.

vite.config.js
severinbeauvais: So there isn't a property  you can set?

tests/unit/StaffNotation.spec.ts
severinbeauvais: LGTM

index.html
severinbeauvais: Future problem. Don't worry about it now.

vite.config.js
JazzarKarim: I've made some research online. I can't find anything or anyone mentioning if there's a property override for this. I also took a look at the documentation, I couldn't find a property like the one you mentioned. I'm assuming this is the only way to hide the message here?

severinbeauvais: What about setting `Vue.config.devtools=false`? [ref](https://stackoverflow.com/a/54947933/8679162)

Or, fuggedaboutit.

JazzarKarim: Actually, I've found a super cool way to do it 😁 Please check the latest commit Sev!

JazzarKarim: It might be a good idea to make this change for both Edit and Create too 👍

package.json
```diff
@@ -1,21 +1,20 @@
...
-    "lint:nofix": "vue-cli-service lint --no-fix",
-    "test:e2e": "vue-cli-service test:e2e",
-    "test:unit": "vue-cli-service test:unit --testPathPattern --coverage",
-    "test:unit-silent": "vue-cli-service test:unit --silent",
-    "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
+    "dev": "vite",
+    "build": "vite build",
+    "serve": "vite preview",
+    "lint": "eslint . --ext .js,.ts,.vue",
+    "test:unit": "vitest run --silent",
```
severinbeauvais, line 12: This scares me a bit -- how silent are the tests now? Do we still see test errors and console output?

However, I think this might be a good option to add to the test:coverage script (since that command is concerned with coverage, not errors).

vite.config.js
severinbeauvais: Let's give this 5 more minutes of thought and then I vote to fuggedaboutit.

severinbeauvais:
LGTM but I don't want to be the only reviewer. Please poke some other devs :)

package.json
JazzarKarim: Console output:
console output initial
console output end

You're right. So, how about I revert the change here and add a new script:
`"test:coverage": "vitest run --silent"`
Thoughts?

severinbeauvais: Can you add the silent option to the existing test:coverage script?

Then, this horse is dead.

JazzarKarim: I forgot the script was already there haha. OK, sounds good. 🐴

package.json
```diff
@@ -9,8 +9,8 @@
     "build": "vite build",
     "serve": "vite preview",
     "lint": "eslint . --ext .js,.ts,.vue",
-    "test:unit": "vitest run --silent",
-    "test:coverage": "vitest run --coverage"
+    "test:unit": "vitest run",
+    "test:coverage": "vitest run --silent"
```
severinbeauvais, line 13: You need to leave the coverage option in there.

tests/setup.ts
```diff
@@ -0,0 +1,13 @@
...
+ * why some of the errors are showing up due to Vue not having the plugins it needs.
+ */
+import Vue from 'vue'
+import Vue2Filters from 'vue2-filters'
+import Vuelidate from 'vuelidate'
+import Vuetify from 'vuetify'
+
+Vue.use(Vue2Filters) // needed by SbcFeeSummary
+Vue.use(Vuelidate) // needed by BaseAddress
+Vue.use(Vuetify)
```
seeker25, line 13: What about router?
There's probably more that could be put in here possibly

package.json
severinbeauvais: That's the option for reporting how well the unit tests cover the different code paths.

Eg,
```
----------------------------------------------------------|----------|----------|----------|----------|-------------------|
File                                                      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------------------------------------------------|----------|----------|----------|----------|-------------------|
All files                                                 |    46.98 |    11.14 |     14.8 |     47.6 |                   |
 src                                                      |      100 |      100 |      100 |      100 |                   |
  App.vue                                                 |      100 |      100 |      100 |      100 |                   |
 src/components                                           |      100 |      100 |      100 |      100 |                   |
  tabs.vue                                                |      100 |      100 |      100 |      100 |                   |
```

tests/unit/BusinessActions.spec.ts
```diff
@@ -2,6 +2,7 @@ import LegalServices from '@/services/legal-services'
 import { createPinia, setActivePinia } from 'pinia'
 import { useBusinessStore, useRootStore } from '@/stores'
 import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
+import { vi } from 'vitest'
```
seeker25, line 5: I swear there's a way to do this without importing, just not sure how

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-540-gla4g369.web.app

tests/unit/BusinessActions.spec.ts
severinbeauvais: Use `vitest` instead?

package.json
```diff
@@ -36,64 +36,54 @@
...
-    "@types/jest": "^27.5.2",
     "@types/lodash": "^4.14.194",
-    "@types/pdfjs-dist": "^2.10.378",
     "@types/sinon": "^10.0.14",
     "@types/vue-the-mask": "^0.11.1",
     "@types/vuelidate": "^0.7.16",
     "@typescript-eslint/eslint-plugin": "^5.59.2",
     "@typescript-eslint/parser": "^5.59.2",
+    "@vitejs/plugin-vue2": "^2.2.0",
+    "@vitest/coverage-v8": "^0.34.2",
```
severinbeauvais, line 65: 👍 

Smells like we need another ticket to update Create and Edit UIs.

vite.config.js
severinbeauvais: Are you reverting the vitest config change?

tests/unit/BusinessActions.spec.ts
JazzarKarim: Please check latest commit. Apparently, the import was not needed whatsoever 😆

package.json
JazzarKarim: Awesome! 🙌

vite.config.js
JazzarKarim: I thought we'll hide it only with test coverage. I'll revert the change Sev, on it!

tests/setup.ts
JazzarKarim: True Travis but I'm trying to stay as consistent between UIs as possible. The other two don't have the router here! Let me know if I should make this change here and in other ones.

tests/wait-for-update.ts
```diff
@@ -0,0 +1,13 @@
+import Vue from 'vue'
+import flushPromises from 'flush-promises'
+
+/**
+ * Utility method to get around timing issues.
+ */
+export async function waitForUpdate (numTimesToFlushPromises = 1) {
```
seeker25, line 7: cool, probably will borrow this

tests/unit/BusinessActions.spec.ts
severinbeauvais: Haha, nice!

seeker25 (approved):
LGTM - address other's issues before merging

package.json
severinbeauvais: Are you creating the new ticket or am I? Or do you want to make the change as part of 17164?

vite.config.js
severinbeauvais: If the silent option hides unit test errors and console statements then we don't want that option when running the tests manually.

For the test coverage option, our goal is to see coverage info, not errors, so the silent option makes sense (assuming the coverage info is not silenced as well).

JazzarKarim: Got it. Sounds good Sev 👍

package.json
JazzarKarim: New tickets:
https://github.com/bcgov/entity/issues/17495 and https://github.com/bcgov/entity/issues/17497

I added the good first issue label to them also. Let me know if you disagree Sev.

ketaki-deodhar (approved):
👍 Looks good. Lots of learning similar to other UI update

pwei1018:
Temporary Url for review: https://business-filings-dev--pr-540-gla4g369.web.app

PR #543, opened 2023-09-21
17842 Fixed Pdf Ppload

JazzarKarim:
*Issue #:* /bcgov/entity#17842

*Description of changes:*
- Fixed PDF Upload when adding a court order
pdf works
psdf works 2
pdf works 3


src/components/common/FileUploadPdf.vue
```diff
@@ -205,15 +205,15 @@ export default class FileUploadPdf extends Vue {
     try {
       const arrayBuffer = await file.arrayBuffer()
       const data = new Uint8Array(arrayBuffer) // put it in a Uint8Array
-      const document = await this.pdfjsLib.getDocument({ data: data })
```
seeker25, line 208: Is it possible we could write some unit tests that can identify any issues? or copy over unit tests from other projects so this doesn't bust again somehow?

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-543-3qn65htz.web.app

src/components/common/FileUploadPdf.vue
JazzarKarim: I'll have a look Travis!

src/components/common/FileUploadPdf.vue
```diff
@@ -205,15 +205,15 @@ export default class FileUploadPdf extends Vue {
...
+      const document = await this.pdfjsLib.getDocument({ data }).promise
       const perms = await document.getPermissions()
       return { isEncrypted: false, isContentLocked: !!perms }
-    } catch (err) {
-      if (err.name === 'PasswordException') {
+    } catch (error) {
+      if (error.name === 'PasswordException') {
         return { isEncrypted: true, isContentLocked: true }
       }
-      throw err // re-throw any other error
```
severinbeauvais, line 215: If the error isn't "PasswordException", don't we want to re-throw it?

src/components/common/FileUploadPdf.vue
```diff
@@ -205,15 +205,15 @@ export default class FileUploadPdf extends Vue {
...
     }
+    return { isEncrypted: false, isContentLocked: false }
```
severinbeauvais, line 216: The only time this statement is executed is after the catch block. I think this is not a safe return value, Thoughts?

src/components/common/FileUploadPdf.vue
seeker25: Thank you, I think we use this in 3 projects now? it would be awesome if we had some unit tests we could spread across all the repos

src/components/common/FileUploadPdf.vue
JazzarKarim: Let's go with your PR Sev in which you only added the `.promise`. That should be good. I think that resolves your 2 comments.

severinbeauvais: OK.

src/components/common/FileUploadPdf.vue
severinbeauvais: See comment above.

src/components/common/FileUploadPdf.vue
JazzarKarim: Yes Travis. However, PDF upload is done differently in Filings UI than the other 2 UIs. The other 2 have a mixin and the actual component while here it's just one file that has everything.

Create UI has some tests there for PDF upload while Edit UI and this here doesn't have many. So, I totally agree with you.

JazzarKarim: Travis, I added some unit tests for the component (I was able to reuse some of the stuff in Create UI). Please see latest commit!

severinbeauvais (approved):

seeker25 (approved):
LGTM

PR #546, opened 2023-09-29
17216 - add start date while building dissolution filing

ketaki-deodhar:
*Issue #:* /bcgov/entity#17216

*Description of changes:*

Add start date while building dissolution filing. This is a cleaner solution and start date can be saved to and fetched from `draftFiling` when filing a Dissolution in Create UI

### DIFF
--- a/src/stores/businessStore.ts
+++ b/src/stores/businessStore.ts
@@ -1,4 +1,4 @@
-import { AllowedActionsIF, ApiBusinessIF, BusinessStateIF, BusinessWarningIF } from '@/interfaces'
+import { AllowedActionsIF, ApiBusinessIF, ApiDateTimeUtc, BusinessStateIF, BusinessWarningIF } from '@/interfaces'
 import { defineStore } from 'pinia'
 import { CorpTypeCd, EntityState } from '@/enums'
 import { DateUtilities, LegalServices } from '@/services/'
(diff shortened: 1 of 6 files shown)


src/stores/businessStore.ts
```diff
@@ -218,6 +224,10 @@ export const useBusinessStore = defineStore('business', {
       this.businessInfo.foundingDate = val
     },
 
+    setStartDate (val: string) {
+      this.businessInfo.startDate = val
```
severinbeauvais, line 228: For clarity, type of val should be `ApiDateTimeUtc`.

Also, don't you need to call this in App.vue when the business data is loaded?

src/stores/businessStore.ts
```diff
@@ -57,6 +58,11 @@ export const useBusinessStore = defineStore('business', {
       return DateUtilities.apiToDate(state.businessInfo.foundingDate)
     },
 
+    /** The start date. */
+    getStartDate (state: BusinessStateIF): Date {
+      return DateUtilities.apiToDate(state.businessInfo.startDate)
```
severinbeauvais, line 63: It might be simpler to NOT convert this value in multiple places. What do you think?

src/stores/businessStore.ts
ketaki-deodhar: For first,  I will change it
For second, no do not need to call it. This will be set in store when `businessInfo` is set (`fetchBusiness` api call)

src/stores/businessStore.ts
ketaki-deodhar: Please see my recent commit. Would something like that work better?

severinbeauvais (approved):

PR #548, opened 2023-10-12
15600 display correct name for SP/GP

jamespaologarcia:
*Issue #:* /bcgov/entity#15600

Display correct name for SP/GP. Show Operating Name instead of Entity Name for SP/GP if it exists. 


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-548-mkpdafap.web.app

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "6.9.0",
+  "version": "6.9.1",
```
severinbeauvais, line 3: This is a significant change (new Legal Names API) -- change version to 6.10.0.

Also, at the top of the PR, change it to "commmit into" the Legal Name feature branch.

src/interfaces/business-interfaces.ts
```diff
@@ -46,6 +56,7 @@ export interface ApiBusinessIF {
   lastLedgerTimestamp?: ApiDateTimeUtc // not used
   lastModified?: ApiDateTimeUtc // not used
   legalName: string
+  operatingName: string
```
severinbeauvais, line 59: This property is not part of the business object, as far as I know. Did you find an example of where it is? (Then I'd fix my code.) Otherwise please remove it.

src/stores/businessStore.ts
```diff
@@ -273,21 +283,26 @@ export const useBusinessStore = defineStore('business', {
      * @param context the Vuex context (passed in automatically)
      */
     loadBusinessInfo (): Promise<any> {
-      // need to return a promise because action is called via dispatch
+      // Need to return a promise because the action is called via dispatch.
```
severinbeauvais, line 286: Our code style has been both types of comments: phrases (not capitalized, no period), and full sentences (usually multi-line). So this was OK before.

severinbeauvais (commented):
1. ~You need to commit `.env.example` with the new Legal API URL, same as in https://github.com/bcgov/business-create-ui/pull/575.~ _done_
2. ~You need to update `devops/.vaults` with the new Legal Name API URL mapping (for fetching the URL from the 1Password db), same as in PR above.~ _done_
3. ~You need to update `firebase.json` to allow the app to connect to the new Legal API URL, same as in PR above.~ _done_
4. ~You need to fix the broken unit test(s).~ _done_

leodube-aot (approved):
LGTM!

src/components/EntityInfo/EntityHeader.vue
```diff
@@ -102,6 +102,7 @@ export default class EntityHeader extends Vue {
   @Prop({ required: true }) readonly tempRegNumber!: string // may be null
 
   @Getter(useBusinessStore) getEntityName!: string
+  @Getter(useBusinessStore) getLegalName!: string
```
severinbeauvais, line 105: This isn't used in this file. Please remove it.

src/stores/businessStore.ts
```diff
@@ -277,17 +282,24 @@ export const useBusinessStore = defineStore('business', {
...
+            // Set data to store
             this.setBusinessInfo(businessInfo)
-            // return the business info object
+            if (this.isFirm) {
+              // Extract the operatingName from the businessInfo
+              const alternateName = businessInfo.alternateNames?.find(
+                x => x.identifier === businessInfo.identifier
+              )?.operatingName
+              // Set the operatingName if it exists
+              if (alternateName) this.setLegalName(alternateName)
```
severinbeauvais, line 300: This overwrites the actual legal name. I think a better solution is to save the operating name in a new store property, and return it when code asks for legal name (if this is a firm).

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-548-mkpdafap.web.app

src/interfaces/business-interfaces.ts
```diff
@@ -66,7 +66,8 @@ export interface ApiBusinessIF {
   startDate: ApiDateTimeUtc
   submitter?: string // not used
   taxId?: string // aka Business Number // may be absent
-  warnings: Array<BusinessWarningIF>
+  warnings: Array<BusinessWarningIF>,
+  operatingName?: string
```
severinbeauvais, line 70: Looks good but:
- interface fields don't need commas separating them
- please keep this list in alphabetical order (to make it easier for the next dev to find things)

src/stores/businessStore.ts
```diff
@@ -28,7 +28,8 @@ export const useBusinessStore = defineStore('business', {
       state: null,
       stateFiling: null,
       startDate: null,
-      warnings: []
+      warnings: [],
+      operatingName: null
```
severinbeauvais, line 32: Alphabetize, please.

src/stores/businessStore.ts
```diff
@@ -186,6 +191,11 @@ export const useBusinessStore = defineStore('business', {
       return (this.getBusinessState === EntityState.HISTORICAL)
     },
 
+    /** Is True if entity is a Sole Proprietorship or General Partnership. */
+    isTypeFirm (): boolean {
+      return (this.isTypeSoleProp || this.isTypePartnership)
```
severinbeauvais, line 196: Are these 2 getters declared anywhere in this app? I can't find them.

severinbeauvais (commented):
Please add a unit tests that verifies the operating name is correctly displayed for a firm (SP or GP) when the operating name is different from the legal name.

Example: https://github.com/bcgov/business-create-ui/blob/54373b3f09ba482bda1ac793f85f5343d7d596ff/tests/unit/EntityInfo.spec.ts#L139

src/stores/businessStore.ts
jamespaologarcia: Removing this one.

Just found out there's already a similar one:

image

severinbeauvais: This may also be why the unit tests failed (with getLegalName returning undefined).

PS You can (and should) run both linting and the unit tests locally:
```
npm run lint
npm run test:unit
```

severinbeauvais:
Please change the target branch to `feature-legal-name`.

image

tests/unit/App.spec.ts
```diff
@@ -1970,6 +1970,7 @@ describe('App as a COMPLETED Registration Application', () => {
   it('fetches NR data properly', () => {
     expect(rootStore.getNameRequest.nrNum).toBe('NR 1234567')
     expect(businessStore.getLegalName).toBe('My Name Request')
+    expect(businessStore.getOperatingName).toBe('My Name Request')
```
severinbeauvais: This check is good as it verifies that operating name was populated correctly when this SP registration is loaded, but it would be even better if the operating name was different from the legal name.

Testing this may be a bit tricky. Basically, I think you'll have to update the (operating) name in the NR above (lines 1895+).

I believe this is where the legal name is set for a draft app: https://github.com/bcgov/business-filings-ui/blob/3479b284afd2232bcab6243ce408d83d714967fb/src/App.vue#L654

In this special case, it appears we don't have a separate legal name (in the business object). But this is a special case so I'm OK with that. (We plan to update this special case later anyway.)

src/interfaces/business-interfaces.ts
```diff
@@ -51,6 +61,7 @@ export interface ApiBusinessIF {
   naicsDescription: string // firm only
   naicsKey: string // firm only
   nextAnnualReport: ApiDateTimeUtc // BCOMP only
+  operatingName?: string
```
severinbeauvais, line 64: I'm thinking you don't need both the alternate names array and the operating name.

I think it's better to store the alternate names and compute the operating name from it (in business store, line 93). Basically, the code you have on line 298 would get moved to line 93.

src/stores/businessStore.ts
```diff
@@ -83,9 +83,15 @@ export const useBusinessStore = defineStore('business', {
       return state.businessInfo.lastDirectorChangeDate
     },
 
-    /** The legal name. */
+    /** The legal name, or operating name if is firm. */
     getLegalName (state: BusinessStateIF): string {
-      return state.businessInfo.legalName
+      return this.isFirm && this.getOperatingName ? this.getOperatingName : state.businessInfo.legalName
```
severinbeauvais, line 88: Please put `this.isFirm && this.getOperatingName` in parentheses. Ternaries are nice except when they look more complicated than if-then-else 😉

src/stores/businessStore.ts
```diff
@@ -83,9 +83,15 @@ export const useBusinessStore = defineStore('business', {
...
+    },
+
+    getOperatingName (state: BusinessStateIF): string {
+      const { alternateNames, identifier } = state.businessInfo
+      const operatingName = alternateNames?.find((x) => x.identifier === identifier)?.operatingName
+      return operatingName || null
```
severinbeauvais, line 94: Looks good.

Nice use of spreading 👍

src/stores/businessStore.ts
```diff
@@ -277,17 +283,16 @@ export const useBusinessStore = defineStore('business', {
...
         const businessId = sessionStorage.getItem('BUSINESS_ID')
 
-        // if there is no business id, return error
+        // If there is no business id, return an error.
         if (!businessId) {
           reject(new Error('Missing business id'))
           return
         }
-
         LegalServices.fetchBusiness(businessId)
```
severinbeauvais, line 291: The whitespace above was to separate the error checking from the rest of the code. Please restore it.

src/interfaces/business-interfaces.ts
```diff
@@ -26,6 +35,7 @@ export interface BusinessWarningIF {
 // FUTURE: verify/indicate unused properties
 export interface ApiBusinessIF {
   adminFreeze: boolean
+  alternateNames?: Array<AlternateNameIF>
   allowedActions: AllowedActionsIF
```
severinbeauvais, line 39: "all" comes before "alt" :)

severinbeauvais (commented):
I'd still like to see some unit tests to verify that the operating name is displayed instead of the legal name for firms, please. (Including one for the IA/registration.)

Everything else (except new comments) looks fine.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-548-mkpdafap.web.app

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-548-mkpdafap.web.app

tests/unit/EntityHeader.spec.ts
```diff
@@ -271,3 +271,55 @@ describe('Entity Header - AUTHORIZED TO CONTINUE OUT badge', () => {
...
+describe('Entity Header - Operating Name', () => {
+  const router = mockRouter.mock()
+  it('displays operating name if firm', async () => {
+    // set store properties
+    businessStore.setBusinessInfo(
+      {
+        legalName: 'My Business',
+        alternateNames: [
+          {
+            operatingName: 'Wayne Enterprises'
```
severinbeauvais, line 284: I'm confused why you don't need an identifier in here, but the test seems to be verifying what I would expect and it seems to pass, so 🤷‍♂️

tests/unit/EntityHeader.spec.ts
```diff
@@ -271,3 +271,55 @@ describe('Entity Header - AUTHORIZED TO CONTINUE OUT badge', () => {
...
+        ]
+      } as any
+    )
+    businessStore.setGoodStanding(true)
+    businessStore.setLegalType(CorpTypeCd.BENEFIT_COMPANY)
+    // mount the component and wait for everything to stabilize
+    const wrapper = shallowMount(EntityHeader, {
+      vuetify,
+      router,
+      propsData: { businessId: 'BEN1052377', tempRegNumber: null }
```
severinbeauvais, line 319: Small thing but this isn't a valid business id -- should be "BC...".

severinbeauvais (approved):
A couple of small questions/comments but otherwise looks good to merge.

PR #555, opened 2023-10-24
AGM Location Change Third PR

JazzarKarim:
*Issue #:* /bcgov/entity#18161

*Description of changes:*
- Created unit tests for AGM Location
- Created unit tests for AGM year
- Fixed EntityMenu test from part 1 ticket
- Misc. fixes

**Note: I'm kind of blocked right on since I can't file the filing yet. I can't proceed on this before https://github.com/bcgov/business-schemas/pull/146 (schema) and https://github.com/bcgov/lear/pull/2258 (filer) are merged.**

### DIFF
--- a/src/components/AgmLocationChange/AgmYear.vue
+++ b/src/components/AgmLocationChange/AgmYear.vue
@@ -15,6 +15,7 @@
 
 <script lang="ts">
 import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
+import { Debounce } from 'vue-debounce-decorator'
 
 @Component({})
 export default class AgmYear extends Vue {
@@ -39,6 +40,12 @@ export default class AgmYear extends Vue {
     this.emitValid(this.value)
   }
 
+  @Watch('value')
+  @Debounce(300)
+  onValueChanged (val: string): void {
+    this.emitValid(val)
+  }
+
   /** Validate AGM year field */
   @Watch('validateForm')
   validateAgmYear (): void {
@@ -50,9 +57,8 @@ export default class AgmYear extends Vue {
 
   /** Emits an event with the changed input (ie, updated v-model). */
   @Emit('input')
-  emitInput (val: string): void {
-    this.emitValid(val)
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  emitInput (val: string): void {}
 
   /** Emits an event indicating whether or not this component is valid. */
   @Emit('valid')
--- a/src/components/AgmLocationChange/AgmYear.vue
+++ b/src/components/AgmLocationChange/AgmYear.vue
@@ -15,7 +15,6 @@
 
 <script lang="ts">
 import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
-import { Debounce } from 'vue-debounce-decorator'
 
 @Component({})
 export default class AgmYear extends Vue {
@@ -40,12 +39,6 @@ export default class AgmYear extends Vue {
     this.emitValid(this.value)
   }
 
-  @Watch('value')
-  @Debounce(300)
-  onValueChanged (val: string): void {
-    this.emitValid(val)
-  }
-
   /** Validate AGM year field */
   @Watch('validateForm')
   validateAgmYear (): void {
@@ -57,8 +50,9 @@ export default class AgmYear extends Vue {
 
   /** Emits an event with the changed input (ie, updated v-model). */
   @Emit('input')
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  emitInput (val: string): void {}
+  emitInput (val: string): void {
+    this.emitValid(val)
+  }
 
   /** Emits an event indicating whether or not this component is valid. */
   @Emit('valid')
--- a/tests/unit/AgmLocation.spec.ts
+++ b/tests/unit/AgmLocation.spec.ts
@@ -89,13 +89,11 @@ describe('AgmLocation', () => {
       })
 
     const input = wrapper.find('#agm-location')
-    let testString = ''
-    for (let i = 0; i <= 100; i++) {
-      testString += 'a'
-    }
-    await input.setValue(testString)
+    let a = 'a'
+    let a101 = a.repeat(101)
+    await input.setValue(a101)
 
-    expect(wrapper.emitted('update:agmLocation').pop()[0]).toBe(testString)
+    expect(wrapper.emitted('update:agmLocation').pop()[0]).toBe(a101)
     expect(wrapper.emitted('valid').pop()[0]).toBe(false)
     wrapper.destroy()
   })
--- a/tests/unit/AgmLocation.spec.ts
+++ b/tests/unit/AgmLocation.spec.ts
@@ -89,8 +89,8 @@ describe('AgmLocation', () => {
       })
 
     const input = wrapper.find('#agm-location')
-    let a = 'a'
-    let a101 = a.repeat(101)
+    const a = 'a'
+    const a101 = a.repeat(101)
     await input.setValue(a101)
 
     expect(wrapper.emitted('update:agmLocation').pop()[0]).toBe(a101)
--- a/tests/unit/AgmLocation.spec.ts
+++ b/tests/unit/AgmLocation.spec.ts
@@ -1,0 +1,102 @@
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+import { mount } from '@vue/test-utils'
+import AgmLocation from '@/components/AgmLocationChange/AgmLocation.vue'
+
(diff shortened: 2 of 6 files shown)


src/components/AgmLocationChange/AgmYear.vue
```diff
@@ -40,12 +39,6 @@ export default class AgmYear extends Vue {
     this.emitValid(this.value)
   }
 
-  @Watch('value')
-  @Debounce(300)
-  onValueChanged (val: string): void {
```
JazzarKarim, line 45: Not needed. This way we'll also stay consistent between all the other filings.

severinbeauvais: Does this component emit the validity flag on every character typed? The purpose of the debounce was to reduce the "noise" coming from this component.

tests/unit/AgmLocation.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        vuetify,
+        propsData: {
+          rules: rules
+        }
+      })
+    const vm: any = wrapper.vm
+
+    expect(vm.agmLocation).toBe('')
+    expect(wrapper.emitted('update:agmLocation')).toBeUndefined()
+    expect(wrapper.emitted('valid').pop()[0]).toBe(false)
```
severinbeauvais, line 30: How many "valid" events were emitted, here and below?

tests/unit/AgmLocation.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        propsData: {
+          rules: rules
+        }
+      })
+
+    const input = wrapper.find('#agm-location')
+    let testString = ''
+    for (let i = 0; i <= 100; i++) {
+      testString += 'a'
+    }
```
severinbeauvais, line 95: https://www.w3schools.com/jsref/jsref_repeat.asp

tests/unit/EntityMenu.spec.ts
```diff
@@ -503,6 +503,9 @@ describe.skip('Entity Menu - Request AGM Extension click tests', () => {
     vm = wrapper.vm
     vi.spyOn(vm, 'enableAgmExtension', 'get').mockReturnValue(true)
 
+    // Fix TypeError: GetFeatureFlag(...).includes is not a function test error (unrelated to extensions)
+    vi.spyOn(vm, 'enableAgmLocationChg', 'get').mockReturnValue(true)
```
severinbeauvais, line 507: Oh, so this is because the mock for GetFeatureFlag (lines 486-487) returns a boolean instead of an array.

I'd prefer to see a smarter mock above than this "unrelated" work-around, if possible.

src/components/AgmLocationChange/AgmYear.vue
JazzarKarim: You're absolutely right. Returned the debounce + fixed tests accordingly.

tests/unit/AgmLocation.spec.ts
JazzarKarim: What do you mean Sev? It's only one event which is an array of false and an empty string.
valid events
The first value is the emitted validity.

tests/unit/AgmLocation.spec.ts
JazzarKarim: This is pretty neat! There's always new stuff to learn, thanks Sev!

tests/unit/EntityMenu.spec.ts
JazzarKarim: The issue is inEntityMenu.vue on line 260-261:
```javascript
  get enableAgmLocationChg (): boolean {
    return !!GetFeatureFlag('supported-agm-location-chg-entities').includes(this.getLegalType)
  }
```

The tests are actually passing in here. The compiler is showing however an error that we cannot use includes on something that hasn't been set. Keep in mind this is for location change which is why it's unrelated since this test is for extensions.

The current code (latest commit) mocks the return value of the getter above to be `BEN`.

Please let me know if you that's not OK Sev.

tests/unit/AgmLocation.spec.ts
```diff
@@ -89,13 +89,11 @@ describe('AgmLocation', () => {
       })
 
     const input = wrapper.find('#agm-location')
-    let testString = ''
-    for (let i = 0; i <= 100; i++) {
-      testString += 'a'
-    }
-    await input.setValue(testString)
+    let a = 'a'
+    let a101 = a.repeat(101)
```
severinbeauvais: You can even do `'a'.repeat(101)` 😃

tests/unit/AgmLocation.spec.ts
severinbeauvais: In this test, "valid" is only emitted once. But in other tests it might be emitted 100 times...

tests/unit/EntityMenu.spec.ts
```diff
@@ -499,8 +499,9 @@ describe.skip('Entity Menu - Request AGM Extension click tests', () => {
       mixins: [{ methods: { isAllowed: () => true } }],
       propsData: { businessId: 'BC1234567' }
     })
-
     vm = wrapper.vm
+
+    vi.spyOn(vm, 'enableAgmLocationChg', 'get').mockReturnValue(CorpTypeCd.BENEFIT_COMPANY)
```
severinbeauvais, line 504: OK, this is less of a hack than the other version.

I was thinking that lines 486-487 could look like this:
```
vi.spyOn(utils, 'GetFeatureFlag').mockImplementation(flag => {
 if (flag === 'enable-agm-extension') return True
 if (flag === 'supported-agm-location-chg-entities') return 'BEN'
})
```

This might not work or might not fix the root cause. But I think the error you were seeing is because GetFeatureFlag was returning a boolean, and then some code was trying to use it like an array (eg, true.include(...)).

tests/unit/EntityMenu.spec.ts
severinbeauvais: See my other comment below. Resolving this one.

tests/unit/AgmLocation.spec.ts
JazzarKarim: I just checked Sev, the others emit two. One for the onCreated and the other is what I'm trying to do (which is to be expected). For example:
array test

tests/unit/EntityMenu.spec.ts
JazzarKarim: Ohhh I see. This worked Sev! Makes a lot of sense, thank you so much.

tests/unit/AgmLocation.spec.ts
severinbeauvais: I bet if you entered each "a" like a user typing, you'd have a lot more valid events (esp without the debouncing).

This is fine. Just remember to have a look at the number of events emitted in scenarios like there. "Popping" the last one might give you the answer the test expects, but there may a lot events before that (probably unnecessarily).

tests/unit/EntityMenu.spec.ts
severinbeauvais: Small correction of what I wrote:

> some code was trying to use it like an array

or *string**.

tests/unit/AgmLocation.spec.ts
```diff
@@ -89,8 +89,7 @@ describe('AgmLocation', () => {
       })
 
     const input = wrapper.find('#agm-location')
-    const a = 'a'
-    const a101 = a.repeat(101)
+    const a101 = 'a'.repeat(101)
     await input.setValue(a101)
```
severinbeauvais, line 93: Don't change it this time, but in the future you may want to simulate a user typing in a bunch of "a"s so that it exercises the debouncing/validation code as realistically as possible.

tests/unit/EntityMenu.spec.ts
```diff
@@ -499,9 +501,8 @@ describe('Entity Menu - Request AGM Extension click tests', () => {
       mixins: [{ methods: { isAllowed: () => true } }],
       propsData: { businessId: 'BC1234567' }
     })
-    vm = wrapper.vm
 
-    vi.spyOn(vm, 'enableAgmLocationChg', 'get').mockReturnValue(CorpTypeCd.BENEFIT_COMPANY)
+    vm = wrapper.vm
     vi.spyOn(vm, 'enableAgmExtension', 'get').mockReturnValue(true)
```
severinbeauvais, line 506: You probably don't need this since the default flag (line 487) is True. So maybe the default flags should be False and "".

tests/unit/AgmLocation.spec.ts
JazzarKarim: Got it, I'll keep that mind.

leodube-aot (approved):

ketaki-deodhar (approved):
Looks good!

PR #556, opened 2023-10-25
AGM Location Change 4th and final PR

JazzarKarim:
*Issue #:* /bcgov/entity#18161

*Description of changes:*
- Removed the option to cancel payment in case AGM Location change payment was unsuccessful
- Misc. fixes

~Note: The work for this ticket is mostly done. However, I won't merge until filer and schema are merged and verifying that everything looks good. I believe things should be fine in the filings history list.~ _Can merge now since I can file and it looks good._

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -672,7 +672,6 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
   // for template
   readonly EnumUtilities = EnumUtilities
   readonly FilingStatus = FilingStatus
-  readonly FilingTypes = FilingTypes
 
   /** Whether a COA is pending. */
   get isCoaPending (): boolean {
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -825,7 +825,7 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
     })
   }
 
-  /** Check if todo item is cancellable (has a draft). */
+  /** Check if task item is cancellable (has a draft). */
   isCancellableTodoItem(item: TodoItemIF): boolean {
     return (item.name === FilingTypes.AGM_LOCATION_CHANGE)
   }
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -403,7 +403,7 @@
                   </v-btn>
                   <v-btn
                     v-else
-                    :class="{ 'btn-resume-payment': !isAgmLocationChange(item) }"
+                    :class="{ 'btn-resume-payment': !isCancellableTodoItem(item) }"
                     color="primary"
                     :disabled="!item.enabled"
                     @click.native.stop="doResumePayment(item)"
@@ -413,7 +413,7 @@
 
                   <!-- dropdown menu -->
                   <v-menu
-                    v-if="!isAgmLocationChange(item)"
+                    v-if="!isCancellableTodoItem(item)"
                     offset-y
                     left
                   >
@@ -825,8 +825,8 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
     })
   }
 
-  /** Check if todo item is an AGM Location change. */
-  isAgmLocationChange (item: TodoItemIF): boolean {
+  /** Check if todo item is cancellable (has a draft). */
+  isCancellableTodoItem(item: TodoItemIF): boolean {
     return (item.name === FilingTypes.AGM_LOCATION_CHANGE)
   }
 
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -826,7 +826,7 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
   }
 
   /** Check if task item is cancellable (has a draft). */
-  isCancellableTodoItem(item: TodoItemIF): boolean {
+  isCancellableTodoItem (item: TodoItemIF): boolean {
     return (item.name === FilingTypes.AGM_LOCATION_CHANGE)
   }
 
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -403,7 +403,7 @@
                   </v-btn>
                   <v-btn
                     v-else
-                    class="btn-resume-payment"
+                    :class="{ 'btn-resume-payment': !isAgmLocationChange(item) }"
                     color="primary"
                     :disabled="!item.enabled"
                     @click.native.stop="doResumePayment(item)"
@@ -413,6 +413,7 @@
 
                   <!-- dropdown menu -->
                   <v-menu
+                    v-if="!isAgmLocationChange(item)"
                     offset-y
                     left
                   >
@@ -671,6 +672,7 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
   // for template
   readonly EnumUtilities = EnumUtilities
   readonly FilingStatus = FilingStatus
+  readonly FilingTypes = FilingTypes
 
   /** Whether a COA is pending. */
   get isCoaPending (): boolean {
@@ -821,6 +823,11 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin, E
         this.checkIfCompleted(id, count)
       }, 1000)
     })
+  }
+
+  /** Check if todo item is an AGM Location change. */
+  isAgmLocationChange (item: TodoItemIF): boolean {
+    return (item.name === FilingTypes.AGM_LOCATION_CHANGE)
   }
 
   /** Loads a todo item into the Todo Items array. */
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -394,10 +394,8 @@
                 <template v-else-if="isStatusPending(item)">
                   <v-btn
                     v-if="EnumUtilities.isPayMethodOnlineBanking(item)"
-                    :class="
-                      [
-                        !isCancellableTodoItem(item) ? 'btn-change-payment-type' : 'btn-change-payment-type-cancellable'
-                      ]"
+                    class="btn-change-payment-type"
+                    :class="{ 'cancellable' : isCancellableTodoItem(item) }"
                     color="primary"
                     :disabled="!item.enabled"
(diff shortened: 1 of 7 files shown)


src/components/Dashboard/TodoList.vue
```diff
@@ -413,6 +413,7 @@
 
                   <!-- dropdown menu -->
                   <v-menu
+                    v-if="!isCancellableTodoItem(item)"
```
JazzarKarim, line 416: I created dummy data for two todo tasks (annual report and AGM location change) with both the payment being pending.
different types of payment pending

As you can see, only the AGM location will not have the cancel payment option since it does not have a salvageable draft. I also went for a general name here since we might have more of those in the future. We can just plug in the extra types in the function below (we're going to probably add it for Extensions 😁 ).

src/components/Dashboard/TodoList.vue
```diff
@@ -403,7 +403,7 @@
                   </v-btn>
                   <v-btn
                     v-else
-                    class="btn-resume-payment"
+                    :class="{ 'btn-resume-payment': !isCancellableTodoItem(item) }"
```
severinbeauvais, line 406: Also on line 397?

src/components/Dashboard/TodoList.vue
severinbeauvais: Two thumbs up! 👍 👍

src/components/Dashboard/TodoList.vue
severinbeauvais, line 406: I wasn't going to pick on this, but I think I will anyway (sorry).

Instead of removing the classname if this item isn't cancellable, why not add a classname if it isn't cancellable, and adjust the styles below accordingly? (as in, "adjust the border if the selector does not include cancellable")

That way, this button always has a classname.

severinbeauvais: Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/:not

JazzarKarim: No worries!

Sev, I'm a bit confused here though. My thought was that having no class looks good since we don't want to set the borders to 0 if it's not cancellable. So basically: 

If not cancellable => no setting of the border (no class).
If cancellable => set border to 0 (the class is there).

So having a class when it isn't cancellable will not have any properties since I believe it looks good as is.

Please let me know if I'm missing something so that I can fix it!

JazzarKarim: I don't believe it's needed. That class is for a different kind of button ("Change Payment Type") that's irrelevant to uncancellable filings.

severinbeauvais: Hey man, you always want to be classy 😉 

It's a button. It should have some way to identify or select it in the DOM (eg, for testing). Even if the class isn't used to add CSS properties.

JazzarKarim: Well, I am your student 😎 

OK, got it. I'll have a class for both cases then Sev!

severinbeauvais: You're sure a failed payment would want to display that other button? I think you get one button vs the other depending on the pay method that was used. But I don't know what a pay error would look like...

severinbeauvais: Just a second, dynamic class, in addition to the existing classname, right?

severinbeauvais: AFAIK, the only pay error we would see (since this is a zero fee filing) is an error creating or linking the payment (with Pay API). I don't know if there are other pay errors for "not enough funds" or whatever. Needs to be looked into.

JazzarKarim: You're right. I'll be adding the case to that line.

However, I'm looking at the different buttons and conditions we have (it all depends on FilingStatus), the only time we have a cancel payment is in this case. The other cases (FilingStatus.ERROR for example), we get the Retry Payment option which is fine I believe.

severinbeauvais: Oh, I'm getting confused.

Do you know, on a pay error that we might see here, whether the filing is in Pending or Error status? There are several possible buttons (Change Payment Type, Resume Payment, Cancel Payment, Retry Payment).

JazzarKarim: Sev, I went ahead and checked the possible buttons for every single `FilingStatus` that we have. The ones that have buttons which are applicable here in this case are:
- `FilingStatus.PENDING`:
pending

- `FilingStatus.ERROR`:
error

Now, which one of the above we will see if the filing doesn't go through, I don't know yet. If the FilingStatus was `ERROR`, we would not usually see the other button. It's only when the FilingStatus is `PENDING` that we see the other chevron button to open the small menu.

severinbeauvais: Curious: what does View Details show for the each case?

I'm thinking we'd get the error status if the payment can't be processed, and Retry should work. I don't know if we'd ever see the first one, but if your code handles both cases then we're safe and that fine by me.

JazzarKarim: For PENDING:
for pending details

For ERROR:
for error details

Yup, I believe the retry should work (it happened with me a couple of days back in another filing and that worked). I guess I can fully test it when filer is merged.

severinbeauvais: Thanks.

I'm _guessing_ that the payment record could be created and been interrupted even for this zero-fee filing, and Resume should set things back in motion.

I'm _guessing_ that an unsuccessful payment (creation?) is more likely, and Retry should restart it.

I think we should assume that both cases are possible and handle them accordingly. Can this conversation be resolved now?

JazzarKarim: Yes, exactly. I don't see a situation/scenario in which this filing might get stuck. Currently all scenarios are handled.

Yes Sev! Unless you say otherwise, I believe we should be good.

JazzarKarim: Sev, sorry for taking too long to reply. My head, for some reason, wasn't processing this although it was super simple haha.

Please have a look at last commit, hopefully it's fine now!

JazzarKarim:
**Update: Filer ticket has been merged. Everything looks fine!**
filer working

src/components/Dashboard/TodoList.vue
```diff
@@ -394,7 +394,10 @@
                 <template v-else-if="isStatusPending(item)">
                   <v-btn
                     v-if="EnumUtilities.isPayMethodOnlineBanking(item)"
-                    class="btn-change-payment-type"
+                    :class="
+                      [
+                        !isCancellableTodoItem(item) ? 'btn-change-payment-type' : 'btn-change-payment-type-cancellable'
+                      ]"
```
severinbeauvais, line 400: This is fine, but I was thinking of this:

in template:
```
  class="btn-change-payment-type"
  :class="{ 'cancellable' : isCancellableTodoItem(item) }"
```

in CSS:
```
  .btn-change-payment-type.cancellable,
  .btn-resume-payment.cancellable {
    border-top-right-radius: 0;
    border-bottom-right-radius: 0;
  }
```

severinbeauvais (approved):

src/components/Dashboard/TodoList.vue
```diff
@@ -416,7 +415,7 @@
 
                   <!-- dropdown menu -->
                   <v-menu
-                    v-if="!isCancellableTodoItem(item)"
+                    v-if="isCancellableTodoItem(item)"
```
severinbeauvais, line 418: Did you mean to change this?

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-556-2djj8nd1.web.app

severinbeauvais:
@JazzarKarim Check the red validation bars. Should they be the full height of the section? Also watch their top/bottom radii.

ketaki-deodhar (approved):
LGTM!

severinbeauvais:
@JazzarKarim I noticed the SaveErrorDialog (and properties/methods) have been removed. Aren't they needed in case of a pay error, and with the ability to retry?

src/components/Dashboard/TodoList.vue
JazzarKarim: Yes Sev!

Although the previous code worked, I was a bit weird to read and understand because of negations. Also, the function's name didn't actually resemble what it was doing.

I just made the change to simplify the logic and make it easier to understand.

JazzarKarim:
> @JazzarKarim Check the red validation bars. Should they be the full height of the section? Also watch their top/bottom radii.

You're right Sev. Pretty sure the UX folks won't like that. I'll see if I can fix it.

severinbeauvais:
Can  you have a quick look at this please?

image

severinbeauvais:
Hey, I'm unable to file an AGM Location Change from the temp URL, possibly because agmLocation is empty?

image

src/views/AgmLocationChg.vue
```diff
@@ -14,15 +14,6 @@
...
     />
 
-    <StaffPaymentDialog
-      :staffPaymentData.sync="staffPaymentData"
-      attach="#agm-location-chg"
-      :dialog="staffPaymentDialog"
-      :loading="filingPaying"
-      @exit="staffPaymentDialog=false"
-      @submit="onClickFilePay(true)"
-    />
```
severinbeauvais, line 24: Can staff still file an AGM Location Change?

severinbeauvais:
> Sev, your question made me notice something important. For other no fee filings (like continuation out), we don't show the StaffPaymentDialog. So, I connected with Yui and asked him if this was also the case here. He said yes.
> 
> So, I went ahead and removed the dialog and everything associated with it.

Great catch! I've added the same todo item to the AGM Extension ticket.

(I hope we don't suddenly add a fee to this filing though.)

JazzarKarim:
> Can you have a quick look at this please?

What's wrong here Sev?

JazzarKarim:
> Hey, I'm unable to file an AGM Location Change from the temp URL, possibly because agmLocation is empty?
> 

It's actually because the main branch in LEAR is not in DEV currently. However, you're still right. I fixed that in my latest commit.

severinbeauvais:
> > Can you have a quick look at this please?
> 
> What's wrong here Sev?

The hover circle isn't centered on the checkbox square.

src/views/AgmLocationChg.vue
JazzarKarim: Yes Sev! I just tested that too.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-556-2djj8nd1.web.app

PR #559, opened 2023-10-30
AGM Location Change 5th PR for styling issues

JazzarKarim:
*Issue #:* bcgov/entity#18161 + bcgov/entity#16764

*Description of changes:*
Fixed what was leftover from previous PR (https://github.com/bcgov/business-filings-ui/pull/556):
  - Fixed validation red error borders
  - Fixed certify button ripple

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -393,7 +393,7 @@ export default {
         'setRecordsAddress',
         'setRegisteredAddress',
         'setTasks',
-        'setUserFullName',
+        'setUserInfo',
         'setUserKeycloakGuid'
       ]),
 
@@ -438,6 +438,7 @@ export default {
       // and save user's Keycloak GUID
       try {
         const userInfo = await AuthServices.fetchUserInfo(this.getAuthApiUrl).then(response => response?.data)
+        this.setUserInfo(userInfo)
         await this.updateLaunchDarkly(userInfo)
         this.setUserKeycloakGuid(userInfo.keycloakGuid)
       } catch (error) {
@@ -574,21 +575,16 @@ export default {
       }
     },
 
-    /**
-     * Updates Launch Darkly with current user info.
-     * Set the full name in the Authentication store.
-     */
+    /** Updates Launch Darkly with current user info. */
     async updateLaunchDarkly (userInfo: any): Promise<void> {
       // since username is unique, use it as the user key
       const key: string = userInfo.username
       const email: string = userInfo.contacts[0]?.email || userInfo.email
       const firstName: string = userInfo?.firstname
       const lastName: string = userInfo?.lastname
-      const fullName: string = firstName + ' ' + lastName
       // store Keycloak roles in custom object
       const custom = { roles: this.getKeycloakRoles } as any
 
-      this.setUserFullName(fullName)
       await UpdateLdUser(key, email, firstName, lastName, custom)
     },
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -393,6 +393,7 @@ export default {
         'setRecordsAddress',
         'setRegisteredAddress',
         'setTasks',
+        'setUserFullName',
         'setUserKeycloakGuid'
       ]),
 
@@ -573,16 +574,21 @@ export default {
       }
     },
 
-    /** Updates Launch Darkly with current user info. */
+    /**
+     * Updates Launch Darkly with current user info.
+     * Set the full name in the Authentication store.
+     */
     async updateLaunchDarkly (userInfo: any): Promise<void> {
       // since username is unique, use it as the user key
       const key: string = userInfo.username
       const email: string = userInfo.contacts[0]?.email || userInfo.email
       const firstName: string = userInfo?.firstname
       const lastName: string = userInfo?.lastname
+      const fullName: string = firstName + ' ' + lastName
       // store Keycloak roles in custom object
       const custom = { roles: this.getKeycloakRoles } as any
 
+      this.setUserFullName(fullName)
       await UpdateLdUser(key, email, firstName, lastName, custom)
     },
 
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -174,4 +174,14 @@ export default class Certify extends Vue {
     margin-top: -1.25rem;
   }
 }
+
+:deep() {
+  .v-input--selection-controls__ripple {
+    border-radius: 0%;
+    height: 19px;
+    width: 19px;
+    left: -1px;
+    margin: 4px;
+  }
+}
 </style>
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -20,6 +20,7 @@
           filled
           persistent-hint
           label="Person's legal name"
+          :disabled="disableEdit"
           :value="certifiedBy"
           :rules="[ v => !!v || 'A person\'s legal name is required.']"
           @input="emitCertifiedBy($event)"
@@ -77,6 +78,9 @@ export default class Certify extends Vue {
 
   /** Prompt the validations. Used for global validations. */
   @Prop({ default: false }) readonly validateForm!: boolean
+
+    /** Disable Text Input field. */
+    @Prop({ default: false }) readonly disableEdit!: boolean
 
   certifyName = ''
 
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -177,11 +177,7 @@ export default class Certify extends Vue {
 
 :deep() {
   .v-input--selection-controls__ripple {
-    border-radius: 0%;
-    height: 19px;
-    width: 19px;
-    left: -1.5px;
-    margin: 4px;
+    margin-top: -3px;
   }
 }
 </style>
--- a/src/components/common/Certify.vue
+++ b/src/components/common/Certify.vue
@@ -180,7 +180,7 @@ export default class Certify extends Vue {
     border-radius: 0%;
     height: 19px;
     width: 19px;
-    left: -1px;
+    left: -1.5px;
     margin: 4px;
   }
 }
--- a/src/interfaces/current-user-interface.ts
+++ b/src/interfaces/current-user-interface.ts
@@ -1,11 +1,10 @@
 /** The Current User (aka: Keycloak) interface. */
 export interface CurrentUserIF {
   email: string
-  firstName: string
-  fullName: string
+  firstname: string
   keycloakGuid: string
-  lastName: string
+  lastname: string
   loginSource: string
   roles: Array<string>
-  userName: string
+  type: string
 }
--- a/src/interfaces/current-user-interface.ts
+++ b/src/interfaces/current-user-interface.ts
@@ -1,10 +1,11 @@
 /** The Current User (aka: Keycloak) interface. */
 export interface CurrentUserIF {
   email: string
-  firstname: string
+  firstName: string
+  fullName: string
   keycloakGuid: string
-  lastname: string
+  lastName: string
   loginSource: string
   roles: Array<string>
-  type: string
+  userName: string
 }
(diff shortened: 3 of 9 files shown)


src/components/common/Certify.vue
```diff
@@ -174,4 +174,14 @@ export default class Certify extends Vue {
     margin-top: -1.25rem;
   }
 }
+
+:deep() {
+  .v-input--selection-controls__ripple {
```
JazzarKarim, line 179: Had to configure this manually from scratch.
Before:
Screenshot (112)

After:
Screenshot (111)

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

JazzarKarim:
For the Validation red border errors;
Before:
before validations

After:
after validation errors

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

src/views/AgmLocationChg.vue
```diff
@@ -90,18 +90,15 @@
               </header>
 
               <div>
-                <v-card
-                  flat
-                  class="py-4"
-                >
+                <v-card flat>
```
severinbeauvais, line 93: Screenshot, please.

severinbeauvais: Oops, nvm.

severinbeauvais:
> For the Validation red border errors; Before:
> 
> After: 

I'm thinking you should remove the top-left and bottom-right border radius for the red border. That's what our UI/UX guys wanted in previous filings.

src/components/common/Certify.vue
severinbeauvais: Your change makes this checkbox behave differently from other checkboxes. Why not do something like this?

image

JazzarKarim: Yup, I thought the circle thing was wrong. It's just the location of the ripple was wrong. I was thinking of something else.

After latest commit:
Screenshot (113)

JazzarKarim:
> > For the Validation red border errors; Before:
> > After:
> 
> I'm thinking you should remove the top-left and bottom-right border radius for the red border. That's what our UI/UX guys wanted in previous filings.

Great catch Sev! After latest commit:
borders radii

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

src/components/common/Certify.vue
severinbeauvais: I'm ok to resolve this conversation.

severinbeauvais:
Please check for console errors when  you first land on the page and click File Now.

severinbeauvais:
> Please check for console errors when you first land on the page and click File Now.

Please test and see what you get.

JazzarKarim:
> > Please check for console errors when you first land on the page and click File Now.
> 
> Please test locally and see what you get.

Is it the "[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "error"" errors?

If I'm not mistaken, that has been the case for a while. I can try and fix all of them in this PR. Let me know Sev.

severinbeauvais:
Do not merge until Business Digital Credentials is out of Dev environment.

JazzarKarim:
If non-staff:
if non staff certify

If staff:
if staff

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

src/views/AgmLocationChg.vue
```diff
@@ -454,6 +455,12 @@ export default class AgmLocationChg extends Mixins(CommonMixin, DateMixin,
     // see https://v3.vuejs.org/api/options-lifecycle-hooks.html#mounted
     await this.$nextTick()
 
+    // Pre-populate the certified block with the logged in user's name (if not staff)
+    if (!this.isRoleStaff) {
+      const userInfo = await AuthServices.fetchUserInfo(this.getAuthApiUrl).then(response => response?.data)
+      this.certifiedBy = userInfo?.firstname + ' ' + userInfo?.lastname
```
severinbeauvais, line 461: Are you sure we didn't already load this data in App.vue?

severinbeauvais:
> On load, If non-staff: (snip)
> 
> If staff: (snip)

In both cases, the name is editable, right?

src/views/AgmLocationChg.vue
JazzarKarim: We use this same function in App.vue but nothing is being set in the store. I had two choices, either set the logged in user's name in the store or add this here. I went with this route. Let me know if I should go with the first one Sev (or if you think something else is better).

JazzarKarim:
> > On load, If non-staff: (snip)
> > If staff: (snip)
> 
> In both cases, the name is editable, right?

Yes!

src/views/AgmLocationChg.vue
severinbeauvais: It's better for an API call to be made as few times as possible.

Also, keeping it here could slow down the page's loading by up to 30 seconds (or whatever the network timeout is), and I was hoping we could get rid of the loading spinner and associated plumbing.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

src/App.vue
```diff
@@ -573,16 +574,21 @@ export default {
...
       // since username is unique, use it as the user key
       const key: string = userInfo.username
       const email: string = userInfo.contacts[0]?.email || userInfo.email
       const firstName: string = userInfo?.firstname
       const lastName: string = userInfo?.lastname
+      const fullName: string = firstName + ' ' + lastName
       // store Keycloak roles in custom object
       const custom = { roles: this.getKeycloakRoles } as any
 
+      this.setUserFullName(fullName)
```
severinbeauvais, line 591: Don't do this here. It's a "side effect" -- not the purpose of this function; no one would look here to see where user full name is set.

Also, I think you should save the entire user info object into store -- we've already done the work of fetching it; there might be other info in there we could use later.

src/stores/authenticationStore.ts
```diff
@@ -20,6 +20,11 @@ export const useAuthenticationStore = defineStore('authentication', {
       return useRootStore().userKeycloakGuid
     },
 
+    /** The user's Full Name. */
+    getUserFullName (): string {
+      return useRootStore().userFullName
+    },
```
severinbeauvais, line 26: Either move this to rootStore.ts or move the property (and setter) to this store.

My vote is to move this to the root store. This store uses `RootStateIF` (unfortunately named) which contains user account and auth data. Meanwhile the root store (which uses `StateIF`) contains more "info" type data.

JazzarKarim: Sev, I moved it to the rootStore and did the changes accordingly.

JazzarKarim:
Last commit, I'm doing: https://github.com/bcgov/entity/issues/16764

Non-staff:
new certify consent

Staff:
new certify consent staff

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

src/interfaces/current-user-interface.ts
```diff
@@ -1,11 +1,10 @@
...
-  fullName: string
+  firstname: string
   keycloakGuid: string
-  lastName: string
+  lastname: string
   loginSource: string
   roles: Array<string>
-  userName: string
+  type: string
 }
```
severinbeauvais, line 10: This interface is used by the authentication store, which gets its data from the KC token (not from the auth "users/@me" call).

Sample data:
image

Don't use THIS interface it you want to type the userInfo object.

src/views/ConsentContinuationOut.vue
```diff
@@ -193,6 +193,7 @@
                   :isCertified.sync="isCertified"
                   :certifiedBy.sync="certifiedBy"
                   :class="{ 'invalid-certify': !certifyFormValid && showErrors }"
+                  :disableEdit="!isRoleStaff"
```
severinbeauvais, line 196: I thought it was supposed to be editable in all cases?

src/views/ConsentContinuationOut.vue
```diff
@@ -503,6 +505,13 @@ export default class ConsentContinuationOut extends Mixins(CommonMixin, DateMixi
 
     this.dataLoaded = true
 
+    // Pre-populate the certified block with the logged in user's name (if not staff)
+    if (!this.isRoleStaff) {
+      const firstName = this.getUserInfo?.firstname
+      const lastName = this.getUserInfo?.lastname
+      this.certifiedBy = firstName + ' ' + lastName
```
severinbeauvais, line 512: If getUserInfo is null, this would show `undefined undefined`.

severinbeauvais (changes requested):
~Sorry, you broke an interface.~ _fixed now_

src/views/ConsentContinuationOut.vue
JazzarKarim: The ticket says for end-user and follow SP/GP registration's way. I checked there and it was only editable if staff.

src/App.vue
severinbeauvais: I'm resolving this conversation because the initial issue is fixed.

However, you have created another issue, but it's captured in another conversation so it's still OK to resolve this one.

src/views/ConsentContinuationOut.vue
severinbeauvais: Please check with Mihai. I think there's something underway to allow it to be editable for regular users too.

src/views/ConsentContinuationOut.vue
severinbeauvais: I like this fix 👍

severinbeauvais: Oh, please fix the other place you do this :)

src/interfaces/state-interface.ts
```diff
@@ -19,6 +19,7 @@ export interface StateIF {
   entityStatus: EntityStatus // for draft app only
   keycloakRoles: Array<string>
   stateFiling: StateFilingIF // the state filing object
+  userInfo: any
```
severinbeauvais, line 22: Yeah, I did the same thing. Looks good.

src/views/ConsentContinuationOut.vue
JazzarKarim: I sent Mihai a message Sev. Waiting for his answer.

JazzarKarim: Sev, Mihai confirmed that it's OK (https://github.com/bcgov/entity/issues/16764#issuecomment-1791493214).

severinbeauvais: Yup. And it's editable for the AGM filings, right?

severinbeauvais (approved):
LGTM.

Get yourself some other approvals and let's merge this puppy.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-559-ku35gn15.web.app

src/views/ConsentContinuationOut.vue
JazzarKarim: For AGM Filings, it's editable in all cases. However, I'm going to double check with Mihai just in case something has changed.

tests/unit/AgmLocationChg.spec.ts
```diff
@@ -144,7 +145,7 @@ describe('AGM Location Chg view', () => {
 
     const wrapper = shallowMount(AgmLocationChg, { mocks: { $route, $router } })
     wrapper.vm.$data.dataLoaded = true
-    await Vue.nextTick()
+    await waitForUpdate(2)
```
ketaki-deodhar, line 148: This looks something new :)

ketaki-deodhar (approved):
LGTM!

tests/unit/AgmLocationChg.spec.ts
JazzarKarim: This function existed before but we rarely use it! It's found in the tests for PDF Upload for example when we need to wait for DOM updates.

The other way to fix it is have 2 `await Vue.nextTick()`. This function takes place instead of having these duplicate lines.

src/views/ConsentContinuationOut.vue
JazzarKarim: Going to merge now. Mihai said there'll be another ticket in case something changes. I think the requirement for that is still in the works by him. If something needs to change, we'll fix it there in the another ticket or I'll just open another PR.

PR #565, opened 2023-11-02
17869 - Adding CFS account number BUSINESS-FILINGS-UI

AbrahamRostampoor:
*Issue #:* [/bcgov/entity#17869](https://github.com/bcgov/entity/issues/17869)

*Description of changes:*
- Added CFS Account Number to PaymentPendingOnlineBanking.vue


severinbeauvais:
Abraham, I assigned this PR to you (which allows the checks, including gcbrun, to run).

/gcbrun

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-565-wuxtflvp.web.app

src/components/Dashboard/TodoList/PaymentPendingOnlineBanking.vue
severinbeauvais, line 1: This component was designed to be simple and dumb. As such, the only data it had was what was passed in to it (via props).

Now, this component also accesses the store and session storage, which reduces the "purity" of the component.

Ideally, you would pass in just the CFS account id.

seeker25: I think we were trying to avoid calling PAY-API unless the component was mounted or displayed (avoid overloading PAY-API) and the call only exists for ONLINE BANKING, which is why I thought it would make sense to put it inside of here.

seeker25: I guess the alternative would be to put the store, session storage, service call in TodoList and pass it down

severinbeauvais: I agree with you.

Since the parent already has access to the account id and pay URL, could you pass those as props and remove the store getter and session storage call?

severinbeauvais:
And please fix the broken unit test. Thanks!

src/components/Dashboard/TodoList/PaymentPendingOnlineBanking.vue
```diff
@@ -30,16 +36,25 @@
...
+
+  cfsAccountId: string | null = null;
 
   /** The draft title of the subject filing. */
   get draftTitle (): string {
     return this.filing?.draftTitle || 'filing'
   }
+
+  async mounted () {
+    const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id
```
seeker25, line 55: This already exists outside and should be passed in as a prop:

image

src/services/pay-services.ts
```diff
@@ -18,4 +18,21 @@ export default class PayServices {
...
+   * Fetches the CFS account ID from the pay-api.
+   * @param payApiUrl the URL of the pay-api
+   * @param orgId the organization ID for which to fetch the CFS account ID
+   * @returns the CFS account ID
+   */
+  static async fetchCfsAccountId (payApiUrl: string, orgId: number): Promise<string> {
+    const url = `${payApiUrl}accounts/${orgId}`
+    try {
+      const response = await axios.get(url)
+      return response?.data.cfsAccount.cfsAccountNumber
```
seeker25, line 32: How does this handle if you manually override the payment method to non CFS?

`response?.data?.cfsAccount?.cfsAccountNumber` <- I'd recommend using incase someone is forcefully switched (unlikely)

src/components/Dashboard/TodoList/PaymentPendingOnlineBanking.vue
```diff
@@ -30,16 +36,25 @@
...
 import { Component, Prop, Vue } from 'vue-property-decorator'
+import { PayServices } from '@/services/'
 
 @Component({})
 export default class PaymentPendingOnlineBanking extends Vue {
   /** The subject filing. */
   @Prop({ required: true }) readonly filing!: any
+  @Prop({ required: true }) readonly getPayApiUrl!: string
+
+  cfsAccountId: string | null = null;
```
seeker25, line 47: `cfsAccountId: string = null;`

src/components/Dashboard/TodoList/PaymentPendingOnlineBanking.vue
```diff
@@ -30,16 +36,25 @@
 
 <script lang="ts">
 import { Component, Prop, Vue } from 'vue-property-decorator'
+import { PayServices } from '@/services/'
 
 @Component({})
 export default class PaymentPendingOnlineBanking extends Vue {
   /** The subject filing. */
   @Prop({ required: true }) readonly filing!: any
+  @Prop({ required: true }) readonly getPayApiUrl!: string
```
seeker25, line 45: payApiUrl

src/components/Dashboard/TodoList/PaymentPendingOnlineBanking.vue
```diff
@@ -30,16 +36,25 @@
...
+
+  cfsAccountId: string = null;
 
   /** The draft title of the subject filing. */
   get draftTitle (): string {
     return this.filing?.draftTitle || 'filing'
   }
+
+  async mounted () {
+    this.cfsAccountId = await PayServices.fetchCfsAccountId(this.getPayApiUrl, this.accountId)
```
seeker25, line 56: `this.payApiUrl`

src/services/pay-services.ts
```diff
@@ -18,4 +18,21 @@ export default class PayServices {
...
       .catch(() => {}) // ignore errors
   }
+
+  /**
+   * Fetches the CFS account ID from the pay-api.
+   * @param payApiUrl the URL of the pay-api
+   * @param orgId the organization ID for which to fetch the CFS account ID
+   * @returns the CFS account ID
+   */
+  static async fetchCfsAccountId (payApiUrl: string, orgId: number): Promise<string> {
```
seeker25, line 28: would rename `orgId` to `accountId`, so it's consistent with above

src/services/pay-services.ts
```diff
@@ -18,4 +18,21 @@ export default class PayServices {
       .then(response => response?.data)
       .catch(() => {}) // ignore errors
   }
+
+  /**
+   * Fetches the CFS account ID from the pay-api.
+   * @param payApiUrl the URL of the pay-api
+   * @param orgId the organization ID for which to fetch the CFS account ID
```
seeker25, line 25: would rename orgId to accountId, so it's consistent with above

seeker25 (approved):
LGTM, can't merge for a bit I believe because of code freeze

seeker25:
@AbrahamRostampoor could you update the package and package-lock, and we'll merge?

PR #568, opened 2023-11-03
18024 Incremental AGM Extension changes

severinbeauvais:
*Issue #:* bcgov/entity#18024

*Description of changes:*
- app version = 7.0.3
- removed obsolete dialogs + code
- stopped watching changes to certify data
- cleaned up indentation
- added initial message box + CSS
- added NotEligibleExtension dialog
- removed waiveFees=true from AGM Extension
- removed waiveFees=true from AGM Location Chg
- added some unit test suites
- used allowable actions value from Legal API
- show alert on error starting app

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

PR #570, opened 2023-11-07
18272 Add unit test for AboutTheBusiness component

leodube-aot:
*Issue #:* /bcgov/entity#18272

*Description of changes:*
- Added unit tests for AboutTheBusiness component

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "7.0.3",
+  "version": "7.0.4",
   "private": true,
   "appName": "Filings UI",
   "sbcName": "SBC Common Components",
--- a/tests/unit/AboutTheBusiness.spec.ts
+++ b/tests/unit/AboutTheBusiness.spec.ts
@@ -1,35 +1,57 @@
 import Vue from 'vue'
 import Vuetify from 'vuetify'
-import { mount } from '@vue/test-utils'
+import VueRouter from 'vue-router'
+import { shallowMount } from '@vue/test-utils'
+import { createPinia, setActivePinia } from 'pinia'
+import { useBusinessStore } from '@/stores'
 import AboutTheBusiness from '@/components/AgmExtension/AboutTheBusiness.vue'
+import mockRouter from './mockRouter'
+import { AgmExtEvalIF } from '@/interfaces'
+import { DateUtilities } from '@/services'
 
 Vue.use(Vuetify)
-const vuetify = new Vuetify({})
+Vue.use(VueRouter)
 
-describe('AboutTheBusiness', () => {
-  it('displays normally', () => {
-    const wrapper = mount(AboutTheBusiness, {
+const vuetify = new Vuetify({})
+setActivePinia(createPinia())
+const businessStore = useBusinessStore()
+
+describe('About The Business component', () => {
+  const router = mockRouter.mock()
+
+  it('handles empty data', async () => {
+    const wrapper = shallowMount(AboutTheBusiness, {
+      vuetify,
+      router,
+      propsData: { data: null }
+    })
+    await Vue.nextTick()
+
+    expect(wrapper.find('#entity-in-good-standing').exists()).toBe(false)
+    expect(wrapper.find('#entity-date-of-incorporation').exists()).toBe(false)
+  })
+
+  it('displays business info properly', async () => {
+    // set store properties
+    businessStore.setGoodStanding(true)
+    businessStore.setFoundingDate('1971-05-12T00:00:00-00:00')
+
+    // mount the component and wait for everything to stabilize
+    const wrapper = shallowMount(AboutTheBusiness, {
+      vuetify,
+      router,
       propsData: {
         data: {
-          isGoodStanding: true,
-          incorporationDate: new Date('2023-12-31T08:00:00.000Z')
-        }
-      },
-      vuetify
+          isGoodStanding: businessStore.isGoodStanding,
+          incorporationDate: businessStore.getFoundingDate
+        } as unknown as AgmExtEvalIF
+      }
     })
+    await Vue.nextTick()
 
-    expect(wrapper.find('.v-card').attributes('id')).toBe('about-the-business')
-    expect(wrapper.find('header i').attributes('class')).toContain('mdi-domain')
-    expect(wrapper.find('header h2').text()).toBe('About the Business')
-
-    const rows = wrapper.findAll('.content > .row')
-
-    expect(rows.at(0).find('.col-sm-3').text()).toBe('Business in Good Standing')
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "7.0.2",
+  "version": "7.0.3",
```
severinbeauvais, line 3: Please run `npm i` and commit the package lock file also.

tests/unit/AboutTheBusiness.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  it('handles empty data', async () => {
+    const wrapper = shallowMount(AboutTheBusiness, {
+      vuetify,
+      router,
+      propsData: { data: null }
+    })
+    await Vue.nextTick()
+
+    expect(wrapper.find('#entity-in-good-standing').exists()).toBe(false)
+    expect(wrapper.find('#entity-date-of-incorporation').exists()).toBe(false)
```
severinbeauvais, line 31: Why would the elements not exist?

tests/unit/AboutTheBusiness.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  it('displays business info properly', async () => {
+    // set store properties
+    businessStore.setGoodStanding(true)
+    businessStore.setFoundingDate('1971-05-12T00:00:00-00:00')
```
severinbeauvais, line 37: You shouldn't need the store anywhere in the tests, because the component itself doesn't use it.

tests/unit/AboutTheBusiness.spec.ts
```diff
@@ -0,0 +1,57 @@
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+import VueRouter from 'vue-router'
+import { shallowMount } from '@vue/test-utils'
```
severinbeauvais, line 4: Whenever possible, in unit tests, I think it's better to deep-mount the component under test.

Pro: Vuetify sub-components render and function correctly (they are not stubbed).

Con: All other sub-components are rendered and this may complicate the testing. Remember, we want to test THIS component, not everything under it (which should have their own tests). If you need a mix of Vuetify but not other sub-components, you can stub them out like this:

image

tests/unit/AboutTheBusiness.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+setActivePinia(createPinia())
+const businessStore = useBusinessStore()
+
+describe('About The Business component', () => {
+  const router = mockRouter.mock()
+
+  it('handles empty data', async () => {
+    const wrapper = shallowMount(AboutTheBusiness, {
+      vuetify,
+      router,
```
severinbeauvais, line 25: I think you don't need the router. Try without it.

tests/unit/AboutTheBusiness.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    // mount the component and wait for everything to stabilize
+    const wrapper = shallowMount(AboutTheBusiness, {
+      vuetify,
+      router,
+      propsData: {
+        data: {
+          isGoodStanding: businessStore.isGoodStanding,
+          incorporationDate: businessStore.getFoundingDate
+        } as unknown as AgmExtEvalIF
```
severinbeauvais, line 47: This is fine, but another possibility, to avoid incomplete objects, might be:
```
data: {
  ...EmptyAgmExtEval,
  isGoodStanding: businessStore.isGoodStanding,
  incorporationDate: businessStore.getFoundingDate
} as AgmExtEvalIF
```

tests/unit/AboutTheBusiness.spec.ts
severinbeauvais, line 1: I should merge [my PR](#568) first as I did a bit of work here (which you can reject since yours is more complete).

leodube-aot: If your PR is close to merge I don't mind if you merge first while I work through the comments on this one. I'll finish up my review of your PR right now.

tests/unit/AboutTheBusiness.spec.ts
leodube-aot: Nice, I definitely prefer this approach much more casting as unknown first.

tests/unit/AboutTheBusiness.spec.ts
leodube-aot: I think what happens is that since the `data` prop is required, passing `null` to it does not create the component. [EntityHeader.spec.ts](https://github.com/bcgov/business-filings-ui/blob/ba9afd9d7ce0d4a1339c4f15f099da2be8baf41d/tests/unit/EntityHeader.spec.ts#L24) has a similar test in it, but perhaps a better way of doing this is to check that the component elements do not exist (like `v-card`).

severinbeauvais: Ah, I see.

You probably don't need to test that the prop is not provided, as that will result in a compile error. I suggest testing with an empty data object (either `EmptyAgmExtEval` or `{}`).

severinbeauvais: Fixed and looks good. Thanks!

JazzarKarim (approved):
LGTM Leo! 👍

tests/unit/AboutTheBusiness.spec.ts
```diff
@@ -2,33 +2,52 @@ import Vue from 'vue'
...
-    expect(rows.at(0).find('.col-sm-9').text()).toBe('Yes')
-
     expect(rows.at(1).find('.col-sm-3').text()).toBe('Date of Incorporation')
-    expect(rows.at(1).find('.col-sm-9').text()).toBe('December 31, 2023')
+
+    // verify displayed text
+    expect(wrapper.find('#entity-in-good-standing').text()).toBe('Yes')
+    expect(wrapper.find('#entity-date-of-incorporation').text())
+      .toBe(DateUtilities.dateToPacificDate(
+        new Date('2023-12-31T08:00:00.000Z'), true
```
severinbeauvais, line 49: Can you check for an actual long string date here (eg, "January 1, 2024")?

severinbeauvais:
Please get other approvals before I approve. Thanks!

tests/unit/AboutTheBusiness.spec.ts
severinbeauvais: Yes, you got the date right.

This is better because your are testing the output of `dateToPacificDate()`. Previously, any error would have cancelled out.

PaulGarewal (approved):
LGTM 👍 Reviewing the Unit Testing was really helpful  🙌

severinbeauvais (approved):

PR #571, opened 2023-11-07
18024 Incremental AGM Extension changes

severinbeauvais:
*Issue #:* bcgov/entity#18024

*Description of changes:*
- app version = 7.0.6
- updated ExtensionRequest validation styling
- misc small fixes
- added isEligible to interface
- show dialog if not eligible
- implemented unit tests

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

PR #573, opened 2023-11-08
18369 agm extension evaluation

PaulGarewal:
*Issue #:* /bcgov/entity#18369

*Description of changes:*

- added logic to the AGMextension evaluation for eligible and not eligible
-  added unit testing


src/views/AgmExtension.vue
```diff
@@ -63,6 +63,10 @@
             <AgmExtensionEvaluation
               class="mt-8"
               :data.sync="data"
+              :isEligible="false"
```
severinbeauvais, line 66: You don't need to pass this in -- it's already part of the data object on line 65.

src/views/AgmExtension.vue
```diff
@@ -63,6 +63,10 @@
...
+              :agmYearText="''"
```
severinbeauvais, line 67: You don't use this in the sub-component.

src/components/AgmExtension/AgmExtensionEvaluation.vue
```diff
@@ -88,5 +106,13 @@ import { AgmExtEvalIF } from '@/interfaces'
 })
 export default class AgmExtensionEvaluation extends Vue {
   @Prop({ required: true }) readonly data!: AgmExtEvalIF
+
+  @Prop({ default: '' }) readonly extensionDurationText!: string
+
+  @Prop({ default: '' }) readonly dueDateText!: string
```
severinbeauvais, line 112: I was rather hoping that THIS component could figure these strings instead of having them passed in.

It's the idea of separation of concerns: don't involve extra components in computations if they don't need to know. In this case, the parent component is simply a framework for the data, not the source of these particular props. Karim's component does all the computations. And THIS component can react to Karim's properties and display the proper strings.

Another way to look at it is trying to make self-contained components with as-simple-as-possible interfaces to the rest of their world (props/events).

src/interfaces/agm-ext-eval-interface.ts
```diff
@@ -1,5 +1,6 @@
 /** Object that stores AGM Extension evaluation fields. */
 export interface AgmExtEvalIF {
+  isEligible: boolean
```
severinbeauvais, line 3: When I merge my PR #571 , you'll see I added this to the end of the props list. Doesn't matter either way.

I also initialized it to False instead of null below, meaning no-one has to update it for it to be correct initially. This doesn't really matter since it will probably be set initially by Karim's component logic.

src/interfaces/agm-ext-eval-interface.ts
```diff
@@ -14,10 +15,11 @@ export interface AgmExtEvalIF {
 
 /** An empty AGM Extension Evaluation object. Note: don't assign this - make a copy instead. */
 export const EmptyAgmExtEval: AgmExtEvalIF = {
+  isEligible: null,
   isGoodStanding: null,
   incorporationDate: null,
   isFirstAgm: null,
   agmYear: null,
-  extensionDuration: NaN,
```
severinbeauvais, line 21: Both NaN and null as valid "non-numeric" values for a number, so your change is fine.

PS You can check is a variable is a number with `isNaN(x)`.

image

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -36,10 +36,93 @@ describe('ExtensionRequest', () => {
...
+        data: {
+          agmYear: 2023,
+          isEligible: false
+        },
+        extensionDurationText: 'The business has reached maximum possible extension for this AGM.',
+        dueDateText: 'The business is overdue for this AGM and extension cannot be requested.'
+      },
+      vuetify
+    })
+
```
severinbeauvais, line 116: Please also verify that the yellow box is displayed when not eligible.

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -11,7 +11,7 @@ const vuetify = new Vuetify({})
 setActivePinia(createPinia())
 const rootStore = useRootStore()
 
-describe('ExtensionRequest', () => {
+describe('ExtensionEvaluation', () => {
   it('displays normally', () => {
     // init store
     rootStore.keycloakRoles = ['']
```
severinbeauvais, line 17: Your component doesn't use the store, so you can remove this (and the ones below) and all references to the store above.

src/views/AgmExtension.vue
severinbeauvais: Removed, thanks.

src/views/AgmExtension.vue
severinbeauvais: Removed, thanks.

tests/unit/AgmExtensionEvaluation.spec.ts
severinbeauvais: Removed, thanks.

tests/unit/AgmExtensionEvaluation.spec.ts
severinbeauvais: Fixed, thanks.

tests/unit/AgmExtensionEvaluation.spec.ts
severinbeauvais, line 1: Try to keep the unit tests concise.

Below, you have a lot of duplicate testing.

src/components/AgmExtension/AgmExtensionEvaluation.vue
JazzarKarim: Sev, here's what I was thinking:

This component's non-eligible text will be set in the parent after my logic is done. This is since I believe the text depends on which scenario didn't pass in my component. I might need to follow-up with Yui or Mihai about that though. Not sure if it's the same text for ineligibility for all scenarios or not.

I guess what you're thinking is instead of letting the parent decide the text, we remove these props and have this component set the text depending on whether eligible is true or false? That would work only if the text is the same across all scenarios like I mentioned above though if that makes sense.

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -139,7 +139,7 @@ describe('ExtensionEvaluation', () => {
     const yellowEvalBox = wrapper.find('.message-box')
 
     expect(yellowEvalBox.exists()).toBe(true)
-    expect(yellowEvalBox.isVisible())
+    expect(yellowEvalBox.isVisible()).toBe(true)
```
severinbeauvais, line 142: If the conditional is a `v-if` then check if it exists.

if the conditional is a `v-show` then check if it's visible.

You don't need both here :)

src/components/AgmExtension/AgmExtensionEvaluation.vue
severinbeauvais: Kinda the second one.

For audit purposes, I'm expecting your component to set a value(s) indicating which scenario the business is in (which would get saved in the filing JSON). Then, THIS component can simply look at that value(s) and show the proper text.

tests/unit/AgmExtensionEvaluation.spec.ts
severinbeauvais: You could reduce a lot more things that are optional or whatever, but this is OK for now.

src/components/AgmExtension/AgmExtensionEvaluation.vue
JazzarKarim: I see. OK.

@PaulGarewal Can you please follow-up with Yui/Mihai and ask them what would the text be depending on the ineligibility scenario? The UXPin only shows one.

src/components/AgmExtension/AgmExtensionEvaluation.vue
```diff
@@ -51,7 +52,22 @@
...
+            </template>
+            <template v-else-if="data.requestExpired && data.alreadyExtended">
+              The business has reached maximum possible extension for this AGM.
+              The period to request an extension has expired.
+            </template>
+          </template>
+          <template v-else>
+            {{ data.extensionDuration }} months
+          </template>
+        </v-col>
```
severinbeauvais, line 70: I like it. My only request is, could you keep the variables in the same order in the conditionals?

PaulGarewal: I caught that just now too 😄

src/components/AgmExtension/AgmExtensionEvaluation.vue
severinbeauvais: Fixed, thanks.

src/interfaces/agm-ext-eval-interface.ts
severinbeauvais: 571 is merged. Just watch out that the merge doesn't add your prop in a second time.

tests/unit/AgmExtensionEvaluation.spec.ts
severinbeauvais: Fixed, thanks.

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -36,10 +31,222 @@ describe('ExtensionRequest', () => {
...
+    const rows = wrapper.findAll('.content .row')
+
+    expect(rows.at(0).find('.col-sm-9').text()).toBe('2023')
+
+    expect(rows.at(1).find('.col-sm-9').text()).toBe('months')
+
+    expect(rows.at(2).find('.col-sm-9').text()).toBe('')
+
+    wrapper.destroy()
+  })
```
severinbeauvais, line 118: This is essentially the same test as above, since isEligible=True gives the same results regardless of alreadyExtended.

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -36,10 +31,222 @@ describe('ExtensionRequest', () => {
...
```
severinbeauvais, line 144: Same as above.

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -36,10 +31,222 @@ describe('ExtensionRequest', () => {
...
+    // init store
+
+    const wrapper = mount(AgmExtensionEvaluation, {
+      propsData: {
+        data: {
+          agmYear: 2023,
+          isEligible: true,
+          alreadyExtended: true,
+          requestExpired: true
+        }
```
severinbeauvais, line 78: I think you should pass in extensionDuration and agmDueDate and verify the output of those here.

severinbeauvais (approved):
Looks good. I have some comments on the unit tests. Please get another approval or 2 before merging. Thanks.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-573-n4nq5i3o.web.app

jamespaologarcia (approved):
LGTM

PR #577, opened 2023-11-09
Extension Request Scenario 1

JazzarKarim:
*Issue #:* /bcgov/entity#18273

*Description of changes:*
- Added the Logic for Scenario 1 for Extension Request component
[AGM Extensions Scenarios Final V1.0.docx](https://github.com/bcgov/business-filings-ui/files/13313693/AGM.Extensions.Scenarios.Final.V1.0.docx)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

PR #580, opened 2023-11-14
18273 Extension Request Scenario 2

leodube-aot:
*Issue #:* /bcgov/entity#18273

*Description of changes:*
- Added the Logic for Scenario 1 for Extension Request component

### DIFF
--- a/src/components/AgmExtension/ExtensionRequest.vue
+++ b/src/components/AgmExtension/ExtensionRequest.vue
@@ -250,6 +250,7 @@ export default class ExtensionRequest extends Vue {
     return this.data.isFirstAgm
   }
 
+  /** Called when isFirstAgm radio group changes. */
   @Watch('data.isFirstAgm')
   onIsFirstAgmChanged (val: boolean): void {
     if (val) {
--- a/src/components/AgmExtension/ExtensionRequest.vue
+++ b/src/components/AgmExtension/ExtensionRequest.vue
@@ -172,7 +172,6 @@ import { useRootStore } from '@/stores'
 import { DatePicker } from '@bcrs-shared-components/date-picker'
 import AgmYear from '@/components/AgmLocationChange/AgmYear.vue'
 import { DateUtilities } from '@/services'
-import { CONSOLE_LEVELS } from '@sentry/utils'
 
 @Component({
   components: {
@@ -256,6 +255,20 @@ export default class ExtensionRequest extends Vue {
     } else {
       return true
     }
+    // // Field disabled on first load when value is null
+    // if (this.data.isFirstAgm == null) {
+    //   return true
+    // }
+
+    // // Set agmYear as incorporation date if not null and true
+    // if (this.data.isFirstAgm) {
+    //   this.data.agmYear = this.data.incorporationDate.getFullYear().toString()
+    // } else {
+    //   this.data.agmYear = ''
+    // }
+
+    // // Field disabled when true
+    // return this.isFirstAgm
   }
 
   /** Called when extension date picker changes. */
--- a/src/components/AgmExtension/ExtensionRequest.vue
+++ b/src/components/AgmExtension/ExtensionRequest.vue
@@ -239,22 +239,23 @@ export default class ExtensionRequest extends Vue {
     }
   }
 
-  /**
-   * Whether to disable the editing of AGM Year field.
-   * Editable (false) when Is this first AGM is set to No.
-   * Non-Editable (true) when we first load (null) or Is first AGM is Yes.
-   * Set the agmYear as incorporation date if not null and true.
-   */
+  /** Whether to disable the editing of AGM Year field. */
   get isFirstAgm (): boolean {
-    if (this.data.isFirstAgm !== null) {
-      if (this.data.isFirstAgm) {
-        this.data.agmYear = this.data.incorporationDate.getFullYear().toString()
-        return true
-      } else {
-        return false
-      }
-    } else {
+    // Field disabled on first load when value is null
+    if (this.data.isFirstAgm == null) {
       return true
+    }
+
+    // Field disabled when it is first agm
+    return this.data.isFirstAgm
+  }
+
+  @Watch('data.isFirstAgm')
+  onIsFirstAgmChanged (val: boolean): void {
+    if (val) {
+      this.data.agmYear = this.data.incorporationDate.getFullYear().toString()
+    } else {
+      this.data.agmYear = ''
     }
   }
 
@@ -264,14 +265,14 @@ export default class ExtensionRequest extends Vue {
     this.data.prevExpiryDate = val
   }
 
-  @Watch('intendedAgmDateText')
   /** Called when intended AGM date picker changes. */
+  @Watch('intendedAgmDateText')
   onAgmIntendedDatePickerChanged (val: string): void {
     this.data.intendedAgmDate = val
   }
 
-  @Watch('previousAgmDateText')
   /** Called when previous AGM date picker changes. */
+  @Watch('previousAgmDateText')
   onPreviousAgmDatePickerChanged (val: string): void {
     this.data.prevAgmDate = val
   }
--- a/src/components/AgmExtension/ExtensionRequest.vue
+++ b/src/components/AgmExtension/ExtensionRequest.vue
@@ -165,10 +165,8 @@
 
 <script lang="ts">
 import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
-import { Getter } from 'pinia-class'
 import { VcardTemplate } from '@/components/common'
 import { AgmExtEvalIF } from '@/interfaces'
-import { useRootStore } from '@/stores'
 import { DatePicker } from '@bcrs-shared-components/date-picker'
 import AgmYear from '@/components/AgmLocationChange/AgmYear.vue'
 import { DateUtilities } from '@/services'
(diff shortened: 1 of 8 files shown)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-580-nt8l0mqi.web.app

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -238,23 +239,17 @@ export default class ExtensionRequest extends Vue {
...
-      return 'placeholder for future work'
+      return DateUtilities.addMonthsToDate(16, this.data.prevAgmDate)
     }
   }
 
-  /**
-   * Whether to disable the editing of AGM Year field.
-   * Editable (false) when Is this first AGM is set to No.
-   * Non-Editable (true) when we first load (null) or Is first AGM is Yes.
-   * Set the agmYear as incorporation date if not null and true.
```
severinbeauvais: Please "port" these comments to the code: explain why this getter is different values depending on state.

leodube-aot: Yeah I was attempting to refactor that part a bit but it wasn't working. I have reverted them back to how they were for now, but I may try to get my refactored version working again later.

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -281,48 +276,117 @@ export default class ExtensionRequest extends Vue {
   }
 
   @Watch('data.isPrevExtension')
+  @Watch('previousAgmDateText')
   @Watch('extensionExpiryDateText')
   onIsPrevExtensionChanged (): void {
+    /* eslint-disable brace-style */
+    const currentDate = DateUtilities.yyyyMmDdToDate(this.getCurrentDate)
```
severinbeauvais, line 288: Since the calculations below depend on current date, which could change, please add it as a Watch for this method.

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -281,48 +276,117 @@ export default class ExtensionRequest extends Vue {
   }
 
   @Watch('data.isPrevExtension')
+  @Watch('previousAgmDateText')
   @Watch('extensionExpiryDateText')
   onIsPrevExtensionChanged (): void {
+    /* eslint-disable brace-style */
```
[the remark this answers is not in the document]
leodube-aot: I think we commented on this line at around the same time haha, see my other comment

src/components/AgmExtension/ExtensionRequest.vue
severinbeauvais, line 1: Line 188: could we add current date to the data object instead of having it as a getter? This would do 2 things:
- record the current date in the data object for future auditing
- remove this component's dependency on the store (making for a simpler component to wire up -- think about a car transmission with few connections vs many)

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -281,48 +281,117 @@ export default class ExtensionRequest extends Vue {
...
```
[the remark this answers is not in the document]
severinbeauvais: I see what you're done now.

We didn't have a style guideline for if-then-else formatting other than what lint prescribed, but I wasn't always happy with it -- the code below is a great example.

I say leave this in for now. The code is much more readable like this.

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -281,48 +281,117 @@ export default class ExtensionRequest extends Vue {
...
+      else if ((DateUtilities.daysBetweenTwoDates(cutOffDate, expiryDate)) > 365) {
+        this.data.isEligible = false
+      }
+      // ELSE ELIGIBLE, ExtensionDuration = 12 months - ExpiryDate - PrevAgmDate
+      else {
+        this.data.isEligible = true
+        this.data.extensionDuration = 12 - DateUtilities.subtractDates(
+          this.previousAgmDateText,
+          this.extensionExpiryDateText
+        )
```
severinbeauvais, line 394: In the next PR, let's see about moving some of these blocks into other functions, with the goal of keeping the "high level" logic in here and moving the "details" to the other functions.

src/components/AgmExtension/ExtensionRequest.vue
severinbeauvais: That works for now.

Sometimes you have to make a real mess before you figure out how to clean it up 😆

src/components/AgmExtension/ExtensionRequest.vue
leodube-aot: I agree, even with the comments it is still difficult to understand what is going on at a glance. Moving some of the functionality out of the blocks to other functions should really help this.

src/services/date-utilities.ts
```diff
@@ -368,8 +368,8 @@ export default class DateUtilities {
    * @example (18, 2023-02-03) -> "2024-08-03"
    */
   static addMonthsToDate (month: number, date: string): string {
-    if (!date) {
-      date = '2023-01-01'
+    if (!month || !date) {
+      return null
```
leodube-aot, line 372: This is to address [this comment](https://github.com/bcgov/business-filings-ui/pull/577#discussion_r1389645250) on Karim's previous PR

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -238,28 +235,28 @@ export default class ExtensionRequest extends Vue {
...
     }
   }
 
-  /**
-   * Whether to disable the editing of AGM Year field.
-   * Editable (false) when Is this first AGM is set to No.
-   * Non-Editable (true) when we first load (null) or Is first AGM is Yes.
-   * Set the agmYear as incorporation date if not null and true.
-   */
+  /** Whether to disable the editing of AGM Year field. */
```
leodube-aot, line 242: Refactored this to only handle enabling/disabling the AGM Year field, and moved logic of setting AGM Year into a watcher on line 253. The point of this is that the line `this.data.agmYear` would prevent the agmYear from getting set if `isFirstAgm == false` while it was inside this getter. Moving it to the watcher resolves this issue.

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -268,61 +266,136 @@ export default class ExtensionRequest extends Vue {
...
+        15, this.previousAgmDateText)
+      const cutOffDate = DateUtilities.yyyyMmDdToDate(cutOffYyyyMmDd)
+
+      // IF CurrentDate > (PrevAgmDate + 15 Months + 5 days)
+      // THEN NOT ELIGIBLE
+      if ((DateUtilities.daysBetweenTwoDates(cutOffDate, currentDate) - 5) > 0) {
+        this.data.isEligible = false
+      }
+      // ELSE ELIGIBLE, ExtensionDuration = 6 months
+      else {
```
JazzarKarim, line 365: Leo, I don't see this else statement in Mihai's notes. Did you confirm this by any chance with him? There might be a special case for the extension duration. The else here is probably correct I think but just in case!
rule 2

severinbeauvais:
Your rebase ate your app version changes. Also, there's 2 new unit test errors.

src/components/AgmExtension/ExtensionRequest.vue
leodube-aot: Hey Karim, I found in the non-final scenarios document that there is an "else" statement (the scenario numbers between the 2 docs don't match but this is for `isFirstAgm == false && isPrevExtension == false`. I'll check in with Mihai to be sure however and will update in this thread!

Screenshot 2023-11-14 at 11 56 17 AM

leodube-aot: Just checked with Mihai and the else block is correct, we want to grant the extension on the basis of `isFirstAgm == false && isPrevExtension == false` and it is less than the time constraint. 

It was just missing from the final scenarios document. It was actually in the outdated version and was not copied over the the final version that you have!

tests/unit/ExtensionRequest.spec.ts
```diff
@@ -36,16 +36,13 @@ describe('ExtensionRequest', () => {
     expect(rows.at(1).find('.col-sm-3').text()).toBe('AGM Year')
     expect(rows.at(1).find('.col-sm-9 .v-text-field').exists()).toBe(true)
 
-    expect(rows.at(2).find('.col-sm-3').text()).toBe('Previous AGM date or a reference date')
-    expect(rows.at(2).find('.col-sm-9').findComponent(DatePicker).exists()).toBe(true)
-
```
leodube-aot, line 41: I'm not 100% sure why this worked originally since when the component loads with empty values, this should not be rendered. Even in dev when the extension request first loads this row is not present in the DOM.

severinbeauvais (approved):

src/components/AgmExtension/ExtensionRequest.vue
JazzarKarim: Awesome, thanks a lot Leo! Resolving.

JazzarKarim (approved):
LGTM Leo! 👍 

@severinbeauvais Sev, will you be wiring the `alreadyExtended` and `requestExpired` in your get everything together ticket or will it be done in this ticket you think?

severinbeauvais:
> @severinbeauvais Sev, will you be wiring the `alreadyExtended` and `requestExpired` in your get everything together ticket or should it be done in this ticket you think?

I think it should be done in THIS ticket, though not necessarily this PR. My framework ticket doesn't care about the eligibility fields. THIS ticket is what calculates those fields.

JazzarKarim:
> > @severinbeauvais Sev, will you be wiring the `alreadyExtended` and `requestExpired` in your get everything together ticket or should it be done in this ticket you think?
> 
> I think it should be done in THIS ticket, though not necessarily this PR. My framework ticket doesn't care about the eligibility fields. THIS ticket is what calculates those fields.

Makes sense. OK, got it. Then it can be done in a subsequent PR.

PR #581, opened 2023-11-15
18469 Capitalize AGM in output titles

severinbeauvais:
*Issue #:* bcgov/entity#18469

*Description of changes:*
- app version = 7.0.11
- added missing cases for AGM filings
- added special case to convert "Agm" to "AGM"
- fixed a lint warning

### DIFF
--- a/src/services/enum-utilities.ts
+++ b/src/services/enum-utilities.ts
@@ -299,6 +299,8 @@ export default class EnumUtilities {
       case FilingTypes.ADMIN_FREEZE:
         // FUTURE: add freeze/unfreeze checks here
         return FilingNames.ADMIN_FREEZE
+      case FilingTypes.AGM_EXTENSION: return FilingNames.AGM_EXTENSION
+      case FilingTypes.AGM_LOCATION_CHANGE: return FilingNames.AGM_LOCATION_CHANGE
       case FilingTypes.ALTERATION: return FilingNames.ALTERATION
       case FilingTypes.ANNUAL_REPORT: return FilingNames.ANNUAL_REPORT + (agmYear ? ` (${agmYear})` : '')
       case FilingTypes.CHANGE_OF_ADDRESS: return FilingNames.CHANGE_OF_ADDRESS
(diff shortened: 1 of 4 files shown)


src/services/enum-utilities.ts
```diff
@@ -299,6 +299,8 @@ export default class EnumUtilities {
       case FilingTypes.ADMIN_FREEZE:
         // FUTURE: add freeze/unfreeze checks here
         return FilingNames.ADMIN_FREEZE
+      case FilingTypes.AGM_EXTENSION: return FilingNames.AGM_EXTENSION
+      case FilingTypes.AGM_LOCATION_CHANGE: return FilingNames.AGM_LOCATION_CHANGE
```
severinbeauvais, line 303: I think this method isn't currently used for the AGM filings, but it's better for this method to be complete anyway.

src/services/enum-utilities.ts
```diff
@@ -332,13 +334,15 @@ export default class EnumUtilities {
...
+   * Converts a string in "camelCase" (or "PascalCase") to a string of separate, title-case words,
    * suitable for a title or proper name.
    * @param s the string to convert
    * @returns the converted string
    */
   static camelCaseToWords (s: string): string {
-    return s?.split(/(?=[A-Z])/).join(' ').replace(/^\w/, c => c.toUpperCase()) || ''
+    const words = s?.split(/(?=[A-Z])/).join(' ').replace(/^\w/, c => c.toUpperCase()) || ''
+    // SPECIAL CASE: convert 'Agm' to uppercase
+    return words.replace('Agm', 'AGM')
```
severinbeauvais, line 345: The alternative is a large effort to include the document name in the API response (requiring API work and possibly informing third parties/delays) and then use that.

This is a special case, but it's a simple one, and we could add more here if needed, so I think it's not the best solution but it's a solution we can afford and it's functional and understandable and safe. Thoughts?

severinbeauvais:
Before:
image

After:
image

src/services/enum-utilities.ts
JazzarKarim: I think it's OK for now Sev. If more and more arise in the future, then we can think about going the long and hard way.

JazzarKarim (approved):
LGTM!

ketaki-deodhar (approved):
Solution without big changes for now. this works well. LGTM!

PR #585, opened 2023-11-17
18273+18558 Extension Request Cleanup and Tests

leodube-aot:
*Issue #:* /bcgov/entity#18273 and /bcgov/entity#18558

*Description of changes:*
- Connected `alreadyExtended` and `requestExpired` for `AgmExtensionEvaluation.vue`.
- Implemented `extensionRequestValid` and emit result to parent component.
- Added `evaluateResult` flag to only show extension request warning box and duration of extension text field on valid extension request.
- Moved scenario logic into separate functions called from a `checkEligibility()` function.
- Added unit tests for `ExtensionRequest` component to cover every scenario.
- Updated unit tests for `AgmExtensionEvaluation`
- Created unit tests for `date-utilities`


src/components/AgmExtension/AgmExtensionEvaluation.vue
```diff
@@ -10,7 +10,7 @@
 
     <template #content>
       <v-card
-        v-if="!data.isEligible"
+        v-if="evaluateResult && !data.isEligible"
```
leodube-aot, line 13: Added `evaluateResult` to only display the yellow message box, as well as the custom text in "Duration of Extension" when the extension request is valid (i.e. all fields for a specific scenario are entered)

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -180,27 +183,18 @@ import { DateUtilities } from '@/services'
...
   intendedAgmDateText = ''
-
-  /** The previous AGM date (reference date) date text. */
   previousAgmDateText = ''
 
-  isEligible = false // whether the extension can be granted.
-
-  @Emit('update:data')
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  private emitData (data: AgmExtEvalIF): void {}
```
leodube-aot, line 199: Moved emit function to bottom of file

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -209,14 +203,36 @@ export default class ExtensionRequest extends Vue {
     ]
   }
 
+  /** Whether the extension request is valid or not */
+  get extensionRequestValid (): boolean {
+    const isFirstAgmValid = this.data.isFirstAgm && !!this.data.agmYear
+    const isSubsequentAgmValid = (
+      this.data.isFirstAgm === false &&
```
leodube-aot, line 210: I specifically check `=== false` a few times in this file. That's because some values in the `data` object can be `null`, but that doesn't mean they are valid. 

For example, the answers to "Is this the first AGM?" can be "Yes" or "No", with values `true` and `false` respectively. To check `isSubsequentAgmValid`, then the "No" field must be ticked (which evaluates to false). However when neither field is ticked (not a valid extension request), it remains `null`. Therefore there can be no ambiguity between null and false.

Screenshot 2023-11-17 at 1 22 49 PM

tests/unit/date-utilities.spec.ts
```diff
@@ -0,0 +1,190 @@
+import { DateUtilities } from '@/services'
```
leodube-aot, line 1: I created this test file for the `date-utilities`, then copied over the tests from `date-mixins.spec.ts` and modified them a bit to get them to work. The `date-mixins` are deprecated but are still used in a few places in the application.

At the bottom I added new tests for `addMonthsToDate` and `subtractDates`

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-585-c2c631r7.web.app

severinbeauvais:
This should probably throw an error, I think?

image

src/components/AgmExtension/AgmExtensionEvaluation.vue
severinbeauvais: This partly resolves #18558. Want to take that ticket and fix it here?

severinbeauvais: PS You don't have to.

leodube-aot:
@severinbeauvais Yeah you're correct, there should be more clear error handling. I'll get that fixed. The one field I can confirm we don't validate is the Intended AGM date (based on what Mihai has said), but perhaps at should still not allow dates in the past for that one.

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -255,149 +271,206 @@ export default class ExtensionRequest extends Vue {
...
   }
 
-  @Watch('data.currentDate')
-  @Watch('data.isPrevExtension')
-  @Watch('previousAgmDateText')
-  @Watch('extensionExpiryDateText')
-  onIsPrevExtensionChanged (): void {
-    /* eslint-disable brace-style */
+  /** Eligibility for first agm with no extension requested this year */
+  /* eslint-disable brace-style */
```
severinbeauvais, line 320: What's wrong with the brace style, and should you re-enable this check below this  method?

src/components/AgmExtension/AgmExtensionEvaluation.vue
leodube-aot: I don't mind fixing it while I'm already working in the code!

src/components/AgmExtension/AgmExtensionEvaluation.vue
```diff
@@ -108,6 +108,7 @@ import { DateUtilities } from '@/services'
 })
 export default class AgmExtensionEvaluation extends Vue {
   @Prop({ required: true }) readonly data!: AgmExtEvalIF
+  @Prop({ default: false }) readonly evaluateResult!: boolean
```
severinbeauvais, line 111: This is specific enough, and not part of "data", that I'm OK with this being a new prop 👍

src/components/AgmExtension/ExtensionRequest.vue
leodube-aot: Mainly the linter expects the next conditional block to start right away:

```
if () {
} else {
}
```

Which doesn't really allow for comments between them like this:

```
// if block
if () {
}

// else block
else {
}
``` 

Since the logic is a bit tricky to understand I separated each block, but it's mainly a stylistic/readability choice. I could also have the comments inside the conditional blocks but I find it slightly less readable.

But yeah it's probably good to re-enable the brace style afterwards, I'll do that.

severinbeauvais: Sounds good. You may resolve this conversation after re-enabling the eslint rule.

leodube-aot:
**Newest changes:**
- Fixed bcgov/entity#18558
- Added more validation to the text and date picker components, also made it so that if the extension request is invalid, the specific section(s) that are invalid show the red border, instead of the entire component.
- Need to fix a couple unit tests still

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -11,9 +11,12 @@
     <template #content>
       <div
         class="px-6 py-7"
-        :class="{ 'invalid-section': !extensionRequestValid && showErrors }"
+        :class="{ 'invalid-section': !firstSectionValid && showErrors }"
```
leodube-aot, line 14: Added specific valid flags for each section of the component to highlight specific areas that are not valid.

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -11,9 +11,12 @@
...
       >
-        <v-row no-gutters>
+        <v-row
+          v-if="!is30MonthsAfterIncorp"
```
leodube-aot, line 17: The UXPin says that if incorporation date is 30 months ago, it is safe to assume that it is not the first agm extension, so hide the radio group and only show the `agmYear` textbox.

src/interfaces/agm-ext-eval-interface.ts
```diff
@@ -25,7 +25,11 @@ export const EmptyAgmExtEval: AgmExtEvalIF = {
   incorporationDate: null,
   isFirstAgm: null,
   agmYear: null,
+  prevAgmDate: null,
+  isPrevExtension: null,
+  prevExpiryDate: null,
+  intendedAgmDate: null,
```
leodube-aot, line 31: Needed to explicitly set values for the `EmptyAgmExtEval`, otherwise the date pickers don’t update the dates properly. I think it is likely because if you `@Watch('data.intendedAgmDate')` and `data.intendedAgmDate` is undefined, when it initially gets set it doesn't seem to trigger the watcher.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-585-c2c631r7.web.app

src/interfaces/agm-ext-eval-interface.ts
severinbeauvais: You are correct. If a variable is undefined, it's not reactive. It's like it doesn't exist (which is kind of the definition of undefined 😀 ).

src/components/AgmExtension/AgmExtensionEvaluation.vue
```diff
@@ -19,78 +19,93 @@
           an extension cannot be granted.
         </p>
       </v-card>
-
-      <v-row
-        no-gutters
-        class="px-6 pt-7"
+      <div
+        v-else-if="!evaluateResult"
+        style="overflow: auto;"
```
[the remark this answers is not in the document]
leodube-aot: Yeah so the issue I was trying to resolve is that when only displaying the message box, the bottom margins get hidden in the overflow (a bit tough to see in these screenshots, but the message box is aligned at the bottom of the v-card).

Screenshot 2023-11-20 at 10 30 05 AM
Screenshot 2023-11-20 at 10 30 22 AM

And adding padding will just expand the size of the message box. I tried `d-flex` and `flex-grow-1` but they didn't seem to work. 

I'll add a style property to base.scss and use that instead.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-585-c2c631r7.web.app

src/components/AgmExtension/AgmExtensionEvaluation.vue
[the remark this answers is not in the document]
severinbeauvais: LGTM

severinbeauvais:
Leo, does this PR complete everything for AGM Extensions? If not, what's left?

severinbeauvais:
Leo, should there be a check that the intended AGM date is not later than the due date? Maybe this wouldn't cause a validation error but could display the yellow box with some message. Check with Mihai and Yui.

leodube-aot:
@severinbeauvais I checked and there is one more formula to be added to the AGM extension calculation bcgov/entity#18579. Otherwise, this ticket has completed all the hooking up required.

For the intended date, [Mihai's comment](https://github.com/bcgov/entity/issues/18053#issuecomment-1813296398) explains why we don't validate it.

severinbeauvais:
> Otherwise, this ticket has completed all the hooking up required.

Yeah, it looks that way -- thank you! But I am still unable to file an AGM extension. Do we know if this is solely an API issue?

src/interfaces/agm-ext-eval-interface.ts
severinbeauvais: If `prevAgmDate`, `isPrevExtension`, `prevExpiryDate` and `intendedAgmDate` are no longer optional then please remove the "?"s in the interface above. Thx.

tests/unit/AgmExtensionEvaluation.spec.ts
```diff
@@ -2,29 +2,35 @@ import Vue from 'vue'
...
+  let wrapper: any
 
-    const wrapper = mount(AgmExtensionEvaluation, {
+  beforeEach(() => {
+    wrapper = mount(AgmExtensionEvaluation, {
       propsData: {
         data: {
-          agmYear: 2023
-        }
+          ...EmptyAgmExtEval,
```
severinbeauvais, line 18: Nice use of destructuring 👍

tests/unit/ExtensionRequest.spec.ts
```diff
@@ -2,48 +2,351 @@ import Vue from 'vue'
...
+    expect(firstAgmRadios.at(0).attributes('aria-checked')).toBe('true')
+    expect(firstAgmRadios.at(1).attributes('aria-checked')).toBe('false')
+
+    // Verify AGM year text field
+    expect(rows.at(1).find('#year-txt').attributes().disabled).toBeDefined()
+    expect(wrapper.vm.data.agmYear).toBe('2023')
+
+    // Verify isPrevExtension radio group
+    expect(prevExtensionRadios.at(0).attributes('aria-checked')).toBe('false')
+    expect(prevExtensionRadios.at(1).attributes('aria-checked')).toBe('true')
```
severinbeauvais, line 85: I think there's a way to directly check if the radio button wrapper is checked but I can't think of it off the top of my head and this is OK. FYI.

tests/unit/ExtensionRequest.spec.ts
```diff
@@ -2,48 +2,351 @@ import Vue from 'vue'
...
+    })
+    await Vue.nextTick() // wait for DOM to update
+
+    // Not eligible because:
+    // CurrentDate > (IncorporationDate + 18 Months + 5 days)
+    expect(wrapper.vm.data.isEligible).toBe(false)
+    expect(wrapper.vm.data.extensionDuration).toBe(null)
+    expect(wrapper.vm.data.alreadyExtended).toBe(false)
+    expect(wrapper.vm.data.requestExpired).toBe(true)
+  })
```
severinbeauvais, line 212: Nice 👍

src/components/AgmExtension/ExtensionRequest.vue
```diff
@@ -239,165 +287,230 @@ export default class ExtensionRequest extends Vue {
...
-  @Watch('intendedAgmDateText')
-  onAgmIntendedDatePickerChanged (val: string): void {
-    this.data.intendedAgmDate = val
+  /** Watches all data properties to keep track of changes. */
+  @Watch('data.isFirstAgm')
+  @Watch('data.prevAgmDate')
+  @Watch('data.isPrevExtension')
+  @Watch('data.prevExpiryDate')
+  @Watch('data.intendedAgmDate')
+  @Watch('data.currentDate')
```
severinbeauvais, line 322: Would it make sense to use `@Watch('data', { deep: true })` here, or are there data properties you don't want to watch?

severinbeauvais (approved):

PaulGarewal (approved):
LGTM, amazing job Leo 🎉

src/components/AgmExtension/ExtensionRequest.vue
leodube-aot: I don't think so, since when this watcher triggers the `checkEligibility()` function, we are setting a bunch of data properties like `isElgibile`, `extensionDuration`, `requestExpired`, etc. that we don't necessarily want to watch for changes to.

leodube-aot:
> Yeah, it looks that way -- thank you! But I am still unable to file an AGM extension. Do we know if this is solely an API issue?

I think it's just an API issue. The [saveFiling() function in AgmExtension](https://github.com/bcgov/business-filings-ui/blob/main/src/views/AgmExtension.vue#L336) throws an axios error: "Request failed with status code 422" when "File and Pay" is clicked, which prevents redirection to the payment page.

src/interfaces/agm-ext-eval-interface.ts
severinbeauvais: LGTM

PR #587, opened 2023-11-20
fix: DBC feature should be available for historical companies

amanji:
*Issue #:* /bcgov/entity#18498  /bcgov/entity#18652

*Description of changes:*

* Updates EntityMenu logic to display More Actions when company is not active.
  * Still hides menu options that should not be available to historical companies

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

PR #588, opened 2023-11-20
fix: DBC feature should only be available to BCSC user

amanji:
*Issue #:* /bcgov/entity#18652

*Description of changes:*

* Adds logic to only allow DBC feature for a BCSC login source


amanji:
This is currently base on #587 so the number of impacted files will be reduced once the former is merged

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-588-tt22s0q6.web.app

severinbeauvais:
> This is currently base on #587 so the number of impacted files will be reduced once the former is merged

Let's keep it simple -- #587 can be used for these and the previous changes.

Closing this one.

PR #589, opened 2023-11-21
feat: DBC feature is only available to self-registered owner/operators

amanji:
*Issue #:* /bcgov/entity#18688

*Description of changes:*

* Adds additional logic to gate DBC feature by current business rules
* Adds store methods and tests


src/interfaces/filing-registration-interface.ts
```diff
@@ -0,0 +1,14 @@
+import { PartyIF } from './party-interface'
+
+/** A filing's registration object from the API, ie, `filing.registration`. */
+export interface FilingRegistraionIF {
```
severinbeauvais, line 4: Typo.

severinbeauvais: Also, don't need commas after properties.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -16,6 +16,33 @@ export default class AllowableActionsMixin extends Vue {
...
+
+    const { firstname: userFirstName, lastname: userLastName } = this.getUserInfo
+
+    return (
+      completingParty.firstName === proprietor.firstName &&
+      completingParty.lastName === proprietor.lastName &&
+      proprietor.firstName === userFirstName &&
+      proprietor.lastName === userLastName
+    )
+  }
```
severinbeauvais, line 45: Move helper functions below the main (public) function.

src/stores/filingHistoryListStore.ts
```diff
@@ -156,19 +161,32 @@ export const useFilingHistoryListStore = defineStore('filingHistoryList', {
...
-      })
+    async loadFilings (id: string): Promise<any> {
+      try {
+        const filings = await LegalServices.fetchFilings(id)
+        this.setFilings(filings)
+        return filings
+      } catch (error) {
+        console.log('loadFilings() error =', error)
+      }
+    },
```
severinbeauvais, line 172: I think this didn't work when we first tried it, because it's an action. (That may have been a limitation in Vuex but we now use Pinia.) So it works? Nice cleanup, thanks!

Does this return a more specific promise now?

amanji: I can update the return type

src/App.vue
```diff
@@ -496,6 +497,9 @@ export default {
       // now that we have business info, load state filing
       await this.loadStateFiling()
 
+      // now that we have business info, load the registration filing
+      await this.loadRegistrationFiling()
```
severinbeauvais, line 501: Can't you just use the current parties that were fetched on line 487 and stored on line 495?

src/mixins/allowable-actions-mixin.ts
```diff
@@ -16,6 +16,33 @@ export default class AllowableActionsMixin extends Vue {
...
+   */
+  private get isSelfRegisteredOwnerOperator (): boolean {
+    const filing = this.getRegistrationFiling
+    if (!filing) return false
+
+    const { parties } = filing?.registration as FilingRegistraionIF
+    if (!parties) return false
+
+    const { officer: completingParty } =
+      parties.find(party => party.roles.find(role => role.roleType === Roles.COMPLETING_PARTY))
```
severinbeauvais, line 32: Why do you need the completing party of the original registration?

src/App.vue
amanji: I believe those only return the business parties so I wouldn't have info on who the filing party was.

src/mixins/allowable-actions-mixin.ts
amanji: The current rule for the DBC is that the registering owner and operator must be the same individual. Not sure if there's a different way to determine this? I've based it off of what I was told by @argush3 but it's entirely possible I have misinterpreted that.

src/App.vue
severinbeauvais: OK. Resolving.

severinbeauvais, line 501: Can you make specific to your needs please? -> only call this if you're going to use it for DCB (sole prop, not staff, etc)

src/interfaces/filing-history-list-state-interface.ts
```diff
@@ -1,4 +1,5 @@
 import { ApiFilingIF } from '@/interfaces'
+import { FilingRegistraionIF } from './filing-registration-interface'
```
severinbeauvais, line 2: Typo.

And please remove commas after properties in the interfaces you touch. Thx.

src/mixins/allowable-actions-mixin.ts
severinbeauvais: I'm not sure what "registering owner" is.

Note that sometimes staff file registrations, and I'm not sure who the completing party would be then. @argush3 ?

severinbeauvais:
Overall, isn't the check done here something the Legal API should do as part of allowable actions?

cc: @argush3

argush3:
> Overall, isn't the check done here something the Legal API should do as part of allowable actions?
> 
> cc: @argush3

Yes, ideally we would have a check in allowable actions.

I'm not sure how we flag that check to occur though.

src/mixins/allowable-actions-mixin.ts
argush3: Well, my understanding is to use DC, you need to be logged in as the actual proprietor via BCSC.  Is this not the case?

If it is the case that you need to be logged in as the proprietor, then all 3(proprietor, completing party and logged in user) names should match when filing the registration.

Is staff scenario actually allowed for DC users?

severinbeauvais: I am wondering whether the completing party is always the same as the proprietor. What if staff, or a lawyer, filed the registration?

Isn't is enough to compare the proprietor with the current user?

(You're right, staff can't access DBC, but they might have filed the original registration.)

severinbeauvais:
> > Overall, isn't the check done here something the Legal API should do as part of allowable actions?
> > cc: @argush3
> 
> Yes, ideally we would have a check in allowable actions.
> 
> I'm not sure how we flag that check to occur though.

Well, the Legal API knows who the current user is and can compare with the current business proprietor, right?

src/mixins/allowable-actions-mixin.ts
argush3: I don't know the business rules well enough.

I thought to use DBC, the actual proprietor needs to do the registration.  i.e. a staff or a lawyer can't do it for them

argush3:
> Well, the Legal API knows who the current user is and can compare with the current business proprietor, right?

If we are talking about blocking the registration filing, we can probably check the logged in user against the proprietor.

But I need to know when that check needs to happen.

Not all SPs will be using DBC and so shouldn't be blocked on this condition.  

Probably a bit hard to talk about this in this PR.

src/mixins/allowable-actions-mixin.ts
severinbeauvais: I'm trying to get an answer for this.

amanji: Argus is correct. At least for the MVP, all three need to be the same. In later phases this may be relaxed.

severinbeauvais:
> > Well, the Legal API knows who the current user is and can compare with the current business proprietor, right?
> 
> If we are talking about blocking the registration filing, we can probably check the logged in user against the proprietor.

No, this is to block access to a Sole Prop. The original registration was in the past. The check may need to find that registration and compare the completing party, not sure yet.

> But I need to know when that check needs to happen.

I'm hoping the allowable actions from the Legal API can simply return "not allowed" for DBC if the current user isn't the proprietor (and perhaps the completing party).

> Not all SPs will be using DBC and so shouldn't be blocked on this condition.

Right. The only thing that might be blocked is access to the DBC action/pages depending on the current user.

src/App.vue
amanji: We'd need this in order to display the EntityMenu, is there a better place this logic can reside?

amanji: Or are you saying we should just wrap this in a conditional?

src/mixins/allowable-actions-mixin.ts
severinbeauvais: Our BA says,
image

But OK for MVP.

src/App.vue
severinbeauvais: Yes

src/stores/filingHistoryListStore.ts
severinbeauvais: LGTM

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-589-fwm9m23s.web.app

src/stores/filingHistoryListStore.ts
severinbeauvais: I have confirmed that the new code works. Hooray! I hated the previous code.

src/App.vue
severinbeauvais: In fact, I think you could make `loadRegistrationFiling` a store getter and not have to save the restoration filing, etc.

But I'm still hoping we don't need any of this and the user/comp party/prioprietor can be checked in the Legal API and we can use allowable actions to determine whether to display the menu item or not.

severinbeauvais: > But I'm still hoping we don't need any of this and the user/comp party/prioprietor can be checked in the Legal API and we can use allowable actions to determine whether to display the menu item or not.

@argush3 Does this make sense? Does it definitely belong in the Legal API not UI? It is easy to do?

amanji: > In fact, I think you could make `loadRegistrationFiling` a store getter and not have to save the restoration filing, etc.
> 
> But I'm still hoping we don't need any of this and the user/comp party/prioprietor can be checked in the Legal API and we can use allowable actions to determine whether to display the menu item or not.

My only concern is that the data will be fetched every time is `isAllowed` method is run. In this case it's cached. Unless Pinia does this already

severinbeauvais: Yes, getters are cached.

I'm going to hold off further comments until we see about using allowable actions instead.

amanji:
Im thinking about this a bit more and I do agree we should create a rule engine in the API to expose this feature. It could further be used to protect the DC API endpoints.

severinbeauvais:
> Im thinking about this a bit more and I do agree we should create a rule engine in the API to expose this feature. It could further be used to protect the DC API endpoints.

Yes, the "allowable actions" is the API rule engine already. We should keep investigating this approach.

> If this can be used as a stop-gap in the meantime we can always issue a PR to revert this out of the codebase since it's a pretty isolated change.

I'd be ok with this. Will there be time left in your project to change this later?

severinbeauvais:
### Test Results

With proprietor == completing party == current user:
https://business-filings-dev--pr-589-fwm9m23s.web.app/FM1059658?accountid=2288

With proprietor != completing party == current user:
https://business-filings-dev--pr-589-fwm9m23s.web.app/FM1019176/?accountid=2288

src/mixins/allowable-actions-mixin.ts
```diff
@@ -185,4 +186,30 @@ export default class AllowableActionsMixin extends Vue {
...
+    const filing = this.getRegistrationFiling
+    if (!(filing && filing.registration)) return false
+
+    const { parties } = filing.registration
+    if (!parties) return false
+
+    const { officer: completingParty } =
+        parties.find(party => party.roles.find(role => role.roleType === Roles.COMPLETING_PARTY))
+    const { officer: proprietor } =
+        parties.find(party => party.roles.find(role => role.roleType === Roles.PROPRIETOR))
```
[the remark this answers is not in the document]
severinbeauvais: Never mind. The answer is YES, a proprietor can be changed, but the initial registration is not modified.

amanji:
> > Im thinking about this a bit more and I do agree we should create a rule engine in the API to expose this feature. It could further be used to protect the DC API endpoints.
> 
> 
> 
> Yes, the "allowable actions" is the API rule engine already. We should keep investigating this approach.
> 
> 
> 
> > If this can be used as a stop-gap in the meantime we can always issue a PR to revert this out of the codebase since it's a pretty isolated change.
> 
> 
> 
> I'd be ok with this. Will there be time left in your project to change this later?

I think so. I have some runway to add in tests for the queue so I might as well add in a second PR for this work. I don't imagine it would take long because we know exactly what rule to implement.

severinbeauvais:
> > Will there be time left in your project to change this later?
> 
> I think so. I have some runway to add in tests for the queue so I might as well add in a second PR for this work. I don't imagine it would take long because we know exactly what rule to implement.

OK, then make a quick plan with Argus on how to do this, and if that looks feasible then I will approve this PR for merging. Thx.

severinbeauvais:
Is this feature (users check) required for MVP?

If so, and if we have an actionable plan to implement this in the API instead, then I will approve this PR. Otherwise, I'd rather not have this code in the UI for an undetermined period of time.

amanji:
Closing this as logic has been moved to API. Will open a new PR

PR #594, opened 2023-11-28
18024 Finish AGM Extension UI implementation 

leodube-aot:
*Issue #:* /bcgov/entity#18024

*Description of changes:*
- Fix api payload for agm extension to properly file the filing.
- Update DatePicker components to latest (add back date rules as now it validates properly).
- Fix AGM Extension tooltip (bcgov/entity#18680)


src/components/EntityInfo/EntityMenu.vue
```diff
@@ -168,17 +168,21 @@
...
+                      id="agm-ext-list-item"
+                      :disabled="!isAllowed(AllowableActions.AGM_EXTENSION)"
+                      v-on="on"
+                      @click="goToAgmExtensionFiling()"
+                    >
+                      <v-list-item-title>
+                        <span class="app-blue">Request AGM Extension</span>
+                      </v-list-item-title>
+                    </v-list-item>
+                  </div>
```
leodube-aot, line 185: Wrapping the `v-list-item` in a div with a `v-if` appears to fix bcgov/entity#18680

src/views/AgmExtension.vue
```diff
@@ -431,12 +428,15 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin,
...
         extReqForAgmYear: this.data.isPrevExtension,
-        expireDateCurrExt: this.data.prevExpiryDate,
         intendedAgmDate: this.data.intendedAgmDate,
         totalApprovedExt: this.data.extensionDuration,
         expireDateApprovedExt: this.data.agmDueDate,
+        // conditionally add properties if not null
+        ...(this.data.prevAgmDate && { prevAgmRefDate: this.data.prevAgmDate }),
+        ...(this.data.prevExpiryDate && {
+          expireDateCurrExt: this.data.prevExpiryDate
+        }),
```
leodube-aot, line 439: The schema validation fails when `prevAgmRefDate = null` or `expireDateCurrExt = null`. It expects it to either be a string or None. To resolve this we need to confitionally add the properties if their respective `data` values exist. 

Syntax is a bit weird, I took inspiration [from here](https://stackoverflow.com/questions/40652364/add-property-to-object-when-its-not-null), but I can move it out of the object assignment into explicit if statements if that'll look cleaner.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Is it the same template structure as for the other AGM filing?

leodube-aot: It is the same as for AGM Location Change, which also uses a custom tool tip message depending on whether the action is allowed or not.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-594-v8n1w8g8.web.app

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Weird.

Anyways, LGTM.

src/views/AgmExtension.vue
severinbeauvais: That's one way to do it.

Another way is to add in the properties outside this block, ie, after this block do "if ... then data.prevAgmDate = xxx".

Another way to do it is to unconditionally add the properties here and then delete them later, ie, after this block do "if (!data.prevAgmDate) delete data.prevAgmDate".

I'll let you pick whichever one you prefer (including the existing code).

JazzarKarim:
Merging now and will handle any issues/bugs that might pop up after this PR.

PR #600, opened 2023-12-15
18805+18979 Amalgamate Button + Moved to amalgamationApplication + Selection Panel

JazzarKarim:
*Issues #:* bcgov/entity#18805 and bcgov/entity#18979

*Description of changes:*
- Implemented the amalgamate button with it's rules and allowable actions
- Changed from amalgamation to amalgamationApplication + all accompanying changes
- Implemented the Amalgamation Selection Panel

UXPin: https://preview.uxpin.com/7f4cfe5fe4b0ba15a83d518177efe82a08e99167#/pages/160396524/simulate/sitemap

final result

### DIFF
--- a/src/components/EntityInfo.vue
+++ b/src/components/EntityInfo.vue
@@ -1,5 +1,6 @@
 <template>
   <div
+    v-if="!isAmalgamationSelectionRoute"
     id="entity-info"
     :class=" { 'staff': isRoleStaff, 'hover': showHoverStyle }"
   >
@@ -81,6 +82,10 @@ export default class EntityInfo extends Vue {
       this.$route?.name === Routes.STANDALONE_ADDRESSES ||
       this.$route?.name === Routes.STANDALONE_DIRECTORS
     )
+  }
+
+  get isAmalgamationSelectionRoute (): boolean {
+    return this.$route.matched.some(route => route.path.includes(Routes.AMALGAMATION_SELECTION))
   }
 
   get isDCRoute (): boolean {
--- a/src/components/EntityInfo.vue
+++ b/src/components/EntityInfo.vue
@@ -1,6 +1,5 @@
 <template>
   <div
-    v-if="!isAmalgamationSelectionRoute"
     id="entity-info"
     :class=" { 'staff': isRoleStaff, 'hover': showHoverStyle }"
   >
@@ -82,10 +81,6 @@ export default class EntityInfo extends Vue {
       this.$route?.name === Routes.STANDALONE_ADDRESSES ||
       this.$route?.name === Routes.STANDALONE_DIRECTORS
     )
-  }
-
-  get isAmalgamationSelectionRoute (): boolean {
-    return this.$route.matched.some(route => route.path.includes(Routes.AMALGAMATION_SELECTION))
   }
 
   get isDCRoute (): boolean {
--- a/src/components/EntityInfo/EntityMenu.vue
+++ b/src/components/EntityInfo/EntityMenu.vue
@@ -47,7 +47,7 @@
         top
         content-class="top-tooltip"
         transition="fade-transition"
-        :disabled="!isAdminFrozen && !isPendingDissolution"
+        :disabled="!isAdminFrozen && !isPendingDissolution && !amalgamateTooltipText"
       >
         <template #activator="{ on }">
           <span v-on="on">
@@ -321,7 +321,7 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
     } else if (this.isPendingDissolution) {
       return 'This business has a future effective dissolution and cannot be involved in an amalgamation.'
     }
-    return ''
+    return null
   }
 
   /**
--- a/src/components/EntityInfo/EntityMenu.vue
+++ b/src/components/EntityInfo/EntityMenu.vue
@@ -50,6 +50,7 @@
           text
           color="primary"
           :disabled="!isAllowed(AllowableActions.AMALGAMATION)"
+          @click="goToAmalgamationSelection()"
         >
           <v-icon medium>mdi-domain-plus</v-icon>
           <span class="font-13 ml-1">Amalgamate</span>
@@ -69,6 +70,7 @@
                 text
                 color="primary"
                 :disabled="!isAllowed(AllowableActions.AMALGAMATION)"
+                @click="goToAmalgamationSelection()"
                 v-on="on"
               >
                 <v-icon medium>mdi-domain-plus</v-icon>
@@ -361,6 +363,11 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
       return
     }
     this.emitConfirmDissolution()
+  }
+
+  goToAmalgamationSelection (): void {
+    // 0 means "new filing"
+    this.$router.push({ name: Routes.AMALGAMATION_SELECTION, params: { filingId: '0' } })
   }
 
   goToConsentContinuationOutFiling (): void {
--- a/src/components/EntityInfo/EntityMenu.vue
+++ b/src/components/EntityInfo/EntityMenu.vue
@@ -366,7 +366,6 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
   }
 
   goToAmalgamationSelection (): void {
-    // 0 means "new filing"
     this.$router.push({ name: Routes.AMALGAMATION_SELECTION, params: { filingId: '0' } })
   }
 
--- a/src/components/EntityInfo/EntityMenu.vue
+++ b/src/components/EntityInfo/EntityMenu.vue
@@ -41,6 +41,46 @@
       </v-tooltip>
     </span>
 
+    <!-- Amalgamate -->
+    <span v-if="isBusiness && !isHistorical">
+      <template v-if="!isAdminFrozen && !isPendingDissolution">
+        <v-btn
+          id="amalgamate-button"
+          small
+          text
+          color="primary"
+          :disabled="!isAllowed(AllowableActions.AMALGAMATION)"
+        >
+          <v-icon medium>mdi-domain-plus</v-icon>
+          <span class="font-13 ml-1">Amalgamate</span>
+        </v-btn>
+      </template>
+      <template v-else>
+        <v-tooltip
+          top
+          content-class="top-tooltip"
+          transition="fade-transition"
+        >
+          <template #activator="{ on }">
+            <span v-on="on">
+              <v-btn
+                id="amalgamate-button"
+                small
+                text
+                color="primary"
+                :disabled="!isAllowed(AllowableActions.AMALGAMATION)"
+                v-on="on"
+              >
+                <v-icon medium>mdi-domain-plus</v-icon>
+                <span class="font-13 ml-1">Amalgamate</span>
+              </v-btn>
+            </span>
+          </template>
+          {{ amalgamateTooltipText }}
+        </v-tooltip>
+      </template>
+    </span>
+
     <!-- Download Business Summary -->
     <span v-if="isAllowed(AllowableActions.BUSINESS_SUMMARY)">
       <v-tooltip
@@ -237,6 +277,7 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
 
   @Getter(useConfigurationStore) getEditUrl!: string
   @Getter(useBusinessStore) getIdentifier!: string
+  @Getter(useBusinessStore) isAdminFrozen!: boolean
   @Getter(useBusinessStore) isBenBcCccUlc!: boolean
   @Getter(useBusinessStore) isHistorical!: boolean
   @Getter(useRootStore) isPendingDissolution!: boolean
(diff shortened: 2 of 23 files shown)


src/components/EntityInfo/EntityMenu.vue
```diff
@@ -41,6 +41,46 @@
       </v-tooltip>
     </span>
 
+    <!-- Amalgamate -->
+    <span v-if="isBusiness && !isHistorical">
```
JazzarKarim, line 45: If business is historical, we won't show the button.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -41,6 +41,46 @@
...
+      <template v-if="!isAdminFrozen && !isPendingDissolution">
```
JazzarKarim, line 46: amalgamate button cases

src/components/EntityInfo.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div
+    v-if="!isAmalgamationSelectionRoute"
```
JazzarKarim, line 3: Hiding entity info when if we're on the amalgamation selection panel

src/views/AmalgamationSelection.vue
```diff
@@ -0,0 +1,265 @@
...
+
+  /** Get the amalmagated company result name depending on the type. */
+  getRegularAmalgamationText (): string {
+    if (this.isBComp || this.isBcCompany) return 'BC limited company'
+    if (this.isCcc) return 'BC community contribution company'
+    if (this.isUlc) return 'BC unlimited liability company'
+  }
+
+  /** Start Regular Long-form button pressed. */
+  async startRegularAmalgamation (): Promise<any> {
```
JazzarKarim, line 161: Creation of draft and redirecting is a WIP.

src/views/AmalgamationSelection.vue
```diff
@@ -0,0 +1,265 @@
...
+  color: $gray7;
+}
+
+h1 {
+  margin-bottom: 1.25rem;
+  line-height: 2rem;
+  letter-spacing: -0.01rem;
+}
+
+// Have all the 3 options (cards) with same height.
```
JazzarKarim, line 240: Took me some time to perfect this.

Because one card had more text than the other, the sizes and positioning were all messed up.

The code here and below does multiple things. It forces all the cards to be of same size and positions the buttons all at the same level by centering it in the card and pushing it to the bottom.

Tested this also for smaller screens, works well!

Smaller screen:
smaller screens

Phones:
phones screen

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-600-gbnskwh4.web.app

src/components/EntityInfo.vue
severinbeauvais: I recommend moving your conditional to App.vue (which is where the other conditional code for this exists) -- keep it all in one place.

src/components/EntityInfo.vue
```diff
@@ -83,6 +84,10 @@ export default class EntityInfo extends Vue {
     )
   }
 
+  get isAmalgamationSelectionRoute (): boolean {
+    return this.$route.matched.some(route => route.path.includes(Routes.AMALGAMATION_SELECTION))
```
severinbeauvais, line 88: Does simply `this.$route.name` work?

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -41,6 +41,48 @@
...
+          text
+          color="primary"
+          :disabled="!isAllowed(AllowableActions.AMALGAMATION)"
+          @click="goToAmalgamationSelection()"
+        >
+          <v-icon medium>mdi-domain-plus</v-icon>
+          <span class="font-13 ml-1">Amalgamate</span>
+        </v-btn>
+      </template>
+      <template v-else>
```
severinbeauvais, line 59: There's a better way to make a tooltip conditional -- just disable it.

severinbeauvais: Example: https://github.com/bcgov/business-create-ui/blob/0902aa5a7e631ab0abc81567681a77364da6741f/src/components/Amalgamation/BusinessStatus.vue#L8

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -285,6 +328,14 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
     }
   }
 
+  get amalgamateTooltipText (): string {
+    if (this.isAdminFrozen) {
+      return 'This business is frozen and cannot be involved in an amalgamation.'
+    } else {
+      return 'This business has a future effective dissolution and cannot be involved in an amalgamation.'
+    }
```
severinbeauvais, line 336: I recommend an explicit check for `isPendingDissolution` instead of assuming it as the fallback case.

As a fallback, you could return null, and then you could display the tooltip only if `amalgamateTooltipText` is not falsy.

src/interfaces/api-filing-interface.ts
```diff
@@ -65,7 +65,7 @@ export interface ApiFilingIF {
       toLegalType?: CorpTypeCd
     }
 
-    amalgamation?: any // some object
+    amalgamationApplication?: any // some object
```
severinbeauvais, line 68: In a future PR, please type this :)

Also in api-task-interface.ts below.

src/routes.ts
```diff
@@ -53,6 +54,22 @@ export default [
...
+      breadcrumb: [
+        {
+          text: 'Amalgamation Selection',
+          disabled: false,
+          exact: true,
+          to: { name: Routes.AMALGAMATION_SELECTION }
+        }
+      ]
+    }
+  },
```
severinbeauvais, line 72: It makes sense to me to update the breadcrumb when they're on this panel (view) 👍

src/views/AmalgamationSelection.vue
```diff
@@ -0,0 +1,266 @@
...
+    const businessRequest = {
+      filing: {
+        header: {
+          name: FilingTypes.AMALGAMATION_APPLICATION,
+          accountId: accountId
+        },
+        business: {
+          legalType: legalType
+        },
+        amalgamation: {
```
severinbeauvais, line 207: Update this, too.

src/views/AmalgamationSelection.vue
```diff
@@ -0,0 +1,266 @@
...
+   * Must be logged in to use this.
+   * Throws an exception on error.
+   */
+  async createBusiness (businessRequest: any): Promise<any> {
+    const axios = addAxiosInterceptors(Axios.create())
+    const legalApiUrl = sessionStorage.getItem('LEGAL_API_URL')
+
+    const url = `${legalApiUrl}/businesses?draft=true`
+    return axios.post(url, businessRequest)
+  }
```
severinbeauvais, line 190: This should be in legal-services.ts (if it isn't already).

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -314,6 +365,10 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
     this.emitConfirmDissolution()
   }
 
+  goToAmalgamationSelection (): void {
+    this.$router.push({ name: Routes.AMALGAMATION_SELECTION, params: { filingId: '0' } })
```
severinbeauvais, line 369: You don't need params for your new page/view. It's used for the local filings to denote a new filing and to prevent users from going right to that route (since filingId would be undefined).

severinbeauvais: In retrospect, I think you _should_ use this (and check for filingId===0 in created) to prevent users from bypassing the entity info header button and going straight to the selection page.

src/views/AmalgamationSelection.vue
severinbeauvais: I suggest adding a window.alert (or just alert) if someone clicks on the button anyway. Currently it does nothing, which is confusing.

src/components/EntityInfo/EntityMenu.vue
JazzarKarim: This is awesome btw Sev. Was wondering if there was an easy way, I didn't know about this. Thanks!

src/components/EntityInfo.vue
severinbeauvais: Fixed, thanks.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Fixed, thanks.

src/components/EntityInfo.vue
severinbeauvais: Fixed, thanks.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Fixed, thanks.

src/interfaces/api-filing-interface.ts
severinbeauvais: Fixed, thanks.

src/views/AmalgamationSelection.vue
severinbeauvais: Fixed, thanks.

src/views/AmalgamationSelection.vue
severinbeauvais: Fixed, thanks.

src/views/AmalgamationSelection.vue
```diff
@@ -137,19 +137,32 @@
...
 @Component({})
 export default class AmalgamationSelection extends Vue {
+  @Getter(useBusinessStore) getIdentifier!: string
   @Getter(useBusinessStore) getLegalType!: string
   @Getter(useBusinessStore) isBComp!: boolean
   @Getter(useBusinessStore) isBcCompany!: boolean
   @Getter(useBusinessStore) isCcc!: boolean
   @Getter(useBusinessStore) isUlc!: boolean
 
+  filingId = 0 // id param of this selection panel route
```
severinbeauvais, line 154: You only need this in created() so I'd move it there.

src/views/AmalgamationSelection.vue
severinbeauvais: Fixed, thanks.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Fixed, thanks.

src/services/legal-services.ts
```diff
@@ -167,6 +168,19 @@ export default class LegalServices {
       })
   }
 
+  /**
+   * Creates (posts) a draft (temporary) business record.
+   * Must be logged in to use this.
+   * Throws an exception on error.
+   */
+  static async createBusiness (businessRequest: any): Promise<any> {
+    const axios = addAxiosInterceptors(Axios.create())
```
severinbeauvais, line 177: Why do you need this?

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Fixed, thanks.

src/services/legal-services.ts
JazzarKarim: You're right. I just haven't focused working on creating the draft yet. I'll have an upcoming PR that'll handle that.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -41,6 +41,34 @@
       </v-tooltip>
     </span>
 
+    <!-- Amalgamate -->
+    <span v-if="isBusiness && !isHistorical">
+      <v-tooltip
+        top
+        content-class="top-tooltip"
+        transition="fade-transition"
+        :disabled="!isAdminFrozen && !isPendingDissolution && !amalgamateTooltipText"
```
severinbeauvais, line 50: I think you only need the last condition. This would also reduce duplicate code.

src/views/AmalgamationSelection.vue
severinbeauvais: Looks good, thanks.

src/services/legal-services.ts
severinbeauvais: Fixed, thanks.

src/components/EntityInfo/EntityMenu.vue
JazzarKarim: Awesome catch!

severinbeauvais: Looks good, thanks.

severinbeauvais (approved):
OK to merge when ready.

src/interfaces/api-filing-interface.ts
JazzarKarim: I might have to create an interface in the shared components repo for this. I see that most of them are any and the ones that are not, are in the shared repo. I can take of this in the next PR!

Ref:
https://github.com/bcgov/business-filings-ui/blob/f607203812b44da61395e0e7c777c33377ec7158/src/interfaces/api-task-interface.ts#L41

PR #603, opened 2024-01-03
16967 Admin Freeze should not show on ledger

kzdev420:
*Issue #:* /bcgov/entity#16967

*Description of changes:*

### DIFF
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -143,7 +143,6 @@ export default class FilingHistoryList extends Mixins(FilingMixin) {
   is (filing: ApiFilingIF): string {
     switch (true) {
       case filing.availableOnPaperOnly: return 'paper-filing' // must come first
-      case EnumUtilities.isTypeAdminFreeze(filing): return null
       case EnumUtilities.isTypeAgmExtension(filing): return 'agm-extension'
       case EnumUtilities.isTypeAlteration(filing): return 'alteration-filing'
       case EnumUtilities.isTypeAmalgamation(filing): return 'amalgamation-filing'
--- a/src/components/Dashboard/FilingHistoryList.vue
+++ b/src/components/Dashboard/FilingHistoryList.vue
@@ -143,6 +143,7 @@ export default class FilingHistoryList extends Mixins(FilingMixin) {
   is (filing: ApiFilingIF): string {
     switch (true) {
       case filing.availableOnPaperOnly: return 'paper-filing' // must come first
+      case EnumUtilities.isTypeAdminFreeze(filing): return null
       case EnumUtilities.isTypeAgmExtension(filing): return 'agm-extension'
       case EnumUtilities.isTypeAlteration(filing): return 'alteration-filing'
       case EnumUtilities.isTypeAmalgamation(filing): return 'amalgamation-filing'
--- a/src/stores/filingHistoryListStore.ts
+++ b/src/stores/filingHistoryListStore.ts
@@ -30,9 +30,10 @@ export const useFilingHistoryListStore = defineStore('filingHistoryList', {
       return state.filings.filter(filing => {
         // safety check for required fields
         if (!filing.name || !filing.displayName || !filing.effectiveDate || !filing.submittedDate || !filing.status || !filing.displayLedger) {
-          if (filing.displayLedger)
+          if (filing.displayLedger) {
             // eslint-disable-next-line no-console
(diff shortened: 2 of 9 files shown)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-603-wd3ytfwd.web.app

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -143,6 +143,7 @@ export default class FilingHistoryList extends Mixins(FilingMixin) {
   is (filing: ApiFilingIF): string {
     switch (true) {
       case filing.availableOnPaperOnly: return 'paper-filing' // must come first
+      case EnumUtilities.isTypeAdminFreeze(filing): return null
```
severinbeauvais, line 146: I never tried this, but looking at the template code, an expansion panel is probably still created for this item.

Also this might throw off the count of items in this list.

It may be better to simply filter out the "hidden" filings in filingHistoryStore.ts::getFilings() but I haven't tried that either.

src/stores/filingHistoryListStore.ts
```diff
@@ -29,7 +29,7 @@ export const useFilingHistoryListStore = defineStore('filingHistoryList', {
       // return only the valid filings
       return state.filings.filter(filing => {
         // safety check for required fields
-        if (!filing.name || !filing.displayName || !filing.effectiveDate || !filing.submittedDate || !filing.status) {
+        if (!filing.name || !filing.displayName || !filing.effectiveDate || !filing.submittedDate || !filing.status || !filing.displayLedger) {
```
severinbeauvais, line 32: This will work but I think you should add it as a second if statement so that no console.log is displayed (line 34) if we are not to display this ledger item.

src/stores/filingHistoryListStore.ts
```diff
@@ -29,9 +29,11 @@ export const useFilingHistoryListStore = defineStore('filingHistoryList', {
...
       return state.filings.filter(filing => {
         // safety check for required fields
-        if (!filing.name || !filing.displayName || !filing.effectiveDate || !filing.submittedDate || !filing.status) {
-          // eslint-disable-next-line no-console
-          console.log('Invalid filing =', filing)
+        if (!filing.name || !filing.displayName || !filing.effectiveDate || !filing.submittedDate || !filing.status || !filing.displayLedger) {
+          if (filing.displayLedger) {
+            // eslint-disable-next-line no-console
+            console.log('Invalid filing =', filing)
+          }
```
severinbeauvais, line 36: No, don't do it this way. This is mixing the logic in line 32 with the logic in line 33. I think it would be cleaner to have two separate if statements, please.

src/components/Dashboard/FilingHistoryList.vue
severinbeauvais: Code was reverted. Resolving.

src/stores/filingHistoryListStore.ts
severinbeauvais: Looks good now. Thanks.

severinbeauvais (commented):
Looks good.

Could you add a unit test to verify that "hidden" items aren't returned by `getFilings()` but "visible" ones are?

Also, please update the app version.

Finally, please tag the 2 tickets to the next release and also update the release ticket to identify that the UI ticket depends on the API ticket.

Thanks.

tests/unit/FilingHistoryList1.spec.ts
```diff
@@ -605,6 +617,66 @@ describe('Filing History List - misc functionality', () => {
...
+        displayLedger: true,
+        displayName: 'Annual Report',
+        effectiveDate: '2019-11-20 22:17:54 GMT',
+        filingId: 111,
+        isFutureEffective: false,
+        name: FilingTypes.ANNUAL_REPORT,
+        status: FilingStatus.COMPLETED,
+        submittedDate: 'Tue, 02 July 2019 12:00:00 GMT',
+        submitter: 'Submitter 1'
+      } as any,
```
severinbeauvais, line 638: Why does this need casting to `any` but the other 2 objects below don't?

kzdev420: looks like the other 2 objects inherit the above.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-603-wd3ytfwd.web.app

severinbeauvais (approved):
Looks good but I see you need to rebase (and also update the app version again).

PR #604, opened 2024-01-03
19005 Implemented historical - amalgamation reason text

severinbeauvais:
*Issue #:* bcgov/entity#19005

*Description of changes:*
- app version = 7.0.30
- updated some code comments
- added AmalgamatedInto interface object
- added enum mixin and utility functions
- added some business store getters
- refactored getReasonText() and added case for amalgamation
- cast some parameters to pre-empt code errors
- added/updated unit tests

**Note:** This cannot be tested "for real" until a business is made historical by an amalgamation. I tested locally using mocked data as per the RFC / ticket. I also wrote some unit tests that verify all cases, so you can check there for the expected output.

### DIFF
--- a/src/services/enum-utilities.ts
+++ b/src/services/enum-utilities.ts
@@ -127,6 +127,11 @@ export default class EnumUtilities {
     return (item.name === FilingTypes.CORRECTION)
   }
 
+  /** Returns True if filing is a Dissolution. */
+  static isTypeDissolution (item: any): boolean {
+    return (item.name === FilingTypes.DISSOLUTION)
+  }
+
   /** Returns True if filing is an Amalgamation. */
   static isTypeAmalgamation (item: any): boolean {
     return (item.name === FilingTypes.AMALGAMATION_APPLICATION)
--- a/src/stores/rootStore.ts
+++ b/src/stores/rootStore.ts
@@ -1,7 +1,7 @@
 import { defineStore } from 'pinia'
 import { CorpTypeCd, EntityStatus, FilingSubTypes, FilingTypes } from '@/enums'
-import { ApiTaskIF, DissolutionConfirmationResourceIF, FilingDataIF, OfficeAddressIF, PartyIF, RootStateIF,
-  TodoListResourceIF, IsoDatePacific, StateFilingIF } from '@/interfaces'
+import { ApiTaskIF, DissolutionConfirmationResourceIF, FilingDataIF, OfficeAddressIF, PartyIF,
+  RootStateIF, TodoListResourceIF, IsoDatePacific, StateFilingIF } from '@/interfaces'
 import { DateUtilities, EnumUtilities, LegalServices } from '@/services'
 import { useBusinessStore } from './businessStore'
 import { useFilingHistoryListStore } from './filingHistoryListStore'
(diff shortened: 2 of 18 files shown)


src/views/AgmExtension.vue
```diff
@@ -454,7 +454,7 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin,
         ret = await LegalServices.createFiling(this.getIdentifier, filing, isDraft)
       }
       return ret
-    } catch (error) {
+    } catch (error: any) {
```
severinbeauvais, line 457: By default, `error` is of type Unknown, which causes code errors on lines 459 and 460.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-604-sawoy77i.web.app

tests/unit/App.spec.ts
```diff
@@ -1650,9 +1650,127 @@ describe('App as an historical business', () => {
       .returns(new Promise(resolve => resolve({
         data: {
           business: {
-            arMaxDate: '2021-12-10',
-            arMinDate: '2022-12-02',
-            dissolutionDate: '2021-12-06',
+            amalgamatedInto: {
+              amalgamationDate: '2023-12-01T08:00:00+00:00',
```
severinbeauvais, line 1654: Date comes from here.

tests/unit/App.spec.ts
```diff
@@ -1686,10 +1804,10 @@ describe('App as an historical business', () => {
             business: {},
             dissolution: {
               dissolutionType: 'voluntary',
-              dissolutionDate: '2021-12-01'
+              dissolutionDate: '2023-12-01'
             },
             header: {
-              effectiveDate: '2021-12-06T20:05:35.168290+00:00',
+              effectiveDate: '2023-12-01T08:00:00+00:00',
```
severinbeauvais, line 1810: Date comes from here.

tests/unit/App.spec.ts
```diff
@@ -1729,7 +1847,136 @@ describe('App as an historical business', () => {
...
+
+    // GET state filing
+    get.withArgs('businesses/BC1234567/filings/113526')
+      .returns(new Promise(resolve => resolve({
+        data: {
+          filing: {
+            business: {},
+            continuationOut: {},
+            header: {
+              effectiveDate: '2023-12-01T08:01:00+00:00',
```
severinbeauvais, line 1939: Date comes from here.

src/stores/rootStore.ts
```diff
@@ -225,40 +225,74 @@ export const useRootStore = defineStore('root', {
       return state.stateFiling
     },
 
-    /**
-     * A formatted concatenation of the name and the effective date of the filing.
-     * Only used when entity is historical.
-     */
+    /** The historical reason text to display in the info header. */
     getReasonText (): string {
+      const enDash = '–' // ALT + 0150
```
JazzarKarim, line 230: Why are you doing this Sev? Wouldn't it be easier just to use the character in the return statement below?

severinbeauvais: This is not a dash. It's a special [HTML entity](https://www.w3.org/wiki/Common_HTML_entities_used_for_typography) - kind of a long dash.

First, I capture it in a variable since you need to use a special keyboard command to enter it, or cut-and-paste it from somewhere.

Second, I call it out as its own variable since it's not a simple dash. And also for folks to ask me about it 😆

JazzarKarim: Ahhh, I see. OK, sweet!

severinbeauvais: FYI, I know the uxpin shows it as a regular dash but I think it's wrong. When we first implemented a historical reason, a longer dash was desired. I assume the amalgamation case is the same as the original cases.

src/services/enum-utilities.ts
```diff
@@ -127,6 +127,11 @@ export default class EnumUtilities {
     return (item.name === FilingTypes.CORRECTION)
   }
 
+  /** Returns True if filing is a Dissolution. */
+  static isTypeDissolution (item: any): boolean {
```
JazzarKarim, line 131: I don't see this being used anywhere Sev! Maybe we'll be using it later?

src/stores/rootStore.ts
JazzarKarim: OK, makes sense. Deal.

src/services/enum-utilities.ts
severinbeauvais: Yeah. I was in this code already for amalgamations and I noticed this function was missing 🤷‍♂️

PaulGarewal (approved):

PR #608, opened 2024-01-15
19291 Implemented "future effective amalgamation" alert

severinbeauvais:
*Issue #:* bcgov/entity#19291

*Description of changes:*
- app version = 7.0.34
- moved some "unknown" from script to template
- added amalgamation alert component
- added optional data object to business warnings interface
- added getter for future effective amalgamation warning
- added alert to dashboard
- fixed some route type errors
- added test suite
- updated unit tests

**As the Legal API change for this has not yet been implemented, see ticket for local screenshots.**

### DIFF
--- a/src/components/Dashboard/Alerts/Amalgamation.vue
+++ b/src/components/Dashboard/Alerts/Amalgamation.vue
@@ -1,0 +1,127 @@
+<template>
+  <v-expansion-panels
+    id="amalgamation-container"
+    v-model="panel"
+  >
+    <v-expansion-panel class="mb-6">
+      <v-expansion-panel-header
+        hide-actions
+        class="d-flex justify-space-between px-6 py-5"
+      >
+        <h3>
+          <v-icon
+            left
+            color="orange darken-2"
+          >
+            mdi-alert
+          </v-icon>
+          <span>This corporation is part of an amalgamation and is scheduled to become
+            historical on {{ amalgamationDate || '[unknown]' }}.</span>
+        </h3>
+        <v-btn
+          text
+          color="primary"
+          class="details-btn my-n1"
+          @click.stop="togglePanel()"
+        >
+          <span color="primary">{{ panel === 0 ? "Hide Details" : "View Details" }}</span>
+          <v-icon
+            right
+            color="primary"
+          >
+            {{ panel === 0 ? "mdi-chevron-up" : "mdi-chevron-down" }}
+          </v-icon>
+        </v-btn>
+      </v-expansion-panel-header>
+
+      <v-expansion-panel-content>
+        <p class="mb-0">
+          If you have any questions, please contact BC Registries staff:
+        </p>
+        <ContactInfo class="pt-5" />
+      </v-expansion-panel-content>
+    </v-expansion-panel>
+  </v-expansion-panels>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+import { Getter } from 'pinia-class'
+import { ContactInfo } from '@/components/common'
+import { useBusinessStore } from '@/stores'
+import { ApiDateTimeUtc, BusinessWarningIF } from '@/interfaces'
+import { DateUtilities } from '@/services'
+
+@Component({
+  components: { ContactInfo }
+})
+export default class Amalgamation extends Vue {
+  @Getter(useBusinessStore) getBusinessWarnings!: BusinessWarningIF[]
+
+  panel = 1
+
+  togglePanel (): void {
+    this.panel = (this.panel === 1 ? 0 : 1)
+  }
+
+  get amalgamationDate (): string {
+    const warning = this.getBusinessWarnings.find(item =>
+      item.warningType?.includes('FUTURE_EFFECTIVE_AMALGAMATION')
+    )
+    const date = warning?.data?.amalgamationDate as ApiDateTimeUtc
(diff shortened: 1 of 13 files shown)


src/components/Dashboard/Alerts/Amalgamation.vue
```diff
@@ -0,0 +1,127 @@
...
+  togglePanel (): void {
+    this.panel = (this.panel === 1 ? 0 : 1)
+  }
+
+  get amalgamationDate (): string {
+    const warning = this.getBusinessWarnings.find(item =>
+      item.warningType?.includes('FUTURE_EFFECTIVE_AMALGAMATION')
+    )
+    const date = warning?.data?.amalgamationDate as ApiDateTimeUtc
+    return DateUtilities.apiToPacificDate(date, true)
```
severinbeauvais, line 72: This safely returns null if `date` is invalid/null/undefined.

src/components/Dashboard/FilingHistoryList/LimitedRestorationExtensionFiling.vue
```diff
@@ -7,10 +7,10 @@
 
     <p>
       The period of restoration was successfuly extended and is active
-      <strong>until {{ expiryDateFriendly }}</strong>. At the end of the extended limited
-      restoration period, the company will be automatically dissolved. If you require assistance
-      to extend a limited restoration/reinstatement or wish to convert your restoration from a
-      limited period to a full restoration, please contact BC Registries staff:
+      <strong>until {{ expiryDateFriendly || '[Unknown]' }}</strong>. At the end of the extended
```
severinbeauvais, line 10: I think it's better for the script to handle fallback strings. That way, the script can handle the "computations" and the script can focus on language, etc.

src/interfaces/business-state-interfaces.ts
```diff
@@ -29,6 +29,7 @@ export interface BusinessWarningIF {
   filing?: string // not used
   message: string
   warningType: string // FUTURE: use an enum
+  data?: any // optional extra properties (eg, amalgamationDate)
```
severinbeauvais, line 32: As per bcgov/entity#19292.

src/views/Dashboard.vue
```diff
@@ -385,7 +395,7 @@ export default {
 
   methods: {
     goToStandaloneDirectors () {
-      this.$router.push({ name: Routes.STANDALONE_DIRECTORS, params: { filingId: 0 } }) // 0 means "new COD filing"
+      this.$router.push({ name: Routes.STANDALONE_DIRECTORS, params: { filingId: '0' } }) // 0 means "new COD filing"
```
severinbeauvais, line 398: Parameter is now expected to be a string.

tests/unit/DateTooltip.spec.ts
```diff
@@ -15,7 +15,7 @@ describe('Date Tooltip', () => {
 
     // verify content
     expect(wrapper.find('span').text()).toBe('[unknown]')
-    expect(vm.dateTimeString).toBe('[unknown]')
+    expect(vm.dateTimeString).toBeNull()
```
severinbeauvais, line 18: This is checking the value of the computed, not the rendered string.

src/components/common/Certify.vue
```diff
@@ -38,7 +38,7 @@
           </template>
         </v-checkbox>
         <p class="certify-clause signature-date">
-          <strong>Date:</strong> {{ formattedCurrentDate }}
+          <strong>Date:</strong> {{ formattedCurrentDate || '[unknown]' }}
```
leodube-aot, line 41: Any particular reason why "[unknown]" is lowercase here but is capitalized as "[Unknown]" in LimitedRestorationExtensionFiling.vue?

src/views/Dashboard.vue
ketaki-deodhar: will this affect any other applications and do we need to change those as well?  I see you have already changed 2 here :)

jamespaologarcia (approved):
LGTM

src/components/common/Certify.vue
severinbeauvais: I'll update that other component.

We haven't had much UX guidance on error handling, but I've gravitated towards showing unknown as `[unknown]` (which looks like a token) instead of `Unknown` (which looks like a noun). Let me know your thoughts on that.

src/views/Dashboard.vue
severinbeauvais: The old code worked fine and didn't show any errors/warnings in VS Code. I think the new code works, too, but VS Code complained about it, so I fixed it.

Since you asked, I checked Create UI and Edit UI, and they either don't pass URL parameters this way or they are already strings, so I think there's nothing more to update at this time.

The take-away here is that, as our tools improve (eg, Lint, language services), more code issues will be flagged, and we should try to fix those to prevent issues. Thanks for asking 👋

ketaki-deodhar (approved):
LGTM!

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-608-sxp4zp3h.web.app

src/components/common/Certify.vue
leodube-aot: I'm personally good with either if it's consistent!

leodube-aot (approved):

src/components/common/Certify.vue
severinbeauvais: It is now!

PR #610, opened 2024-01-16
Added amalgamation FF check

JazzarKarim:
*Description of changes:*
- Don't display the amalgamate button if the FF doesn't support that entity type

### DIFF
--- a/src/components/EntityInfo/EntityMenu.vue
+++ b/src/components/EntityInfo/EntityMenu.vue
@@ -219,6 +219,7 @@
               >
                 <template #activator="{ on }">
                   <div
+                    v-if="enableAmalgamation"
                     v-on="on"
                   >
                     <v-list-item
@@ -293,6 +294,10 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
 
   get enableAgmLocationChg (): boolean {
     return !!GetFeatureFlag('supported-agm-location-chg-entities').includes(this.getLegalType)
+  }
+
+  get enableAmalgamation (): boolean {
+    return !!GetFeatureFlag('supported-amalgamation-entities').includes(this.getLegalType)
   }
 
   /** The tooltip text for AGM Extension list item. Text is different if action item is disabled. */
--- a/tests/unit/EntityMenu.spec.ts
+++ b/tests/unit/EntityMenu.spec.ts
@@ -292,7 +292,7 @@ describe('Entity Menu - View and Change Business Information click tests', () =>
   })
 })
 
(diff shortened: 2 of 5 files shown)


JazzarKarim:
@severinbeauvais Sev, I just saw your PR. In my case I'm not showing the button altogether based of the FF. I'm guessing we don't want to do that? Shall we proceed with your PR?

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -219,6 +219,7 @@
               >
                 <template #activator="{ on }">
                   <div
+                    v-if="enableAmalgamation"
```
severinbeauvais, line 222: Are you sure we don't want to show it at all right now?

tests/unit/EntityMenu.spec.ts
```diff
@@ -293,6 +293,21 @@ describe('Entity Menu - View and Change Business Information click tests', () =>
 })
 
 describe('Entity Menu - Amagamate button tests', () => {
```
severinbeauvais, line 295: typo

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Ask Mihai.

severinbeauvais:
Can this wait until Thursday's deployment, then? (Not a hotfix?) (Will need to be pushed to Test today, along with its predecessor PRs.)

JazzarKarim:
> Can this wait until Thursday's deployment, then? (Not a hotfix?) (Will need to be pushed to Test today, along with its predecessor PRs.)

It depends on how urgent Mihai wants it. I guess it can wait.

src/components/EntityInfo/EntityMenu.vue
JazzarKarim: Looks like we're going with remove Sev:
remove

severinbeauvais: OK.

I would tease the users. (They could learn what's coming and where it will be.) But not my call.

severinbeauvais (approved):

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Funny that comment came from Yui. In the past, he's advocated for "teaching" users where things are / will be.

JazzarKarim: > I would tease the users. (They could learn what's coming and where it will be.) But not my call.

I agree.

PR #625, opened 2024-02-21
19921 Update LegalObligation logic

severinbeauvais:
*Issue #:* bcgov/entity#19921

*Description of changes:*
- app version = 7.0.46
- don't show LegalObligation if business is not Active
- show LegalObligation for new business
- updated unit tests

### DIFF
--- a/src/components/Dashboard/LegalObligation.vue
+++ b/src/components/Dashboard/LegalObligation.vue
@@ -81,21 +81,28 @@
 </template>
 
 <script lang="ts">
-import { Component, Mixins } from 'vue-property-decorator'
+import { Component, Mixins, Watch } from 'vue-property-decorator'
 import { Getter } from 'pinia-class'
 import ResourceLookupMixin from '@/mixins/resource-lookup-mixin'
-import { useRootStore } from '@/stores'
+import { useBusinessStore, useRootStore } from '@/stores'
 
 @Component({})
 export default class LegalObligation extends Mixins(ResourceLookupMixin) {
+  @Getter(useBusinessStore) isActive!: boolean
   @Getter(useRootStore) isBusinessWithNoMaintenanceFilings!: boolean
 
   readMoreFlag = false
-  showLegalObligation = true
+  showLegalObligation = null as boolean
 
   /** Whether this entity is a business (and not a temporary registration). */
   get isBusiness (): boolean {
     return !!sessionStorage.getItem('BUSINESS_ID')
+  }
+
+  /** Sets initial value of showLegalObligation (and any changes). */
+  @Watch('isActive', { immediate: true })
+  onIsActiveChanged () {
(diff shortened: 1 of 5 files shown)


src/components/Dashboard/LegalObligation.vue
```diff
@@ -81,22 +81,29 @@
...
+import { useBusinessStore, useRootStore } from '@/stores'
 
 @Component({})
 export default class LegalObligation extends Mixins(ResourceLookupMixin) {
+  @Getter(useBusinessStore) isActive!: boolean
   @Getter(useRootStore) isBusinessWithNoMaintenanceFilings!: boolean
 
   readMoreFlag = false
-  showLegalObligation = true
+  showLegalObligation = null as boolean
```
severinbeauvais, line 95: I did this so that it's neither True nor False (for possible debugging), until it's set in the watcher below.

Note that I couldn't just write `showLegalObligation = this.isActive` here because that's evaluated before isActive has its actual value.

I could have set showLegalObligation in a `mounted()` method but I think the watcher design is more accurate as it keeps the value synced (instead of setting it just once).

src/components/Dashboard/LegalObligation.vue
```diff
@@ -81,22 +81,29 @@
...
   /** Whether this entity is a business (and not a temporary registration). */
   get isBusiness (): boolean {
     return !!sessionStorage.getItem('BUSINESS_ID')
   }
+
+  /** Sets initial value of showLegalObligation (and any changes). */
+  @Watch('isActive', { immediate: true })
+  onIsActiveChanged () {
+    this.showLegalObligation = this.isActive
+  }
```
severinbeauvais, line 106: Theoretically, this should fire just once: initially (aka immediately), right after this component is mounted.

If it ever happens that `isActive` changes value then this will fire again and may show/hide the Legal Obligation panel accordingly (as per the other conditions on line 3) and I think that's OK.

src/stores/rootStore.ts
```diff
@@ -167,15 +167,16 @@ export const useRootStore = defineStore('root', {
...
      **/
     isBusinessWithNoMaintenanceFilings (state: RootStateIF): boolean {
       const filingHistoryListStore = useFilingHistoryListStore()
       return (
         // no todo items
         (state.tasks.length === 0) &&
-        // only the IA or Registraion filing history item
+        // only the Amalgamation or IA or Registration filing history item
         (filingHistoryListStore.getFilings.length === 1) && (
+          EnumUtilities.isTypeAmalgamation(filingHistoryListStore.getFilings[0]) ||
```
severinbeauvais, line 179: This is to SHOW the Legal Obligations panel for a new TED.

Before:
image

After:
image

PaulGarewal (approved):
LGTM! 👍

ketaki-deodhar (approved):
Works well!

PR #627, opened 2024-02-23
19906 Updated breadcrumb URLs + refactored authentication store

severinbeauvais:
*Issue #:* bcgov/entity#19906

*Description of changes:*
- app version = 7.0.47
- deleted local Breadcrumb IF and used shared version instead
- misc comment cleanup
- attached global Vuex store to Vue instance
- added sleep to display startup error before alert
- deleted obsolete interfaces
- updated some interfaces
- updated My Business Registry breadcrumb
- updated Staff dashboard breadcrumb
- refactored authentication store to get date from global Vuex store
- added missing state init


src/App.vue
```diff
@@ -228,31 +228,28 @@ export default {
...
+    ]),
+
+    ...mapState(useRootStore, [
+      'getKeycloakRoles',
+      'isAppFiling',
+      'isAppTask',
+      'isRoleStaff',
+      'showFetchingDataSpinner',
+      'showStartingAmalgamationSpinner'
+    ]),
```
severinbeauvais, line 252: Sorted stores and lined up brackets.

src/components/Dashboard/FilingHistoryList.vue
```diff
@@ -110,7 +110,6 @@ export default class FilingHistoryList extends Mixins(FilingMixin) {
   @Getter(useBusinessStore) isBenBcCccUlc!: boolean
   @Getter(useBusinessStore) isCoop!: boolean
   @Getter(useBusinessStore) isFirm!: boolean
-  @Getter(useRootStore) isRoleStaff!: boolean
```
severinbeauvais, line 113: Not used in this component.

src/components/Dashboard/FilingHistoryList/HeaderActions.vue
```diff
@@ -92,6 +92,7 @@ export default class HeaderActions extends Mixins(AllowableActionsMixin) {
   @Prop({ required: true }) readonly index!: number
 
   @Getter(useBusinessStore) isBenBcCccUlc!: boolean
+  // @Getter(useAuthenticationStore) isRoleStaff!: boolean
```
severinbeauvais, line 95: Used in this component but inherited from a mixin. Still, I like to see the declaration in the component.

src/interfaces/authentication-state-interface.ts
severinbeauvais, line 1: Not used.

src/main.ts
```diff
@@ -31,8 +31,11 @@ Vue.use(Vue2Filters) // needed by SbcFeeSummary
 const store = getVuexStore()
 const pinia = getPiniaStore()
 
-// Needed to fix "getActivePinia was called with no active Pinia" error
-// Type assertion to turn off type checking
+// attach global Vuex store to Vue instance
+Vue.prototype.$store = store
```
severinbeauvais, line 35: This is key, since you get the global Vuex store only the first time you call getVuexStore() and we need to store its handle somewhere.

src/interfaces/authentication-state-interface.ts
severinbeauvais, line 1: Not used.

src/stores/authenticationStore.ts
severinbeauvais, line 1: This Pinia now (correctly) accesses state from the global Vuex store instance (which is managed by the imported sbc-common-components module that handles login and accounts).

severinbeauvais:
As a regular user, the BC Registries Dashboard URL (see bottom left):
image

As a regular user, the My Business Registry URL:
image

As a staff user, the Staff Dashboard URL:
image

src/resources/BreadcrumbResources.ts
```diff
@@ -1,34 +1,43 @@
...
+import { useAuthenticationStore } from '@/stores'
+
+setActivePinia(createPinia())
+const authenticationStore = useAuthenticationStore()
 
 /** Returns URL param string with Account ID if present, else empty string. */
 function getParams (): string {
-  const accountId = JSON.parse(sessionStorage.getItem('CURRENT_ACCOUNT'))?.id
+  const accountId = authenticationStore.getCurrentAccount?.id || 0
   return accountId ? `?accountid=${accountId}` : ''
```
JazzarKarim, line 12: Wonder if it would be a good idea to move these lines to the `getRegistryDashboardBreadcrumb` function and get rid of `getParams`.

JazzarKarim (approved):
LGTM!

seeker25 (approved):
High level LGTM

ketaki-deodhar (approved):
Looks good Sev!

src/resources/BreadcrumbResources.ts
severinbeauvais: You're right,  `getParams()` is only used once here so normally I'd combine the code.

However, in Edit UI, `getParams()` is used twice, and in Create UI, it's used three times. For the sake of consistency, how about if we leave it like this for now?

PaulGarewal (approved):

src/resources/BreadcrumbResources.ts
JazzarKarim: Deal, OK with me Sev. Resolving.

PR #648, opened 2024-04-15
Implemented easy legal name fix + unit tests

JazzarKarim:
*Issue #:* /bcgov/entity#20779

*Description of changes:*
- Implemented the easy legal name fix
  - If FF is off: Retain previous behavior
  - If FF is on: Use the alternate name array
- Updated unit tests
- Misc. unit test fixes


src/stores/businessStore.ts
```diff
@@ -91,9 +91,19 @@ export const useBusinessStore = defineStore('business', {
       return state.businessInfo.lastDirectorChangeDate
     },
 
-    /** The legal name. */
+    /** The legal name or alternate name if is firm. */
     getLegalName (state: BusinessStateIF): string {
-      return state.businessInfo.legalName
+      if (!GetFeatureFlag('enable-legal-name-fix')) {
```
JazzarKarim, line 96: The idea here is the following:
- If FF is disabled: keep whatever we had before. Nothing changes.
- If FF is enabled:
  - We check if firm and Alternate names array. If true, this'll return name. 
  - Otherwise, just return the legal name. BENs for example.

tests/unit/EntityMenu.spec.ts
```diff
@@ -214,17 +214,17 @@ describe('Entity Menu - View and Change Business Information click tests', () =>
 
     // set configurations
     configurationStore.setConfiguration(configuration)
+
```
JazzarKarim, line 217: The following unit test changes are not related to this ticket whatsoever.

6 unit tests were failing from previous PR merges. I don't know why the GitHub CI didn't catch that or maybe we missed that the testing coverage was failing. Anyway, I fixed these tests.

src/stores/businessStore.ts
```diff
@@ -91,9 +91,23 @@ export const useBusinessStore = defineStore('business', {
...
+        return state.businessInfo.legalName
+      }
+      if (this.isFirm) {
+        return this.getAlternateName || 'Unknown'
```
JazzarKarim, line 100: The unknown here is just a failsafe value. I confirmed with Argus that firms should and will always have the alternate names array. 

Line 102 is same thing. We should always return legal name in case not firm. That also should always be present.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "7.1.11",
+  "version": "7.2.0",
```
JazzarKarim, line 3: Was thinking this is a minor upgrade since it partially includes the legal name fix. I'm open to your thoughts if you think otherwise (7.1.12). Let me know!

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-648-7xrf7iic.web.app

package.json
severinbeauvais: Minor upgrade sounds good.

src/enums/index.ts
```diff
@@ -22,5 +22,5 @@ export * from './affiliationInvitation'
 
 // external enums
 export { CorpTypeCd, CorpClass } from '@bcrs-shared-components/corp-type-module'
-export { AccountTypes, AmalgamationTypes, FilingCodes, FilingNames, FilingTypes, StaffPaymentOptions }
+export { AccountTypes, AmalgamationTypes, FilingCodes, FilingNames, FilingTypes, NameTypes, StaffPaymentOptions }
```
severinbeauvais, line 25: Actually, I prefer to see these imported from the shared repo in the components that need them. I'm trying to get past the issue of types not being recognized as the same because they were imported differently. LMK your thoughts on this.

severinbeauvais: Don't spend too much time on this in case this all changes later.

JazzarKarim: I agree Sev. Not only that, but I've noticed A LOT of inconsistency throughout the code. Some files are using these enums from `@/enum` while others are using them from the shared components directly (`@bcrs-shared-components/enums`).

I've made this change. This should also made the code consistent. They're all being used from the shared components now whenever a file needs them.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-648-7xrf7iic.web.app

leodube-aot (approved):
LGTM Karim!

tests/unit/TodoList1.spec.ts
```diff
@@ -1849,7 +1850,7 @@ describe('TodoList - UI - Amalgamation Applications', () => {
 describe('TodoList - Click Tests - Coops', () => {
   const { assign } = window.location
 
-  beforeAll(() => {
+  beforeEach(() => {
```
severinbeauvais, line 1853: Does this really have to be set before each test?

severinbeauvais (approved):

tests/unit/TodoList1.spec.ts
JazzarKarim: Yup, some of values were missing, that's why the tests were failing.

src/interfaces/business-state-interfaces.ts
```diff
@@ -17,6 +17,16 @@ export interface AllowedActionsIF {
...
 
+/** The Alternate Name object. */
+export interface AlternateNameIF {
+  entityType: CorpTypeCd
+  identifier: string
+  name: string
+  nameRegisteredDate: ApiDateTimeUtc
+  nameStartDate: IsoDatePacific
+  nameType: NameTypes
+}
```
severinbeauvais, line 28: When you do Edit UI, how about moving this to the shared interfaces (and coming back and importing here and in Filings UI)?

PR #649, opened 2024-04-19
20932 + 20946 legal name easy fix patches

JazzarKarim:
*Issue #:* /bcgov/entity#20932 & /bcgov/entity#20946

*Description of changes:*
- Updated to use the new shared interface with updated fields
- Fixed getting "Unknown" in draft IAs when FF is on

### DIFF
--- a/src/stores/businessStore.ts
+++ b/src/stores/businessStore.ts
@@ -97,9 +97,9 @@ export const useBusinessStore = defineStore('business', {
         return state.businessInfo.legalName
       }
       if (this.isFirm) {
-        return this.getAlternateName || 'Unknown'
+        return this.getAlternateName
       } else {
-        return state.businessInfo.legalName || 'Unknown'
+        return state.businessInfo.legalName
       }
     },
 
--- a/tests/unit/EntityHeader.spec.ts
+++ b/tests/unit/EntityHeader.spec.ts
@@ -571,7 +571,7 @@ describe('Entity Header - Alternate Name', () => {
     })
     await Vue.nextTick()
     // verify displayed text
-    expect(wrapper.find('#entity-legal-name').text()).toBe('Unknown')
+    expect(wrapper.find('#entity-legal-name').text()).toBe('Unknown Name')
   })
 
   it('displays legal name if firm and legal name fix FF is false', async () => {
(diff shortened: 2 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-649-1eqc57r1.web.app

argush3 (approved):

tests/unit/EntityHeader.spec.ts
```diff
@@ -571,7 +571,7 @@ describe('Entity Header - Alternate Name', () => {
     })
     await Vue.nextTick()
     // verify displayed text
-    expect(wrapper.find('#entity-legal-name').text()).toBe('Unknown')
+    expect(wrapper.find('#entity-legal-name').text()).toBe('Unknown Name')
```
severinbeauvais, line 574: Where does this value come from?

severinbeauvais: Oh, it's the fallback value in the template for EntityHeader.vue.

src/stores/businessStore.ts
```diff
@@ -97,9 +97,9 @@ export const useBusinessStore = defineStore('business', {
         return state.businessInfo.legalName
       }
       if (this.isFirm) {
-        return this.getAlternateName || 'Unknown'
+        return this.getAlternateName
       } else {
-        return state.businessInfo.legalName || 'Unknown'
+        return state.businessInfo.legalName
       }
```
severinbeauvais, line 103: So we have a situation here where this sometimes returns null/empty, and it's handled differently throughout the code.

I'm kinda thing that -- in the future -- we should get rid of this store getter and just fetch whatever name we need, and fallback however we need, right in the components that currently use this. It might seem like a step backwards, but it would remove any assumption about how this is used. Thoughts?

severinbeauvais (approved):
GTG

src/stores/businessStore.ts
JazzarKarim: Ah, you're talking about how the code handles that scenario by having `Unknown Name` right? Yup, makes a lot of sense to me.

OK to deal with that later on Sev?

severinbeauvais: Yes, it's a problem for later.

PR #651, opened 2024-04-24
20534-remove-details-field-for-users

jamespaologarcia:
*Issue #:* /bcgov/entity#20534

Updated the details field for continue out so it's required for Staff filings but not for regular users. 

Set the default value to null

### DIFF
--- a/src/views/ContinuationOut.vue
+++ b/src/views/ContinuationOut.vue
@@ -886,7 +886,6 @@ export default class ContinuationOut extends Mixins(CommonMixin, DateMixin,
 
   @Watch('certifyFormValid')
   @Watch('courtOrderValid')
-  @Watch('detailCommentValid')
   @Watch('documentDeliveryValid')
   onHaveChanges (): void {
     this.haveChanges = true
--- a/src/views/ContinuationOut.vue
+++ b/src/views/ContinuationOut.vue
@@ -61,50 +61,6 @@
               </h1>
             </header>
 
-            <!-- Ledger Detail -->
-            <section v-if="isRoleStaff">
-              <header>
-                <h2>Ledger Detail</h2>
-                <p class="grey-text">
-                  Enter a detail that will appear on the ledger for this entity.
-                </p>
-              </header>
-              <div
-                id="detail-comment-section"
-                :class="{ 'invalid-section': !detailCommentValid && showErrors }"
-              >
-                <v-card
-                  flat
-                  class="py-8 px-5"
-                >
-                  <v-row no-gutters>
-                    <v-col
-                      cols="12"
-                      sm="3"
-                      class="pr-4"
-                    >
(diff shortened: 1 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-651-qfwvozrp.web.app

src/views/ContinuationOut.vue
```diff
@@ -97,6 +97,8 @@
                         v-model="detailComment"
                         placeholder="Add a Detail that will appear on the ledger for this entity."
                         :maxLength="maxDetailCommentLength"
+                        :isRoleStaff="isRoleStaff"
+                        :isContinuationOutDetailComment="true"
```
severinbeauvais, line 101: Why do you need these extra properties, if this entire section (code block) is only displayed for staff?

jamespaologarcia: DetailComment component is used elsewhere. I thought since it only applies to ContOut, then I need to specify it. 

I'll update this

src/views/ContinuationOut.vue
```diff
@@ -62,7 +62,7 @@
             </header>
 
             <!-- Ledger Detail -->
-            <section>
+            <section v-if="isRoleStaff">
```
severinbeauvais, line 65: Per the clarified requirements from Naveen, this component (and all its associated variables) should be removed.

Also, this is a staff-only filing, so this code did nothing.

severinbeauvais (changes requested):
Do not merge as is. See comments for details.

src/views/ContinuationOut.vue
severinbeauvais: In the future, if this kind of thing is needed, it's better for common components to know as little as possible about the world around them (which them more portable).

Specifically, it would be better if this component doesn't know what type of filing it's working on. What you could do instead is create a generic prop that, in this case, would tell the component to override the rules.

Anyway, I think none of this is needed in this PR so FYI only :)

severinbeauvais (changes requested):
The updated code change looks good, but now you have some failing unit tests.

If you go to the [testing-coverage details](https://github.com/bcgov/business-filings-ui/actions/runs/8823694041/job/24224691776?pr=651), you'll see that the "Test with Vitest" step failed, not the "Upload coverage" step (which fails more often than not recently).

severinbeauvais:
As a general rule,  you should run `npm run lint` and `npm run test:unit` locally before you ask for a PR to be reviewed.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-651-qfwvozrp.web.app

severinbeauvais (approved):
Approved. Please wait for Ketaki to look it over also.

ketaki-deodhar:
> As a general rule, you should run `npm run lint` and `npm run test:unit` locally before you ask for a PR to be reviewed.

I second that :)

ketaki-deodhar (approved):
Looks good

PR #653, opened 2024-05-06
Renamed script to prevent confusion

severinbeauvais:
*Issue #:* None

*Description of changes:*
- renamed script "serve" to "preview" to prevent confusion with other projects
- removed port specification so that it runs on a different port than "npm run dev" which should also prevent confusion

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "serve": "vite preview --port 8080",
+    "preview": "vite preview",
     "lint": "eslint . --ext .js,.ts,.vue",
     "test:unit": "vitest run",
     "test:coverage": "vitest run --coverage --silent"


JazzarKarim (approved):

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "serve": "vite preview --port 8080",
+    "preview": "vite preview",
```
severinbeauvais, line 10: I wasted an hour trying to figure out why I couldn't load this UI using `npm run serve`, which serves up a previous local build (npm run build) if there is one.

`npm run serve` is still needed for Namerequest UI since it still uses Vue CLI instead of Vite. This can lead to develop confusion when they work on multiple repos at once.

seeker25 (approved):

package.json
severinbeauvais: Credit where credit is due: this solution, which I love, was Karim's idea 😀

JazzarKarim: Thank you Sev! ❤️

PR #659, opened 2024-05-31
21269 - Added new alert for involuntary dissolution

JazzarKarim:
*Issue #:* /bcgov/entity#21269

*Description of changes:*
- Added the new alert for businesses that are in the process of being involuntary dissolved
- Misc. fixes
- Unit tests

NOTE: There's no way to test this right now as the BE is still being worked on: https://github.com/bcgov/lear/pull/2711
So, what I did was is I manually added this block to the business being tested:
```json
businessInfo.warnings = [{
  "code": "DISSOLUTION_IN_PROGRESS",
    "data": {
      "missingARs": 5,
      "missingTransitionFilings": false,
      "targetDissolutionDate": "2024-07-30",
      "warningsSent": 2
    },
    "message": "Business is in the process of involuntary dissolution.",
    "warningType": "INVOLUNTARY_DISSOLUTION"
}]
```

Result:


src/components/Dashboard/Alerts/Amalgamation.vue
```diff
@@ -57,7 +58,7 @@ export default class Amalgamation extends Vue {
 
   get amalgamationDate (): string {
     const warning = this.getBusinessWarnings.find(item =>
-      item.warningType?.includes('FUTURE_EFFECTIVE_AMALGAMATION')
+      item.warningType?.includes(WarningTypes.FUTURE_EFFECTIVE_AMALGAMATION)
```
JazzarKarim, line 61: Changed all warning types to use an enum now. Please see later comment below.

src/components/Dashboard/Alerts/InDissolution.vue
```diff
@@ -0,0 +1,130 @@
...
+  @Prop({ required: true }) readonly showPanel!: boolean
+
+  @Getter(useBusinessStore) getBusinessWarnings!: BusinessWarningIF[]
+  @Getter(useRootStore) getCurrentDate!: string
+
+  /** Return the number of days left for the business until it is dissolved. */
+  get daysLeft (): number {
+    const warning = this.getBusinessWarnings.find(item =>
+      item.warningType?.includes(WarningTypes.INVOLUNTARY_DISSOLUTION)
+    )
```
JazzarKarim, line 69: This is the cleanest way that came to mind. I also followed a similar path as other places in the code.

src/components/Dashboard/Alerts/InDissolution.vue
```diff
@@ -0,0 +1,130 @@
...
+    const targetDissolutionDate = warning?.data?.targetDissolutionDate
+
+    return DateUtilities.daysBetweenTwoDates(new Date(this.getCurrentDate), new Date(targetDissolutionDate)) || -1
```
JazzarKarim, line 72: The -1 here is just a fallback. Theoretically, we should never get that.

src/interfaces/business-state-interfaces.ts
```diff
@@ -28,7 +28,7 @@ export interface BusinessWarningIF {
   code: string // FUTURE: use an enum
   filing?: string // not used
   message: string
-  warningType: string // FUTURE: use an enum
+  warningType: WarningTypes
```
JazzarKarim, line 31: As the number of warning types is continuing to grow, I thought this might be the time to finally do this 😁

tests/unit/Alerts.spec.ts
```diff
@@ -49,6 +51,10 @@ describe('Dashboard - UI', () => {
           get (): boolean { return this.$data._isFrozenInformationAlert },
           set (val: boolean) { this.$data._isFrozenInformationAlert = val }
         },
+        isInDissolutionAlert: {
```
JazzarKarim, line 54: @severinbeauvais Just wanted to say, the test file here is absolutely gorgeous. It was so easy to add a new test case for the new alert 😃

src/components/Dashboard/Alerts/InDissolution.vue
severinbeauvais: So if the days between the two dates is 0 (ie, false) then you use -1 instead? I think that would look weird.

Perhaps you want to do something specific if the result is NaN (ie, display Unknown)?

src/components/Dashboard/Alerts.vue
```diff
@@ -51,6 +54,7 @@ export default class Alerts extends Vue {
   /** The list of alerts. */
   get alerts (): Array<any> {
     const alerts = []
+    if (this.isInDissolutionAlert) alerts.push(InDissolution) // needs to be shown first (high priority)
```
JazzarKarim, line 57: From the designs, this alert will precede all others if it exists. We want to show it first.

tests/unit/Alerts.spec.ts
severinbeauvais: Thanks. I just noticed that this test suite can be optimized a bit by doing the setup and teardown just once ("all") instead of for each test case.

Before:
image

After:
image

Not a huge savings but seems like better code since it's not wasting cycles. Up to you if you want to update this.

src/components/Dashboard/Alerts/InDissolution.vue
JazzarKarim: Good point. I was actually thinking the fallback would happen if we were unable to grab both dates for whatever reason.

I'm thinking, I'll remove that fallback altogether. The `daysBetweenTwoDates` function already has safety checks for all the cases.

src/components/Dashboard/Alerts/InDissolution.vue
```diff
@@ -0,0 +1,130 @@
...
+  @Getter(useRootStore) getCurrentDate!: string
+
+  /** Return the number of days left for the business until it is dissolved. */
+  get daysLeft (): number {
+    const warning = this.getBusinessWarnings.find(item =>
+      item.warningType?.includes(WarningTypes.INVOLUNTARY_DISSOLUTION)
+    )
+    const targetDissolutionDate = warning?.data?.targetDissolutionDate
+
+    return DateUtilities.daysBetweenTwoDates(new Date(this.getCurrentDate), new Date(targetDissolutionDate))
```
severinbeauvais, line 72: The result might still be NaN (which will display as "NaN"). You might want to provide a fallback.

- the result could be > 0 (most likely case)
- the result could legitimately be 0 ?
- the result could be < 0 (is that an error for you?)
- the result could be NaN

tests/unit/Alerts.spec.ts
JazzarKarim: Done. Thanks for this!

src/components/Dashboard/Alerts/InDissolution.vue
JazzarKarim: You're absolutely right.

The idea is, anything less than 1 should be wrong. The moment the days hit 0, the business will become historical due to involuntary dissolution and that alert will go away.

I added a fallback for that in latest commit. Let me know what you think.

src/components/Dashboard/Alerts/InDissolution.vue
severinbeauvais: This conversation is continued below. Resolving this one.

src/components/Dashboard/Alerts/InDissolution.vue
```diff
@@ -0,0 +1,132 @@
...
+
+  /** Return the number of days left for the business until it is dissolved. */
+  get daysLeft (): string {
+    const warning = this.getBusinessWarnings.find(item =>
+      item.warningType?.includes(WarningTypes.INVOLUNTARY_DISSOLUTION)
+    )
+    const targetDissolutionDate = warning?.data?.targetDissolutionDate
+    const daysDifference = DateUtilities.daysBetweenTwoDates(new Date(this.getCurrentDate), new Date(targetDissolutionDate))
+
+    if (!Number.isNaN(daysDifference) && daysDifference > 0) return String(daysDifference)
```
severinbeauvais, line 73: I think you can simplify this. Hint:

image

Also, I see a lint warning on line 71.

src/components/Dashboard/Alerts/InDissolution.vue
severinbeauvais: Looks OK. So the result can never be 0  (ie, today)?

I'm OK to resolve this conversation. (I added another new comment to discuss separately.)

src/components/Dashboard/Alerts/InDissolution.vue
JazzarKarim: 🚀 🚀 Thanks!

src/components/Dashboard/Alerts/InDissolution.vue
JazzarKarim: I added 0 (today) since it might depend on the exact time of when the business becomes historical.

severinbeauvais: LGTM

severinbeauvais (approved):

AimeeGao (approved):
Looks good! Learned a lot from the comments and discussions!

PR #661, opened 2024-06-05
21439 - Fixed Unknown Name when registration is in draft

JazzarKarim:
*Issue #:* /bcgov/entity#21439

*Description of changes:*
When the registration is draft, the code was checking for the name from alternate names but that doesn't exist at that point. When that is the case, we need to use the legal name as before. So, I simply added that check.

Before:
before sp

After:
after sp

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-filings-ui",
-  "version": "7.2.9",
+  "version": "7.2.10",
   "lockfileVersion": 3,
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-661-bjm258l3.web.app

eason-pan-bc (approved):
LGTM, thank you for sharing the knowledge to help me understand the FE

PR #677, opened 2024-07-02
22084 Clear data before reload

severinbeauvais:
*Issue #:* bcgov/entity#22084

*Description of changes:*
- app version = 7.3.9
- clear existing tasks/filings before fetching new data
- added missing await in retry method

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -348,7 +348,12 @@ export default class App extends Mixins(
   /** Called when component is created. */
   created (): void {
     // listen for reload data events
-    this.$root.$on('reloadData', () => this.fetchData())
+    this.$root.$on('reloadData', async () => {
+      // clear Todo List / Filing History List before fetching new data
+      this.setTasks([])
+      this.setFilings([])
+      await this.fetchData()
+    })
   }
 
   /** Called when component is mounted. */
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -348,7 +348,12 @@ export default class App extends Mixins(
   /** Called when component is created. */
   created (): void {
     // listen for reload data events
-    this.$root.$on('reloadData', () => this.fetchData())
+    this.$root.$on('reloadData', async () => {
+      // clear Todo List / Filing History List before fetching new data
+      this.setTasks([])
+      this.setFilings([])
+      await this.fetchData()
+    })
```
severinbeauvais, line 356: As per description in ticket, I believe the problem is that the tasks list isn't cleared (or reloaded) when this event is handled, so the same filing can be in both lists. See lines 692-693 below -- only the tasks or filings list is updated, not both. Since the filing is still in the Todo List then the retry happens again and again.

(I did see, on Mihai's computer, the same filing in both lists.)

This could also be done in fetchData() or storeBootstrapBusinessData(), but I figured clearing the lists here makes the most sense, since only the reload scenario was an issue.

JazzarKarim (approved):

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-677-44f51c1q.web.app

seeker25 (approved):

ArwenQin (approved):
LTGM

PR #678, opened 2024-07-10
22248 - Disable codecov failure

JazzarKarim:
*Issue #:* /bcgov/entity#22248

*Description of changes:*
- Disabled codecov failure

Following in the footsteps of Travis here: https://github.com/bcgov/sbc-auth/pull/2806

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           name: codecov-business-filings
-          fail_ci_if_error: true
+          fail_ci_if_error: false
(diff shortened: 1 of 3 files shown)


severinbeauvais (approved):
Looks fine.

I think you don't have to increment the app version if you do Create and Edit UIs, since the deployed image isn't changing, but no harm in it either.

JazzarKarim:
> Looks fine.
> 
> I think you don't have to increment the app version if you do Create and Edit UIs, since the deployed image isn't changing, but no harm in it either.

Makes sense. I'm thinking we also don't need to for the release. When releasing, we're technically not releasing anything with this commit. I'll revert it here and for the other 2 repos, not going to update it.

PR #685, opened 2024-07-22
21977 Fixed unit tests

severinbeauvais:
*Issue #:* bcgov/entity#21977

*Description of changes:*
- app version = 7.3.15
- changed FiledLabel.vue to check for filing before rendered
- deleted unused/invalid test suite
- renamed a test suite that was never executed


src/components/Dashboard/FilingHistoryList/FiledLabel.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="filed-label d-inline">
+  <div
+    v-if="!!filing"
+    class="filed-label d-inline"
+  >
```
severinbeauvais, line 5: This is because the test suite tries to mount this component with "null" filing and expected "nothing" to be rendered.

tests/unit/FiledAndPendingPaid.ts
severinbeauvais, line 1: This test suite was never run because it is missing "spec" in its name. Since the unit tests herein are not even correct for this test suite, I just deleted the file.

tests/unit/FiledLabel.spec.ts
severinbeauvais, line 1: As above, the name was incorrect. In this case, however, the unit tests matched the component so I didn't need to change anything else.

PR #691, opened 2024-08-20
22822 Changed to route query parameters to allow local filings without internal params

severinbeauvais:
*Issue #:* bcgov/entity#22822

*Description of changes:*
- app version = 7.3.22
- changed @volar-plugins/vetur to volar-service-vetur
- removed ARFilingYear from store and replaced with prop or query param
- now use arMinDate, arMaxDate and nextARDate from business store instead of root store
- changed route "params" to "query" when routing to a local filing
- no longer save root store data from Todo List (use existing business store data or pass as query param
- deleted root store properties
- deleted todo item properties
- added getters to business store
- deleted actions from root store
- changed route "params" to "query" when starting a local filing
- deleted obsolete CSS
- updated unit tests
- **updated example env file with new Status API URL**
- **deleted local correction filing**


package.json
```diff
@@ -83,6 +82,7 @@
     "vite": "4.5.2",
     "vite-plugin-environment": "^1.1.3",
     "vitest": "0.33.0",
+    "volar-service-vetur": "^0.0.54",
```
severinbeauvais, line 85: As per recent `npm i` message.

src/views/AgmExtension.vue
```diff
@@ -274,7 +274,7 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin, FilingM
 
     // this is the id of THIS filing
     // it must be 0 (meaning new filing) -- we do not support resuming a draft filing
-    this.filingId = +this.$route.params.filingId // number or NaN
+    this.filingId = +this.$route.query.filingId // number or NaN
```
severinbeauvais, line 277: This is a check that all local filings make.

Since only an internal route can set the `params`, this ensured that users could not enter a filing route manually in order to file something they would normally have access to.

This is now changed to `query` (eg, "?filingId=123") so that the filings can be routed to from an external UI.

Any thoughts on check we should do to prevent users from manually entering a local route? Or should we not worry about it because the Legal API will check if the filing is allowed anyway?

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-691-0kjaajkv.web.app

JazzarKarim (approved):
👍 👍

src/views/Correction.vue
severinbeauvais, line 1: This filing/view is defunct but was kept around in case we need to reuse it. At this point, I'd be OK with deleting it. Thoughts?

(There's no action that routes to this filing.)

tests/unit/AmalgamationSelection.spec.ts
```diff
@@ -20,7 +20,7 @@ describe('AmalgamationSelection', () => {
 
   beforeEach(() => {
     const router = mockRouter.mock()
-    router.push({ name: 'amalgamation-selection', params: { filingId: '0' } })
```
severinbeauvais, line 23: The page doesn't verify that this route parameter exists, so we don't need to set it here.

src/views/AgmExtension.vue
JazzarKarim: Hmmm, very good point.

I don't recall Legal API doing that check, does it?

severinbeauvais: Yes, I believe it checks if the filing is allowed.

src/views/Correction.vue
JazzarKarim: No strong opinions here from my side. I'm OK with deleting it.

src/views/AgmExtension.vue
JazzarKarim: So, I'm trying to put things together here in my head. 

There's gonna be an external UI which will be a starting point for all these local filings. So let's say AGM extensions, it's going to route directly to for example: `https://dev.business.bcregistry.gov.bc.ca/BC0883544/agm-extension` right?

Then, I guess a check here would be something that we don't need to do. Let's leave it for the BE to take care of that like you mentioned. Thoughts Sev?

severinbeauvais: > There's gonna be an external UI which will be a starting point for all these local filings. So let's say AGM extensions, it's going to route directly to for example: `https://dev.business.bcregistry.gov.bc.ca/BC0883544/agm-extension` right?

Yes, perhaps with a filing ID in the case of resuming a draft.

> Then, I guess a check here would be something that we don't need to do. Let's leave it for the BE to take care of that like you mentioned. Thoughts Sev?

I really liked the idea of the internal params object, as that positively prevented users from manually routing to a specific filing. I can't think of a way to do that now, unless we want to provide a signed URL query parameter or session storage variable maybe. @kialj876 , thoughts?

kialj876: I don't think you should be depending on session storage / local params because it complicates things that want to redirect here (like the new business dashboard). I don't think we need to worry about a user manually going here -- if they are a bad actor doing that they won't be able to file / there should be errors loading the page anyways. All of that needs to be handled on the backend because anyone can query the API

severinbeauvais: Thanks Karim and Kial for thinking about this with me.

So, no change to what I've done so far. Resolving this conversation.

src/views/StandaloneDirectorsFiling.vue
```diff
@@ -424,7 +424,7 @@ export default class StandaloneDirectorsFiling extends Mixins(CommonMixin, DateM
   certifyFormValid = false
   directorFormValid = true
   directorEditInProgress = false
-  filingId: number = null
+  filingId = NaN
```
severinbeauvais, line 427: This just seems more correct and it's inherently typed.

src/views/AmalgamationSelection.vue
```diff
@@ -379,11 +379,4 @@ h1 {
   right: 0;
   text-align: center;
 }
-
-// TODO: Remove disabled styling when short form amalgamations are implemented.
-.theme--light.v-btn.v-btn--disabled:not(.v-btn--flat):not(.v-btn--text):not(.v-btn--outlined) {
-  color: white !important;
-  background-color: $app-blue !important;
-  opacity: 0.4;
-}
```
severinbeauvais, line 388: Obsolete CSS. (V-buttons on this page are no longer disabled.)

src/views/Correction.vue
severinbeauvais: I've deleted the correction view and route and described it in the PR description so we can find the diffs later if we need to restore anything. Resolving.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-691-0kjaajkv.web.app

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-691-0kjaajkv.web.app

src/components/Dashboard/TodoList.vue
```diff
@@ -756,11 +756,6 @@ export default class TodoList extends Mixins(AllowableActionsMixin, DateMixin) {
   @Getter(useBusinessStore) isDisableNonBenCorps!: boolean
   // @Getter(useAuthenticationStore) isRoleStaff!: boolean
 
-  @Action(useRootStore) setARFilingYear!: (x: number) => void
-  @Action(useRootStore) setArMinDate!: (x: string) => void
-  @Action(useRootStore) setArMaxDate!: (x: string) => void
-  @Action(useRootStore) setNextARDate!: (x: string) => void
```
severinbeauvais, line 762: These are no longer set from Todo List, and they are no longer properties in the todo items. Instead, the AR Filing Year is passed as a route query parameter and the others are taken from the business record (store).

src/interfaces/todo-item-interface.ts
```diff
@@ -29,12 +29,7 @@ export interface TodoItemIF {
   // ARs only
   ARFilingYear?: number // YYYY
 
-  // COOP ARs only
-  arMinDate?: string // YYYY-MM-DD
-  arMaxDate?: string // YYYY-MM-DD
-
   // BCOMP ARs only
-  nextArDate?: string // YYYY-MM-DD
```
severinbeauvais, line 37: These are no longer properties of the todo items.

src/routes.ts
```diff
@@ -168,22 +167,6 @@ export default [
...
-      breadcrumb: [
-        {
-          text: FilingNames.CORRECTION,
-          disabled: false,
-          exact: true,
-          to: { name: Routes.CORRECTION }
-        }
-      ]
-    }
-  },
```
severinbeauvais, line 186: Removed local correction route (and view, see below).

src/stores/businessStore.ts
```diff
@@ -12,6 +12,8 @@ export const useBusinessStore = defineStore('business', {
     businessInfo: {
       adminFreeze: null,
       allowedActions: null,
+      arMaxDate: null,
+      arMinDate: null,
```
severinbeauvais, line 16: These are no longer optional in the interface.

Below, I provide getters to these properties.

src/stores/rootStore.ts
```diff
@@ -21,10 +21,6 @@ export const useRootStore = defineStore('root', {
     businessPhone: null,
     businessPhoneExtension: null,
     corpTypeCd: null,
-    ARFilingYear: null,
-    arMaxDate: null,
-    arMinDate: null,
-    nextARDate: null,
```
severinbeauvais, line 27: These are no longer stored here by Todo List. Instead, the AR Filing Year is passed as a route query parameter and the others are taken from the business record (store).

PR #692, opened 2024-09-06
21574 - Local Filing Return to Dashboard

ArwenQin:
*Issue #:* /bcgov/entity#21574

*Description of changes:*
For all filings, including the "local filings" currently done in Filings UI (and which will stay in Filings UI).
In this ticket, the following filings will "return" (actually, redirect) to the new Dashboard UI:
- AGM Extension
- AGM Location Change
- Amalgamation Out (WIP)
- Amalgamation Selection
- Annual Report
- Consent for Amalgamation Out (WIP)
- Consent for Continuation Out
- Continuation Out
- ~Correction (of otherwise unsupported filings such as Annual Report and Conversion)~ - not needed
- ~Digital Credentials (not technically a filing but managed in Filings UI)~ - not needed
- Standalone Directors Filing (aka Change of Directors)
- Standalone Office Address Filing (aka Change of Addresses)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-692-julisb6d.web.app

ArwenQin:
Test Note:
Please Try:
- AGM Extension
https://business-filings-dev--pr-692-julisb6d.web.app/BC0883568/agm-extension?filingId=0
- AGM Location Change
https://business-filings-dev--pr-692-julisb6d.web.app/BC0883568/agm-location-chg?filingId=0
- ~~Amalgamation Out~~: not enabled so cannot be tested
- ~~Amalgamation Selection~~: not enabled so cannot be tested
- Annual Report
https://business-filings-dev--pr-692-julisb6d.web.app/BC0871417/annual-report?filingId=0&arFilingYear=2024
- ~~Consent for Amalgamation Out (WIP)~~:not enabled so cannot be tested
- Consent for Continuation Out
https://business-filings-dev--pr-692-julisb6d.web.app/BC0883568/consent-continuation-out?filingId=0
- Continuation Out
https://business-filings-dev--pr-692-julisb6d.web.app/BC0874938/continuation-out?filingId=0
- Correction (of otherwise unsupported filings such as Annual Report and Conversion) - not needed, correction is in edit UI
- Digital Credentials (not technically a filing but managed in Filings UI) - not needed
- Standalone Directors Filing (aka Change of Directors)
https://business-filings-dev--pr-692-julisb6d.web.app/CP1002613/standalone-directors?filingId=0
- Standalone Office Address Filing (aka Change of Addresses)
https://business-filings-dev--pr-692-julisb6d.web.app/CP1002613/standalone-addresses?filingId=0

ArwenQin:
I will update the unit tests later if the updates are ok.

devops/vaults.env
```diff
@@ -13,6 +13,7 @@ VUE_APP_DASHBOARD_URL="op://web-url/$APP_ENV/business/DASHBOARD_URL"
 VUE_APP_SITEMINDER_LOGOUT_URL="op://web-url/$APP_ENV/siteminder/SITEMINDER_LOGOUT_URL"
 VUE_APP_BUSINESS_CREATE_URL="op://web-url/$APP_ENV/business-create/BUSINESS_CREATE_URL"
 VUE_APP_BUSINESS_EDIT_URL="op://web-url/$APP_ENV/business-edit/BUSINESS_EDIT_URL"
+VUE_APP_BUSINESS_DASH_URL="op://web-url/$APP_ENV/business-edit/BUSINESS_DASH_URL"
```
severinbeauvais, line 16: Change section from business-edit to `business-dash`.

src/views/AgmExtension.vue
```diff
@@ -278,6 +279,11 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin, FilingM
 
     // if required data isn't set, go back to dashboard
     if (!this.getIdentifier || this.filingId !== 0) {
+      if (GetFeatureFlag('use-business-dashboard')) {
+        const dashboardUIUrl = `${this.getBusinessDashUrl}/${this.getIdentifier}`
+        navigate(dashboardUIUrl)
```
severinbeauvais, line 284: To reduce code duplication (and to show that all 4 places in here do the same thing), I recommend creating a small method that builds the URL and does the navigation. You can have an optional filingId parameter to handle line 395.

severinbeauvais: On second thought, check if all filings use Common Mixin and move that method there maybe?

In fact, that method could look at the FF and determine whether to route or redirect. What do you think?

ArwenQin: I agree, I will update the Mixin, thanks.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-692-julisb6d.web.app

ArwenQin:
Updated the common Mixin and re-structured the codes.
I will update the unit tests if these changes are ok.

src/mixins/common-mixin.ts
```diff
@@ -68,4 +74,27 @@ export default class CommonMixin extends Vue {
     }
     return true
   }
+
+  /**
+   * Navigates to the dashboard page, optionally with a filing ID.
+   * @param identifier The identifier to include in the dashboard URL
+   * @param filingId The filing ID to include in the dashboard URL (optional)
+   */
+  protected navigateToDashboard (identifier: string, filingId?: number): void {
```
severinbeauvais, line 83: You should assign a default filingId (eg, null or 0) so you can stay away from undefined, below.

ArwenQin: Updated

src/views/AmalgamationSelection.vue
severinbeauvais, line 1: Whether to change this file is a good question, because it's possible the new business dashboard will provide this page (instead of relying on Filings UI to provide it).

I think changing this page is a safe option, so I accept the changes below.

src/mixins/common-mixin.ts
```diff
@@ -68,4 +74,27 @@ export default class CommonMixin extends Vue {
     }
     return true
   }
+
+  /**
+   * Navigates to the dashboard page, optionally with a filing ID.
+   * @param identifier The identifier to include in the dashboard URL
+   * @param filingId The filing ID to include in the dashboard URL (optional, defaults to null)
+   */
+  protected navigateToDashboard (identifier: string, filingId: number | null = null): void {
```
JazzarKarim, line 83: Can we give this function a more specific name? Might be a bit generic here. I'm thinking, `navigateToNewDashboard` or `navigateToBusinessDashboard`? What do you think?

src/views/AgmLocationChg.vue
```diff
@@ -411,7 +411,7 @@ export default class AgmLocationChg extends Mixins(CommonMixin, DateMixin, Filin
 
     // if required data isn't set, go back to dashboard
     if (!this.getIdentifier || this.filingId !== 0) {
-      this.$router.push({ name: Routes.DASHBOARD })
+      this.navigateToDashboard(this.getIdentifier)
```
severinbeauvais, line 414: According to line 413, you might get here because the identifier is falsy (0 or null or undefined or '') and then you're trying to use the identifier in a navigation!

I don't think you'll ever get here if identifier is falsy, so I recommend just removing that check, here and in the other filings below.

In the worst case, if you get here then you'll try to navigate to the business dashboard with an invalid identifier and the business dashboard can try to handle that. This is the same scenario as the existing check so it's no worse.

JazzarKarim (approved):
Looks good 👍

src/mixins/common-mixin.ts
```diff
@@ -68,4 +74,27 @@ export default class CommonMixin extends Vue {
...
+
+  /**
+   * Navigates to the dashboard page, optionally with a filing ID.
+   * @param identifier The identifier to include in the dashboard URL
+   * @param filingId The filing ID to include in the dashboard URL (optional)
+   */
+  protected navigateToDashboard (identifier: string, filingId?: number): void {
+    if (GetFeatureFlag('use-business-dashboard')) {
+      // Disable 'beforeunload' event
+      window.onbeforeunload = null
```
severinbeauvais, line 86: Why is this needed if the FF is set but not if it isn't (below)?

I kinda don't like that we're changing a window handler here... I prefer to have the code that sets and the code that clears event handlers nearby each other -- kind of a symmetry,  you know?

src/mixins/common-mixin.ts
severinbeauvais, line 83: By the way, we usually don't bother with a union type that includes null. We just assume that any property/variable can be null and code accordingly.

ArwenQin: Thanks, agree. I will update it to navigateToBusinessDashboard

src/mixins/common-mixin.ts
ArwenQin: I noticed this when I click "Save and Resume" - without clearing the event handler, a dialog pops out saying "you are leaving this page, you may lost unsaved..."., because this page is navigating to an external URL. But actually the "Save and Resume" will save the information and this dialog is not needed.
However, when the feature flag is disabled, the Vue Router doesn't trigger it.

src/mixins/common-mixin.ts
ArwenQin: Updated

src/views/AgmLocationChg.vue
ArwenQin: Updated

src/mixins/common-mixin.ts
severinbeauvais: Yes, I see.

In fact, this check for unsaved changes isn't even working correctly right now. Any exit from the filing view without first saving your data should trigger the browser alert, but if you start a filing and change some things and click the breadcrumb to return to the dashboard, you don't get the alert. This is because the user is not leaving the "window" (web app); they are only leaving one of the pages.

(Speaking of breadcrumbs, did you update it to navigate to the business dashboard?)

Let's not worry about this small bug right now.

However, I'm curious why you are getting this alert on save and resume. Theoretically, on successful save, the "haveChanges" flag should be cleared the the event handler should do nothing. Please test a little more to find which scenario is causing this and why. I don't really like the solution above.

As for the symmetry of setting and clearing event handlers, I still think this is important generally, but clearing is not so important when navigating to another UI because the window object gets destroyed and re-created.

src/mixins/common-mixin.ts
JazzarKarim: Thanks! Resolving.

src/views/AmalgamationSelection.vue
severinbeauvais: Arwen, you can resolve this conversation after you've read it.

src/views/AgmExtension.vue
```diff
@@ -277,8 +277,8 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin, FilingM
     this.filingId = +this.$route.query.filingId // number or NaN
 
     // if required data isn't set, go back to dashboard
```
severinbeauvais, line 279: This comment is now obsolete. See comment on line 276 and maybe move it here.

src/views/AgmLocationChg.vue
```diff
@@ -410,8 +410,8 @@ export default class AgmLocationChg extends Mixins(CommonMixin, DateMixin, Filin
     this.filingId = +this.$route.query.filingId // number or NaN
 
     // if required data isn't set, go back to dashboard
```
severinbeauvais, line 412: Same as above.

src/views/AgmExtension.vue
ArwenQin: Updated the comments

src/mixins/common-mixin.ts
ArwenQin: Thanks!
I fixed the problem, the onClickSaveResumeFinish () function didn't clear flag (this.haveChanges = false). I updated it.
I updated the navigate to the business dashboard breadcrumbs too.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-692-julisb6d.web.app

src/mixins/common-mixin.ts
```diff
@@ -68,4 +74,26 @@ export default class CommonMixin extends Vue {
...
   }
+
+  /**
+   * Navigates to the dashboard page, optionally with a filing ID.
+   * @param identifier The identifier to include in the dashboard URL
+   * @param filingId The filing ID to include in the dashboard URL (optional, defaults to null)
+   */
+  protected navigateToBusinessDashboard (identifier: string, filingId: number = null): void {
+    if (GetFeatureFlag('use-business-dashboard')) {
+      // Disable 'beforeunload' event
```
severinbeauvais, line 85: Obsolete comment.

src/views/AgmLocationChg.vue
severinbeauvais: Looks good.

src/mixins/common-mixin.ts
severinbeauvais: Excellent! Good fix.

The breadcrumbs look good, too.

severinbeauvais (approved):
Looks good.

Remember to update the app version.

Also, please provide some Test Notes in the ticket to list what sorts of things QA should verify (and that you've also verified).

ArwenQin:
> Looks good.
> 
> Remember to update the app version.
> 
> Also, please provide some Test Notes in the ticket to list what sorts of things QA should verify (and that you've also verified).

Sure, I will try to update some unit tests tmr. Thank you!

ArwenQin:
@severinbeauvais I updated some unit tests. Is it good to merge now? Thanks!

src/mixins/common-mixin.ts
severinbeauvais: Please remove this comment.

severinbeauvais:
Almost good to merge. There's an obsolete comment. After that, merge away :)

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-692-julisb6d.web.app

PR #694, opened 2024-09-20
Replace auth-api URL to point to the new instance

ketaki-deodhar:
*Description of changes:*

- Replaced auth-api URL to point to the new instance


severinbeauvais (approved):
Looks good. I agree that there's no need to update the app version.

ArwenQin (approved):

PR #700, opened 2024-11-06
23967 Removed spinner for approved continuation in drafts

leodube-aot:
*Issue #:* /bcgov/entity#23967

*Description of changes:*
- Removed spinner for approved continuation in drafts

### DIFF
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -347,7 +347,11 @@
                 <!-- NB: blocks below are mutually exclusive, and order is important -->
 
                 <!-- this loading button pre-empts all buttons below -->
-                <template v-if="inProcessFiling === item.filingId">
+                <!-- special case to not show spinner for approved continuation in drafts -->
+                <template 
+                  v-if="inProcessFiling === item.filingId && 
+                    !(EnumUtilities.isTypeContinuationIn(item) && EnumUtilities.isStatusApproved(item))"
+                  >
                   <v-btn
                     text
                     loading
--- a/src/components/Dashboard/TodoList.vue
+++ b/src/components/Dashboard/TodoList.vue
@@ -347,11 +347,7 @@
                 <!-- NB: blocks below are mutually exclusive, and order is important -->
 
                 <!-- this loading button pre-empts all buttons below -->
-                <!-- special case to not show spinner for approved continuation in drafts -->
-                <template 
-                  v-if="inProcessFiling === item.filingId && 
-                    !(EnumUtilities.isTypeContinuationIn(item) && EnumUtilities.isStatusApproved(item))"
-                  >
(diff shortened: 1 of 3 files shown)


JazzarKarim:
Watch out for lint errors Leo!

JazzarKarim (approved):

src/components/Dashboard/TodoList.vue
```diff
@@ -343,7 +343,11 @@
                 <!-- NB: blocks below are mutually exclusive, and order is important -->
 
                 <!-- this loading button pre-empts all buttons below -->
-                <template v-if="inProcessFiling === item.filingId">
+                <!-- special case to not show spinner for approved continuation in drafts -->
+                <template 
+                  v-if="inProcessFiling === item.filingId && 
+                    !(EnumUtilities.isTypeContinuationIn(item) && EnumUtilities.isStatusApproved(item))"
+                  >
```
severinbeauvais, line 350: Why not change the logic for `inProcessFiling` in the code instead?

I think that, with this change, the 5x polling is still being done in the background. It doesn't need to be done for a continuation in authorization.

leodube-aot: Good point, I've moved the check into the `highlighTask` method since the `inProcessFiling` is just a simple number type. 

I also noticed the bug was occurring for continuation in filings with status `ACTIVE`. I'm not sure what other status types we can expect for continuation in filings, so I changed it to only display the spinner for `PENDING` status continuation ins. Let me know if that makes sense to do or not.

severinbeauvais: Thanks for the change. Looks good to me.

severinbeauvais (approved):

src/components/Dashboard/TodoList.vue
severinbeauvais: > I also noticed the bug was occurring for continuation in filings with status `ACTIVE`. I'm not sure what other status types we can expect for continuation in filings, so I changed it to only display the spinner for `PENDING` status continuation ins. Let me know if that makes sense to do or not.

The continuation in draft had a status of Active? That's unexpected. I think the normal flow is:
DRAFT → AWAITING_REVIEW → APPROVED → PENDING → PAID → COMPLETED

Basically, we only want the retry to work for the PENDING (ie, while we wait for PAID).

Ask if this doesn't make sense.

leodube-aot: Right my bad, I was looking at the wrong status field. I was looking at the one in `continationIn` rather than the `header`.

Screenshot 2024-11-06 at 3 40 54 PM

That makes sense though, thanks for the clarification!

eason-pan-bc:
maybe consider adjusting the version number and run `npm i` again? I just merged a PR in.

eason-pan-bc (approved):

PR #707, opened 2024-12-09
24657 - Notice of Withdrawal filing

ArwenQin:
*Issue #:* /bcgov/entity#24657

*Description of changes:*
- Create the NOW Filing Page
- 4 new components: Record to be withdrawn, Court Order and POA, Folio/Ref Number, Staff Payment


src/components/common/Certify.vue
```diff
@@ -51,7 +51,7 @@
           <strong>Date:</strong> {{ formattedCurrentDate || '[unknown]' }}
         </p>
         <p class="certify-clause">
-          {{ message }}
+          <strong v-if="message">Note:</strong> {{ message }}
         </p>
```
ArwenQin, line 55: Updated the format per design, so needed to update the CertifyText for all business types

src/resources/SP.ts
```diff
@@ -20,6 +20,11 @@ export const BusinessConfigSp = {
       feeCode: null,
       displayName: 'Change Of Directors',
       certifyText: ''
+    },
+    {
+      feeCode: null,
+      displayName: 'Notice of Withdrawal',
+      certifyText: ''
     }
```
ArwenQin, line 28: Do we add it for SP? I added a placeholder here

src/views/NoticeOfWithdrawal.vue
```diff
@@ -0,0 +1,1008 @@
...
+                  class="py-8 px-5"
+                >
+                  <PlanOfArrangement
+                    :autoValidation="showErrors"
+                    :hasDraftPlanOfArrangement="hasPlanOfArrangement"
+                    :hasDraftComeIntoEffect="hasComeIntoEffect"
+                    @emitPoa="hasPlanOfArrangement=$event"
+                    @emitEffect="hasComeIntoEffect=$event"
+                  />
+                </v-card>
```
ArwenQin, line 164: I saw Jacqueline removed this section from design last Friday, will need to confirm with her tmr
image

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

src/components/NoticeOfWithdraw/PlanOfArrangement.vue
severinbeauvais, line 1: Please fix folder name.

src/resources/SP.ts
severinbeauvais: Not needed (nor for GP) because their registrations cannot be FE.

Example: https://dev.create.business.bcregistry.gov.bc.ca/registration-review-confirm?id=TsmwTlWtc9&accountid=2995

src/resources/GP.ts
```diff
@@ -20,6 +20,11 @@ export const BusinessConfigGp = {
       feeCode: null,
       displayName: 'Change Of Directors',
       certifyText: ''
+    },
+    {
+      feeCode: null,
+      displayName: 'Notice of Withdrawal',
+      certifyText: ''
     }
```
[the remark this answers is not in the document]
severinbeauvais: See comment above :)

src/components/NoticeOfWithdraw/PlanOfArrangement.vue
severinbeauvais: Thanks. Resolving.

src/routes.ts
```diff
@@ -87,6 +88,22 @@ export default [
       ]
     }
   },
+  {
+    path: '/notice-of-withdrawal/:filingToBeWithdrawn',
```
severinbeauvais, line 92: Good idea 👍

Background: the other local filings don't do this because their filing id (0 or xxx) was initially passed internally on the route, but now, since an external dashboard routes to local filings, the filing is a URL parameter.

What you've done here is better, but we're not going to change the others quite yet since the external dashboard would also need to be updated.

src/services/enum-utilities.ts
```diff
@@ -220,6 +220,11 @@ export default class EnumUtilities {
     return (item.name === FilingTypes.INCORPORATION_APPLICATION)
   }
 
+  /** Returns True if filing is a Notice of Withdraw. */
```
severinbeauvais, line 223: Typo.

My fingers can't type "withdrawal" in one go, either. I always type "width..." 😆

src/views/NoticeOfWithdrawal.vue
```diff
@@ -0,0 +1,1012 @@
...
+      )
+    }
+
+    /** True when saving, saving and resuming, or filing and paying. */
+    get busySaving (): boolean {
+      return (this.saving || this.savingResuming || this.filingPaying)
+    }
+
+    /** Called when component is created. */
+    created (): void {
```
severinbeauvais, line 423: Please add a check for staff either here or in mounted().

A browser alert should suffice if the user is not staff. There may be other examples like this in here or in Edit UI.

severinbeauvais: Example: https://github.com/bcgov/business-edit-ui/blob/f830cd613f410af4a9ce28c89a1da4418c5ff4cf/src/views/Correction.vue#L81

Or do something better 😄

src/services/enum-utilities.ts
ArwenQin: Thanks! updated

src/resources/GP.ts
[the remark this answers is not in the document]
ArwenQin: thank you, removed

src/resources/SP.ts
ArwenQin: thank you, removed

src/views/NoticeOfWithdrawal.vue
ArwenQin: updated, how about this? Line 426-429

ArwenQin:
Now just waiting for Jacqueline to confirm for the POA Arrangement section. 
I will add some unit tests.

meawong (commented):
Looks good so far

src/views/NoticeOfWithdrawal.vue
severinbeauvais: That should be good. Did you test it? What does it look like and what happens next?

ArwenQin: image
 And it goes back to the dashboard.

severinbeauvais: The experience looks good, but:
- the title is wrong (I'm starting, not resuming)
- it doesn't tell me that I cannot do this because I'm not staff; in other words, this isn't an "error" -- it's intentional

@jacqueline-williams-549 Thoughts?

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
```diff
@@ -0,0 +1,134 @@
...
+    // Local properties
+    private planOfArrangement = false
+    private comeIntoEffect = false
+    private valid = false
+
+    /** Called when component is mounted. */
+    mounted (): void {
+      // Set default draft values if they exist
+      if (this.hasDraftPlanOfArrangement) this.planOfArrangement = this.hasDraftPlanOfArrangement
+      if (this.hasDraftComeIntoEffect) this.comeIntoEffect = this.hasDraftComeIntoEffect
```
severinbeauvais, line 77: This won't update the local properties if the props ever change. Could that ever happen here (eg, draft is loaded after page is displayed)?

As an alternative, you could "watch" the props and use the `{ immediate: true }` option.

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
```diff
@@ -0,0 +1,134 @@
...
+        this.comeIntoEffect = false
+      }
+    }
+
+    /** Emit plan of arrangement. */
+    @Watch('planOfArrangement')
+    @Emit('emitPoa')
+    private emitPoa (): boolean {
+      return this.planOfArrangement
+    }
```
severinbeauvais, line 93: Were you thinking of the updating these emit functions like you did the others?

Also lines 98 and 104 below.

Just curious. If not that's OK.

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
ArwenQin: Yes, sorry, I have not completed the POA section. I thought Jacqueline will remove it from the design. Just confirmed with her we will keep this section for now. I will work on this part

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
```diff
@@ -0,0 +1,122 @@
...
+    components: {
+      VcardTemplate
+    }
+  })
+export default class RecordToBeWithdrawn extends Vue {
+    @Prop({ required: true }) readonly filingToBeWithdrawn!: string
+
+    @Getter(useFilingHistoryListStore) getFilings!: Array<ApiFilingIF>
+
+    getFilingToBeWithdrawn (): ApiFilingIF| undefined {
```
severinbeauvais, line 90: Lint: add space before `|`.

Also we don't usually care about union to undefined or null but you can leave it.

Note: IMO, it's better to return null than undefined if "find" doesn't find it. This is because null properties are reactive but undefined ones are not. You can add `|| null` to line 93 to do this.

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
```diff
@@ -0,0 +1,122 @@
...
+      const filingId = Number(this.filingToBeWithdrawn)
+      const filings = this.getFilings as ApiFilingIF[]
+      return filings.find(filing => filing.filingId === filingId)
+    }
+
+    getFormattedFilingDate (): string {
+      const filing = this.getFilingToBeWithdrawn()
+      if (filing && filing.effectiveDate) {
```
severinbeauvais, line 98: Conditional chaining ftw? (This is OK, too.)

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
```diff
@@ -0,0 +1,122 @@
...
+        const date = new Date(filing.effectiveDate)
```
severinbeauvais, line 99: Use Date Mixin to convert between date formats.

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
```diff
@@ -0,0 +1,122 @@
...
+}
+</script>
+
+  <style lang="scss" scoped>
+  @import '@/assets/styles/theme.scss';
+
+  .col-12 {
+    font-size: $px-16;
+  }
+  </style>
```
severinbeauvais, line 122: Remove indent.

Also, if you think it's cleaner, you can use base classs "font-16" instead of line 120.

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
```diff
@@ -0,0 +1,125 @@
...
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
+import { FormIF } from '@bcrs-shared-components/interfaces'
+
+  @Component({})
```
severinbeauvais, line 40: Remove indent.

Also the styles block below.

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
```diff
@@ -0,0 +1,125 @@
...
+</script>
+
+  <style lang="scss" scoped>
+  @import '@/assets/styles/theme.scss';
+
+  #reference-number-label {
+    font-size: $px-16;
+    font-weight: bold;
+    color: $gray9;
+  }
```
severinbeauvais, line 108: We have classes for all of these already.

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
ArwenQin: I looked at other components, I think it won't happen: CourtOrderPoa.vue

severinbeauvais: So this component isn't rendered (mounted) until a draft filing is resumed (if there is one)?

severinbeauvais: I see you're protected by the `dataLoaded` boolean.

Therefore, as long as the props don't change for another reason, doing it this way, once, in mounted(), is OK.

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
ArwenQin: Using Date Mixin, the "am" is not capitalized, while in design it's "AM":
image

Is it ok?

severinbeauvais: Use the mixin so that the display is consistent throughout the app.

Let Jacqueline know. If the new design is to capitalize it then we will need a new ticket so we can make the change to the  mixin in all UIs.

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
ArwenQin: Updated, same as the others

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
ArwenQin: Updated

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
ArwenQin: Updated, use the default font-14, keep consistent with other components

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
ArwenQin: Done

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
ArwenQin: Done, adjusted the font

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
severinbeauvais: Arwen, please resolve this conversation after you've read it. Thx.

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
severinbeauvais: I was thinking `if (filing?.effectiveDate)` but what you've done instead is also good. Resolving.

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
severinbeauvais: I think you shouldn't use `new Date()` on line 100 either. There's a Date Mixin function to do this.

The reason for the Date Mixin is to handle absolute timezones (esp if the user is not in Pacific tz, and also because we can't trust the user's browser settings).

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
severinbeauvais: I don't see the change to 14px. Did you just remove the font size altogether?

ArwenQin: Yes, the default is 14 px

severinbeauvais: OK, that's good.

It's actually better to use the default, that way if this subcomponent is used anywhere else then it will pick up the default from its new parent.

ArwenQin:
I saw Jacqueline updated the design for POA Arrangement section to include the court order, I will update it now

src/components/NoticeOfWithdrawal/RecordToBeWithdrawn.vue
ArwenQin: The filing.effectiveDate is in "Wed, 27 Nov 2024 17:34:07 GMT" format. I didn't find a Date Mixin function to handle this format. Do you have suggestions? Thanks a lot

severinbeauvais: image

Then you are right -- using `new Date()` is acceptable here. Resolving!

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
severinbeauvais, line 1: Looks good but please move this to the components/dialogs/ folder as it could be reused by other filings.

tests/unit/NoticeOfWithdrawal.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe('Notice of Withdrawal view', () => {
+  beforeEach(() => {
+    // init store
+    rootStore.keycloakRoles = ['staff']
+  })
+
+  it('mounts the sub-components properly', async () => {
+    const $route = { query: { filingId: '0' }, params: { filingToBeWithdrawn: '12345' } }
+
```
ArwenQin, line 43: Could you please help: How to mock the params in route?
If I comment out Line 403 in NoticeOfWithdrawal.vue
(    filingToBeWithdrawn = this.$route.params.filingToBeWithdrawn || '')
the Unit tests run successfully.
With the route.params, it shows **TypeError: Cannot read properties of undefined (reading 'params')**
I tried some solutions but didn't figure out

severinbeauvais: Would this work?

https://github.com/bcgov/business-filings-ui/blob/f9747e4b20d50da231dcd8a45d1feb45ffa5a498/tests/unit/AgmLocationChg.spec.ts#L117

severinbeauvais: Maybe this?

```
const $route = { query: { filingId: '0' filingToBeWithdrawn: '12345' } }
```

ArwenQin: Still showing TypeError...Seems it can mock query but didn't work for param

src/components/NoticeOfWithdrawal/PlanOfArrangement.vue
severinbeauvais: WAIT - I meant to add this comment to the new dialog.

THIS component is obviously specific to this filing.

src/components/NoticeOfWithdrawal/StaffRoleErrorDialog.vue
severinbeauvais, line 1: Looks good but please move this to the components/dialogs/ folder as it could be reused by other filings.

tests/unit/NoticeOfWithdrawal.spec.ts
severinbeauvais: I wonder if the parameter format you used in routes.ts is supposed to be used as a prop?

See https://v3.router.vuejs.org/guide/essentials/passing-props.html

severinbeauvais: Maybe this isn't the right way to do it:

image

Sample router call:

image

https://v3.router.vuejs.org/guide/essentials/navigation.html#programmatic-navigation

severinbeauvais: Or maybe our mock router (see mockRouter.ts) doesn't support your param (which might be fixable with another router option?)?

Sorry, looks like you're doing something new here and I don't know how to help.

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
```diff
@@ -0,0 +1,119 @@
...
+        (v: string) => (!v || !(v.length > 20)) || 'Folio or Reference number is invalid' // too long
+      ]
+      this.$refs.refNumRef.validate()
+    } else this.clearValidations()
+  }
+
+  /** Emit reference number. */
+  @Watch('referenceNumber')
+  @Emit('referenceNumber')
+  emitReferenceNumber (): void {}
```
severinbeauvais, line 92: I know that other places follow this syntax/style for watchers/emitters but it looks wrong to me.

What you can't tell from this is the type of what's being emitted. The Watch decorator implicitly pushes a parameter into the function for the Emit decorator to emit.

I'd prefer to see this:
```
  @Watch('referenceNumber')
  @Emit('referenceNumber')
  emitReferenceNumber (referenceNumber: string): void {}
```

Does it work? What do you think?

severinbeauvais: Or even
```
  @Watch('referenceNumber')
  @Emit('referenceNumber')
  emitReferenceNumber (val: string): void {}
```

ArwenQin: I thought about this.. but it shows an error the variable is declared but never used:
image

What do you prefer?

src/views/NoticeOfWithdrawal.vue
```diff
@@ -0,0 +1,1045 @@
...
+      // do not proceed if user is not staff
+      if (!this.isRoleStaff) {
+        this.staffRoleErrorDialog = true
+        throw new Error('This is a Staff only Filing.')
+      }
+
+      // init
+      this.setFilingData([])
+      this.filingToBeWithdrawn = this.$route.query.filingToBeWithdrawn as string || ''
+
```
ArwenQin, line 442: like: BC0883739/notice-of-withdrawal?filingToBeWithdrawn=153004&filingId=0

src/components/NoticeOfWithdrawal/StaffRoleErrorDialog.vue
ArwenQin: Done, thanks!

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
severinbeauvais: I prefer the verbose one. You can hide the error using eslint-ignore.

src/routes.ts
```diff
@@ -87,6 +88,22 @@ export default [
       ]
     }
   },
+  {
+    path: '/notice-of-withdrawal',
+    name: Routes.NOTICE_OF_WITHDRAWAL,
```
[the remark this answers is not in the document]
severinbeauvais: That looks good.

What was it before?

src/views/NoticeOfWithdrawal.vue
severinbeauvais: OK.

You could move it near the other route query parameter (line 455).

Are you sure id should be a string?

src/routes.ts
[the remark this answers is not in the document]
ArwenQin: It was path: '/notice-of-withdrawal/:filingToBeWithdrawn',

BC0883739/notice-of-withdrawal/153004?filingId=0

tests/unit/NoticeOfWithdrawal.spec.ts
severinbeauvais: Arwen, if this is fixed, please resolve this conversation.

src/routes.ts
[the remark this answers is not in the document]
severinbeauvais: Oh, yeah, that is different. It was a dynamic route instead of a route query parameter. You were extracting it from params but I think that's unusual.

FYI, there's one more way this could be done, and that's to use a colon for the parameter in the route spec, and then have the router convert the parameter to a prop for the component to use. Example: https://github.com/bcgov/sbc-auth/blob/d56df39c91c515dc0224bc72b0ec925c419de5c1/auth-web/src/routes/router.ts#L235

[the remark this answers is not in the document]
severinbeauvais: Don't change your code. Resolving.

src/views/NoticeOfWithdrawal.vue
ArwenQin: It should be a string. But I changed the type to any, just to be safe

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
ArwenQin: Updated

src/views/NoticeOfWithdrawal.vue
```diff
@@ -0,0 +1,1042 @@
...
+    resumeErrorDialog = false
+    saveErrorReason = null as SaveErrorReasons
+    paymentErrorDialog = false
+
+    // other variables
+    totalFee = 0
+    dataLoaded = false
+    loadingMessage = ''
+    filingId = 0 // id of this Notice if Withdrawal filing
+    filingToBeWithdrawn: any // id of filing to be withdrawn
```
ArwenQin, line 391: Should be a string. just to be safe here

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
severinbeauvais: Update all of them?

src/views/NoticeOfWithdrawal.vue
severinbeauvais: I prefer "string" instead of "any".

However, since it's an id, I'm thinking it's actually a number. But it might be easier to use in code as a string.

ArwenQin: Got it, updated

src/components/NoticeOfWithdrawal/ReferenceNumber.vue
ArwenQin: Sure, just updated

ArwenQin:
Updated the Court Order and POA section per new design.
I think it's ready for a PR review now

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

ArwenQin:
Test:
https://business-filings-dev--pr-707-i7dbekyt.web.app/BC0883739/notice-of-withdrawal?filingToBeWithdrawn=153004&filingId=0

eason-pan-bc (commented):
I'm not sure if this is applicable, but I found the padding between rows in he "Record to be withdrawn" section is a bit different between the design on Figma and on the website.

In Figma, it is 24px 
image

On the website, it is 28px 
image

ArwenQin:
> In Figma, it is 24px image
> 

Agree. 24 px seems better. I updated the space to py-6

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

ArwenQin:
Although in design, the default font size is 16 px; I think the actual default font size should be 14-px, consistent with other filings
image

Current I use 14-px, which is the same as other filings

src/views/NoticeOfWithdrawal.vue
```diff
@@ -535,7 +543,8 @@ export default class NoticeOfWithdrawal extends Mixins(CommonMixin, DateMixin, F
           } as StaffPaymentIF
         }
 
-        // load POA and arrangement properties
+        // load Court Order, POA and arrangement properties
+        const courtOrderNumber = filing.noticeOfWithdrawal.courtOrderNumber
```
severinbeauvais, line 547: Can you post sample filing JSON with all properties present?

Ultimately, I'm wondering if the schema should be updated, or if we should add validations in Legal API.

Ref: https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/notice_of_withdrawal.json

ArwenQin: {
    "filing": {
        "header": {
            "name": "noticeOfWithdrawal",
            "certifiedBy": "AQ",
            "email": "Aimee.Gao@gov.bc.ca",
            "date": "2024-12-13",
            "documentOptionalEmail": "test@test.ca",
            "routingSlipNumber": "123456789",
            "priority": false
        },
        "business": {
            "foundingDate": "2024-11-15T21:58:25.000+00:00",
            "identifier": "BC0883739",
            "legalName": "SEVERIN CONVERT ULC TO BC LIMITED CORP.",
            "legalType": "BC"
        },
        "noticeOfWithdrawal": {
            "filingToBeWithdrawn": "153004",
            "referenceNumber": "TT123",
            "courtOrderNumber": "TT0010045",
            "hasPlanOfArrangement": true,
            "hasComeIntoEffect": true
        }
    }
}

severinbeauvais: I do see a mismatch between the JSON and the schema:
- missing filingId (this is probably the filing to withdraw)
- court order properties should be in court order sub-object

Maybe also look at what was done in [23349](https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/23349) regarding validation.

severinbeauvais: Also, I don't think that's how we save the PoA state.

ArwenQin: Got it, I will update for the court order properties including the POA.
For the filingID: is it assigned from LegalServices.createFiling? As this NOW filing is not supported by LegalServices yet, the filingID is not assigned

ArwenQin: Quick question: if there is a POA, does it mean court order number is necessary?
Like in Consent to Continue Out:
image
When the user select POA, they must enter a court order number?

severinbeauvais: No, I think filingId is the id of the filing being withdrawn. I'm not certain -- check with the original author of the schema or the developer of the Legal API validations ticket.

Yes, a court order number is required if the POA checkbox is checked.

ArwenQin: Thank you.
filingId of the filing being withdrawn: I added it here
image
I think it needs to be added in the schema.

POA- I will update this section now

severinbeauvais: > I think it needs to be added in the schema.

I think `filingId` in the schema **is** the filing that is being withdrawn.

ArwenQin: For the Court Order schema:
https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/court_order.json
image
Do we need to add a property for the Come into Effect?


I think the  "effectOfOrder" is for the first check box - POA.
This seems new, I saw the second check box for the first time

ArwenQin: I'm thinking about add a "effectOfArrangement" to the schema:
"effectOfOrder"for the first check box: means there is a POA, same as other filings
"effectOfArrangement" for the second check box: means the POA has come into effect

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

src/views/NoticeOfWithdrawal.vue
ArwenQin: after update:
{
    "header": {
        "name": "noticeOfWithdrawal",
        "certifiedBy": "AQ",
        "email": "Aimee.Gao@gov.bc.ca",
        "date": "2024-12-13",
        "documentOptionalEmail": "test@test.ca",
        "routingSlipNumber": "123415689",
        "priority": false
    },
    "business": {
        "foundingDate": "2024-11-15T21:58:25.000+00:00",
        "identifier": "BC0883739",
        "legalName": "SEVERIN CONVERT ULC TO BC LIMITED CORP.",
        "legalType": "BC"
    },
    "noticeOfWithdrawal": {
        "filingId": "153004",
        "referenceNumber": "TT0001",
        "courtOrder": {
            "fileNumber": "TT0010045",
            "effectOfOrder": "planOfArrangement",
            "effectOfArrangement": "arrangementEffect"
        }
    }
}

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

src/views/NoticeOfWithdrawal.vue
severinbeauvais: I think we label the Reference Number `folioNumber` in other filings, and it's actually in the header object.

I think it's allowed to add properties to a schema object -- like you've done for effectOfArrangement -- but I've seen a more recent UI design and this is a normal court order object (that applies to THIS filing). See other filings for examples.

Meanwhile, the 2 checkboxes apply to the FE filing and are separate from the court order object above. So, you need 2 properties for those, not in the court order object.

Maybe you should take a peek at the more recent UI designs just so you know what's (probably) coming and so that you can get the data handling correct now. Ask Jacqueline for a link.

severinbeauvais:
The page looks very good.

I think there's still too much whitespace around some components (and not enough between others).

I am unable to save the filing and cannot test resume.

Let's concentrate on what it will take to finish this ticket asap, and then we'll create follow-up tickets for more changes or fixes.

ArwenQin:
Removed the folio/reference number section.
Updated the space.
Ready to merge.

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-707-i7dbekyt.web.app

src/views/NoticeOfWithdrawal.vue
eason-pan-bc: > Can you post sample filing JSON with all properties present?
> 
> Ultimately, I'm wondering if the schema should be updated, or if we should add validations in Legal API.
> 
> Ref: https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/notice_of_withdrawal.json

Yes, we may need to update the schema and validations for NoW, since those tickets were done before all the designs were at place.

severinbeauvais:
This filing is blocked by a missing fee code (see ticket 24959) and some Legal API work. Extra work related to those can be done in a future ticket.

However, in this ticket, I'd like to see an updated JSON object on save.

Also, I'd like to see save and resume working. Since you'd only be saving a draft, Legal API should not be performing schema validation. Right now, as staff, I'm still getting a HTTP 401 (not authorized) when trying to save a draft.

src/views/NoticeOfWithdrawal.vue
severinbeauvais: Sample sent to me via chat. Looks good for now.
```
{
   "header": {
       "name": "noticeOfWithdrawal",
       "certifiedBy": "AQ",
       "email": "Aimee.Gao@gov.bc.ca",
       "date": "2024-12-16",
       "documentOptionalEmail": "test@test.ca",
       "bcolAccountNumber": "000000",
       "datNumber": "C1133435",
       "folioNumber": "123",
       "priority": false
   },
   "business": {
       "foundingDate": "2024-11-15T21:58:25.000+00:00",
       "identifier": "BC0883739",
       "legalName": "SEVERIN CONVERT ULC TO BC LIMITED CORP.",
       "legalType": "BC"
   },
   "noticeOfWithdrawal": {
       "filingId": "153004",
       "courtOrder": {
           "fileNumber": "TT0010045",
           "effectOfOrder": "planOfArrangement"
       },
       "planOfArrangement": {
             "partOfPoa": true,
            "comeIntoEffect": true
   }
}
```

severinbeauvais (approved):
LGTM. There are some blockers (other tickets) and there's a follow-on UI ticket to finish this.

eason-pan-bc (approved):
LTGM, like Sev mentioned, we'll need some follow-up tickets to finish it.

meawong (approved):

eason-pan-bc:
Just realize one issue in the payload for the filing. the `filingId` under `noticeOfWithdrawal` should be int instead of string.

Should be this:
```JSON
"noticeOfWithdrawal": {
            "filingId": 155118
        }
```

Not this:
```JSON
"noticeOfWithdrawal": {
            "filingId": "155118"
        }
```
Otherwise, you won't able to pass the schema check. https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/notice_of_withdrawal.json#L17C1-L18C39

ArwenQin:
I will update the filingId in the next ticket #24953

PR #725, opened 2025-03-05
26218 - Clicking on File Now invoking browser alert for a Court Order…

Rajandeep98:
… Filling.

*Issue #:* /bcgov/entity#26218

*Description of changes:*
Clicking on File Now invoking browser alert for a Court Order Filling.

### DIFF
--- a/src/views/CourtOrder.vue
+++ b/src/views/CourtOrder.vue
@@ -512,7 +512,7 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
 
     if (!this.isPageValid) {
       this.showErrors = true
-      this.saving=false
+      this.saving = false
       await this.validateAndScroll(this.validFlags, this.validComponents)
       return
     }
@@ -596,8 +596,10 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
       })
 
     this.saving = false
-    this.haveChanges = false
-    if (success) { this.navigateToBusinessDashboard(this.getIdentifier) }
+    if (success) {
+      this.haveChanges = false
+      this.navigateToBusinessDashboard(this.getIdentifier)
+    }
   }
 
   /**


severinbeauvais:
If I click File and Pay before entering any data, I get the expected validation errors, but also the File and Pay button stays in loading state:

image

Can you please fix this in this ticket/PR? Thanks!

src/views/CourtOrder.vue
```diff
@@ -595,6 +595,7 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
       })
 
     this.saving = false
+    this.haveChanges = false
```
severinbeauvais, line 598: If creating the filing fails then this flag is cleared, allowing the user to exit this page without actually saving and without a warning. Can this be moved somewhere better?

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-725-upzwp2up.web.app

src/views/CourtOrder.vue
```diff
@@ -512,6 +512,7 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
 
     if (!this.isPageValid) {
       this.showErrors = true
+      this.saving = false
```
severinbeauvais, line 515: So this fixes the bug of the button still spinning?

severinbeauvais (approved):
LGTM but get another review or 2 before merging.

src/views/CourtOrder.vue
severinbeauvais: Fixed, thanks.

src/views/CourtOrder.vue
Rajandeep98: yh loading is checking if page valid || saving

JazzarKarim (approved):

ketaki-deodhar (commented):
Merged now but looks good!

src/views/CourtOrder.vue
severinbeauvais: Got it. Thanks! Resolving.

PR #749, opened 2025-05-16
27186 refactor roles to permissions

TVWerdal:
*Issue #:*  bcgov/entity/27186 

*Description of changes:*
- Update to implement new permissions based authentication
   - Refactoring IsRoleStaff
   - Unit test updates for IsRoleStaff
- Deleted code related to dashboard (which is now a separate UI)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "7.5.7",
+  "version": "7.5.8",
   "private": true,
   "appName": "Business Filings UI",
   "sbcName": "SBC Common Components",
--- a/src/components/AnnualReport/AGMDate.vue
+++ b/src/components/AnnualReport/AGMDate.vue
@@ -40,7 +40,7 @@
             <v-date-picker
               v-model="datePicker"
               data-test-id="agm-date-picker"
-              :min="arMinDate"
+              :min="agmMinDate"
               :max="arMaxDate"
               no-title
               @input="menu=false"
@@ -178,6 +178,7 @@ export default class AgmDate extends Mixins(DateMixin) {
   @Getter(useBusinessStore) arMaxDate!: string
   @Getter(useBusinessStore) isEntityCoop!: boolean
   @Getter(useRootStore) getCurrentDate!: string
+  @Getter(useBusinessStore) getLastAnnualGeneralMeetingDate!: string
 
   // Local properties.
   dateText = '' // value in text field
@@ -225,6 +226,11 @@ export default class AgmDate extends Mixins(DateMixin) {
     // where Max AGM Date is 'today' in the AR Filing Year
     // up to Apr 30 in the next year
     return (this.compareYyyyMmDd(this.getCurrentDate, this.arMaxDate, '>'))
+  }
+
+  /** Whether to set Min AGM Date: cannot be earlier than an already reported AGM date. */
+  get agmMinDate (): string {
+    return this.latestYyyyMmDd(this.arMinDate, this.getLastAnnualGeneralMeetingDate)
   }
 
   /** Called when component is mounted. */
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-filings-ui",
-  "version": "7.5.7",
+  "version": "7.5.8",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-filings-ui",
-      "version": "7.5.7",
+      "version": "7.5.8",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/base-address": "2.0.46",
--- a/src/stores/businessStore.ts
+++ b/src/stores/businessStore.ts
@@ -22,6 +22,7 @@ export const useBusinessStore = defineStore('business', {
       hasRestrictions: null,
       identifier: null,
       lastAddressChangeDate: null,
+      lastAnnualGeneralMeetingDate: null,
       lastAnnualReportDate: null,
       lastDirectorChangeDate: null,
       legalName: null,
@@ -99,6 +100,11 @@ export const useBusinessStore = defineStore('business', {
     /** The last annual report change date. */
     getLastAnnualReportDate (state: BusinessStateIF): string {
       return state.businessInfo.lastAnnualReportDate
+    },
+
+    /** The last annual general meeting date. */
+    getLastAnnualGeneralMeetingDate (state: BusinessStateIF): string {
+      return state.businessInfo.lastAnnualGeneralMeetingDate
     },
 
     /** The last director change date. */


severinbeauvais (commented):
In case you were wondering what to do with the app version, I recommend increasing the major number for this!

src/components/EntityInfo.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
     id="entity-info"
-    :class=" { 'staff': isRoleStaff, 'hover': showHoverStyle }"
+    :class=" { 'staff': accessEntityMenu, 'hover': showHoverStyle }"
```
severinbeauvais, line 4: Class "staff" was a neat feature (well I thought so) that I implemented originally but which was never used. I think you can delete it, including the getter and styles on lines 134-138.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -326,6 +326,7 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
   // enums for template
   readonly axios = axios
   readonly AllowableActions = AllowableActions
+  readonly IsAuthorized = IsAuthorized
```
severinbeauvais, line 329: You don't this if it's not used in the template.

src/components/dialogs/NameRequestInvalidDialog.vue
```diff
@@ -92,19 +92,19 @@
...
 export default class NameRequestInvalidDialog extends Vue {
-  // Getter definition for static type checking.
-  @Getter(useRootStore) isRoleStaff!: boolean
+  /** For Template Contact Info message */
+  readonly IsAuthorized = IsAuthorized
+  readonly AuthorizedActions = AuthorizedActions
 
-  // Enum definition for use in template.
+  /** Enum definition for use in template. */
   NameRequestStates = NameRequestStates
```
severinbeauvais, line 108: While you're here, please make this one consistent with the other ones above.

src/components/dialogs/PaymentErrorDialog.vue
```diff
@@ -14,7 +14,7 @@
       <v-card-text id="dialog-text">
         <!-- display common message -->
         <div
-          v-if="!isRoleStaff"
+          v-if="!IsAuthorized(AuthorizedActions.NO_COMPLETING_PARTY_MESSAGE_BOX)"
```
severinbeauvais, line 17: I think this is an unrelated authorized action for this code.

Did you see what I did in Create UI for this error dialog?

src/components/dialogs/PaymentErrorDialog.vue
```diff
@@ -100,7 +100,7 @@
           text
           @click="exit()"
         >
-          {{ !isRoleStaff ? 'Back to My Dashboard' : 'OK' }}
+          {{ !IsAuthorized(AuthorizedActions.STAFF_BREADCRUMBS) ? 'Back to My Dashboard' : 'OK' }}
```
severinbeauvais, line 103: This authorized action makes sense but maybe we should rename it to describe what it does. Thoughts?

src/enums/authorizedActions.ts
```diff
@@ -0,0 +1,42 @@
...
+    OVERRIDE_NIGS,
+    REGISTRATION_FILING,
+    RESTORATION_FILING,
+    SAVE_DRAFT,
+    SBC_BREADCRUMBS,
+    STAFF_BREADCRUMBS,
+    STAFF_COMMENTS,
+    STAFF_PAYMENT,
+    THIRD_PARTY_CERTIFY_STMT
+  }
```
severinbeauvais, line 42: Please delete the ones you're not using in this UI.

src/enums/authorizedActions.ts
```diff
@@ -0,0 +1,42 @@
...
+    FIRM_CONVERSION_FILING,
+    FIRM_DISSOLUTION_FILING, // firm voluntaty dissolution
+    FIRM_EDITABLE_COMPLETING_PARTY,
+    FIRM_EDITABLE_EMAIL_ADDRESS,
+    FIRM_NO_MIN_START_DATE,
+    INCORPORATION_APPLICATION_FILING, // Coops and Corps
+    NO_COMPLETING_PARTY_MESSAGE_BOX,
+    NO_CONTACT_INFO,
+    NOTICE_WITHDRAWAL_FILING,
+    OTHER_DISSOLUTION_FILING, // non-firm voluntary dissolution
```
severinbeauvais, line 32: I renamed this VOLUNTARY_DISSOLUTION_FILING.

src/enums/authorizedActions.ts
```diff
@@ -0,0 +1,42 @@
...
+    COOP_CORRECTION_FILING,
+    CORP_CORRECTION_FILING,
+    COURT_ORDER_FILING,
+    COURT_ORDER_POA,
+    EDITABLE_CERTIFY_NAME,
+    EDITABLE_COMPLETING_PARTY,
+    FILE_AND_PAY,
+    FIRM_ADD_BUSINESS,
+    FIRM_CONVERSION_FILING,
+    FIRM_DISSOLUTION_FILING, // firm voluntaty dissolution
```
severinbeauvais, line 24: Typo.

src/enums/index.ts
```diff
@@ -20,3 +22,4 @@ export * from './warningTypes'
 
 // external enums
 export { CorpTypeCd, CorpClass } from '@bcrs-shared-components/corp-type-module'
+export { AccountTypes } from '@bcrs-shared-components/enums'
```
severinbeauvais, line 25: This enum should be obsolete now, right?

severinbeauvais: Oh, still used for now, same as in Edit UI:

image

src/interfaces/root-state-interface.ts
```diff
@@ -15,7 +15,8 @@ export interface StateFilingIF {
 /** The state model interface for the Root Store. */
 export interface RootStateIF {
   // tombstone data
-  authRoles: Array<string>
+  accountInformation: AccountInformationIF
+  authRoles: Array<AuthorizationRoles>
   currentDate: string // 'today' as YYYY-MM-DD in Pacific timezone
   currentJsDate: Date // 'now' as of dashboard loading in UTC
   keycloakRoles: Array<string>
```
severinbeauvais, line 22: Are you obsoleting Keycloak Roles in this ticket?

src/App.vue
```diff
@@ -110,21 +61,6 @@
...
-      <main v-if="isSigninRoute || dataLoaded">
-        <Breadcrumb :breadcrumbs="breadcrumbs" />
-        <EntityInfo
-          v-if="!isAmalgamationSelectionRoute"
-          @confirmDissolution="confirmDissolutionDialog = true"
-          @notInGoodStanding="nigsMessage = $event; notInGoodStandingDialog = true"
-          @downloadBusinessSummary="downloadBusinessSummary()"
-          @viewAddDigitalCredentials="viewAddDigitalCredentials()"
-        />
-        <router-view />
```
severinbeauvais, line 124: Since App.vue is the main page/container, don't we still need router-view to display the right filing page/view?

src/App.vue
```diff
@@ -540,6 +428,59 @@ export default class App extends Mixins(
     }
   }
 
+  /** Check authorizations. */
+  async checkAuth (): Promise<void> {
+    try {
+      // load account information
+      await this.loadAccountInformation().catch(error => {
```
severinbeauvais, line 435: I suggest you keep this separate (decouple) from checking auth. Example:

https://github.com/bcgov/business-edit-ui/blob/c3bd1aa35bc8f39a5f87a8f07d97585b372298a1/src/App.vue#L426

src/App.vue
```diff
@@ -540,6 +428,59 @@ export default class App extends Mixins(
...
+
+      // safety check
+      if (!this.businessId) {
+        throw new Error('Missing Business ID')
+      }
+      // check if current user is authorized
+      const response = await AuthServices.fetchAuthorizations(
+        this.getAuthApiUrl, this.businessId
+      )
+      this.storeAuthorizations(response) // throws if no role
```
severinbeauvais, line 448: Why do we need this as well as line 462? I think it's redundant.

src/App.vue
```diff
@@ -889,28 +696,6 @@ export default class App extends Mixins(
...
-      console.log('fetchDocument() error =', error)
-      this.downloadErrorDialog = true
-    })
-    this.setFetchingDataSpinner(false)
-  }
-
-  /** Direct to Digital Credentials. **/
-  viewAddDigitalCredentials (): void {
-    this.$router.push({ name: Routes.DIGITAL_CREDENTIALS })
-  }
```
severinbeauvais, line 913: Does the Business Dashboard UI have an action to route to the DBC pages in this UI?

severinbeauvais: Answered my own question -- yes, it does!

image

src/App.vue
severinbeauvais: Yeah, storeAuthorizations is doing the same thing as the code below. I think you can remove it.

src/interfaces/root-state-interface.ts
```diff
@@ -15,7 +15,8 @@ export interface StateFilingIF {
 /** The state model interface for the Root Store. */
 export interface RootStateIF {
   // tombstone data
-  authRoles: Array<string>
+  accountInformation: AccountInformationIF
```
severinbeauvais, line 18: If this is only used to identify SBC Staff users then I think this property (and all the code to fetch it, etc) will be obsolete soon. Just keep this in mind for when we update this UI to use auth roles directly.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -14,8 +14,9 @@ export default class AllowableActionsMixin extends Vue {
   @Getter(useBusinessStore) isEntityFirm!: boolean
   @Getter(useBusinessStore) isEntitySoleProp!: boolean
   @Getter(useBusinessStore) isGoodStanding!: boolean
-  @Getter(useRootStore) isRoleStaff!: boolean
 
+  readonly IsAuthorized = IsAuthorized
+  readonly AuthorizedActions = AuthorizedActions
```
severinbeauvais, line 19: I think you don't need these.

PS One impact of having this here is they will be available in all components that use this mixin, except you need to use `this.IsAuthorized()` instead of `IsAuthorized()`.

After removing these, you can remove "this." from lines 102 and 169 :)

src/mixins/allowable-actions-mixin.ts
```diff
@@ -165,7 +166,7 @@ export default class AllowableActionsMixin extends Vue {
       }
 
       case AllowableActions.STAFF_COMMENT: {
-        return (isBusiness && this.isRoleStaff)
+        return (isBusiness && this.IsAuthorized(AuthorizedActions.STAFF_COMMENTS))
```
severinbeauvais, line 169: We need separate authorized actions for staff comments on filings and staff comments on the business.

Ref:
image

src/routes.ts
```diff
@@ -21,7 +20,6 @@ export default [
   {
     path: '/',
```
severinbeauvais, line 21: What happens now if I navigate to this path? Maybe we can remove this route?

Also need to think about default/fallback route at the bottom of this file...

src/utils/authorizations.ts
```diff
@@ -0,0 +1,142 @@
...
+const DefaultRoles = [
+  AuthorizedActions.AMALGAMATION_FILING,
+  AuthorizedActions.CONTINUATION_IN_FILING,
+  AuthorizedActions.FILE_AND_PAY,
+  AuthorizedActions.FIRM_DISSOLUTION_FILING,
+  AuthorizedActions.INCORPORATION_APPLICATION_FILING,
+  AuthorizedActions.OTHER_DISSOLUTION_FILING,
+  AuthorizedActions.REGISTRATION_FILING,
+  AuthorizedActions.SAVE_DRAFT
+]
```
severinbeauvais, line 142: Should we add roles and checks for all filings?
- AGM Extension
- AGM Location Change
- Amalgamation Out
- etc

See [spreadsheet]( https://docs.google.com/spreadsheets/d/13P4ZA19lJfXJX95N5OfQMu6jYuotDH6h2hDf2DnAv0E/edit?gid=1874127980#gid=1874127980). Note that the roles (authorized actions) are still under discussion -- we will try to combine some of them, for example, staff-only filings. Ref: [28529](https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/28529).

src/views/AgmExtension.vue
```diff
@@ -288,7 +289,7 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin, FilingM
     this.data.incorporationDate = this.getFoundingDate
 
     // Pre-populate the certified block with the logged in user's name (if not staff)
-    if (!this.isRoleStaff && this.getUserInfo) {
+    if (!this.IsAuthorized(AuthorizedActions.THIRD_PARTY_CERTIFY_STMT) && this.getUserInfo) {
```
severinbeauvais, line 292: In Create UI and Edit UI, I used BLANK_CERTIFY_STATE for this.

PS Update the comments from "if staff" to "is authorized" (or whatever -- your choice).

src/views/AgmLocationChg.vue
```diff
@@ -417,7 +418,7 @@ export default class AgmLocationChg extends Mixins(CommonMixin, DateMixin, Filin
   /** Called when component is mounted. */
   mounted (): void {
     // Pre-populate the certified block with the logged in user's name (if not staff)
-    if (!this.isRoleStaff && this.getUserInfo) {
+    if (this.IsAuthorized(AuthorizedActions.THIRD_PARTY_CERTIFY_STMT) && this.getUserInfo) {
```
severinbeauvais, line 421: Same comment as above. And same for all filings below.

src/views/AnnualReport.vue
```diff
@@ -522,6 +521,10 @@ export default class AnnualReport extends Mixins(CommonMixin, DateMixin, FilingM
   saveErrors = []
   saveWarnings = []
 
+  // Enums for Template
+  readonly AuthorizedActions = AuthorizedActions
+  readonly IsAuthorized = IsAuthorized
```
severinbeauvais, line 526: Not needed if not used in template.

And you don't need to use "this." in the script, below.

src/views/AnnualReport.vue
```diff
@@ -1248,7 +1251,7 @@ export default class AnnualReport extends Mixins(CommonMixin, DateMixin, FilingM
       case SaveErrorReasons.FILE_PAY:
         // close the dialog and retry file-pay
         this.saveErrorReason = null
-        if (this.isRoleStaff) await this.onClickFilePay(true)
+        if (this.IsAuthorized(AuthorizedActions.STAFF_PAYMENT)) await this.onClickFilePay(true)
```
severinbeauvais, line 1254: Hmm, we need to use AuthorizedActions.FILE_AND_PAY in all filings, too.

src/views/CourtOrder.vue
```diff
@@ -318,6 +318,10 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
 
   paymentErrorDialog = false
 
+  // Template Enums
+  readonly AuthorizedActions = AuthorizedActions
+  readonly IsAuthorized = IsAuthorized
```
severinbeauvais, line 323: Not needed if not used in template.

I'll stop making this comment. Please check remaining files for same thing.

src/views/CourtOrder.vue
```diff
@@ -485,8 +489,7 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
   /** Called when component is created. */
   created (): void {
     // Safety check to make sure Staff is filing the Continuation Out.
-    if (!this.isRoleStaff) {
-      this.resumeErrorDialog = true
+    if (!this.IsAuthorized(AuthorizedActions.CONTINUATION_OUT_FILING)) {
```
severinbeauvais, line 492: Wrong filing.

tests/unit/ConsentAmalgamationOut.spec.ts
```diff
@@ -295,7 +296,7 @@ describe('Consent to Continue Out for general user and IAs only', () => {
     businessStore.setIdentifier('BC0007291')
     businessStore.setFoundingDate('1971-05-12T00:00:00-00:00')
     rootStore.filingData = []
-    rootStore.keycloakRoles = ['user']
+    rootStore.setAuthRoles([AuthorizationRoles.VIEW])
```
severinbeauvais, line 299: Yes, this is a better choice than `[]` as I did. I might update my code...

tests/unit/auth-services.spec.ts
```diff
@@ -24,9 +25,8 @@ describe('Auth Services', () => {
 
     // call method
     const response = await AuthServices.fetchAuthorizations('', 'CP1234567')
-
     // verify data
-    expect(response).toEqual({ data: AUTHORIZATIONS })
+    expect(response).toEqual({ roles: AUTHORIZATIONS.roles })
```
severinbeauvais, line 29: Or this?
`expect(response).toEqual(AUTHORIZATIONS)`

severinbeauvais:
> Why are a ton of files deleted?

Thayne and I thought we could delete the obsolete dashboard code. In retrospect, maybe that should have been a separate PR. Oh well. Thoughts?

src/components/common/OfficeAddresses.vue
```diff
@@ -300,8 +300,15 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
-      (this.inheritRegisteredAddress || (this.recDeliveryAddressValid && (this.inheritRecDeliveryAddress || this.recMailingAddressValid))))
+    return (
+      (this.deliveryAddressValid &&
+        (this.inheritDeliveryAddress || this.mailingAddressValid)
+      ) && (this.inheritRegisteredAddress ||
+       (this.recDeliveryAddressValid &&
+        (this.inheritRecDeliveryAddress || this.recMailingAddressValid)
+       )
+      )
+    )
```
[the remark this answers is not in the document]
TVWerdal: Yeah, I mainly folded it because the linter was picking it up as too long - I was trying to find the most 'readable' way to present it, lots of complex conditionals though.

src/components/dialogs/PaymentErrorDialog.vue
TVWerdal: This dialogue does not appear to have any restrictions in the create-ui, should I remove it here as well? Or change it to something more related like STAFF_PAYMENT? I'm not entirely sure why I used this particular permission originally, doesn't make a lot of sense.

src/components/dialogs/PaymentErrorDialog.vue
TVWerdal: Yeah, I had a note in the permissions spread sheet saying "Used this permission as only navigation related perm. If a more specific permission is created (STAFF_ACTION?) is created this should be updated." 

Maybe something like 'STAFF_NO_REDIRECT'?

JazzarKarim:
> > Why are a ton of files deleted?
> 
> Thayne and I thought we could delete the obsolete dashboard code. In retrospect, maybe that should have been a separate PR. Oh well. Thoughts?

Yep, I think it would've better if we had a separate PR and even separate techdebt ticket to clean up the old dashboard code from here (something like https://github.com/bcgov/entity/issues/27158). It's all good though, no problem.

src/enums/authorizedActions.ts
TVWerdal: Not used, will be removed in next commit

src/enums/authorizedActions.ts
TVWerdal: Not used, will be removed in next commit

src/interfaces/root-state-interface.ts
TVWerdal: I didn't focus too much on that - it wasn't called out in the ticket itself. Should I be doing that?

src/components/EntityInfo.vue
```diff
@@ -1,7 +1,6 @@
 <template>
   <div
     id="entity-info"
-    :class=" { 'staff': isRoleStaff, 'hover': showHoverStyle }"
```
severinbeauvais, line 4: Since you're getting rid of the hover class, you can delete the property on lines 64-65 and the style on lines 124-126.

src/components/EntityInfo.vue
```diff
@@ -59,11 +57,10 @@ import { Routes, DCRoutes } from '@/enums/routes'
   }
 })
 export default class EntityInfo extends Vue {
-  @Getter(useRootStore) isRoleStaff!: boolean
-
   /** Whether to show the hover style. */
   showHoverStyle = false
-
+  accessEntityMenu = IsAuthorized(AuthorizedActions.ADMIN_DISSOLUTION_FILING) &&
+    IsAuthorized(AuthorizedActions.OVERRIDE_NIGS)
```
severinbeauvais, line 63: What is this used for?

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -124,7 +127,10 @@
         <v-list dense>
           <v-list-item-group color="primary">
             <!-- View/Add Digital Credentials -->
-            <span v-if="isAllowed(AllowableActions.DIGITAL_CREDENTIALS)">
+            <span
+              v-if="isAllowed(AllowableActions.DIGITAL_CREDENTIALS) &&
+                IsAuthorized(AuthorizedActions.DIGITAL_CREDENTIALS)"
+            >
```
severinbeauvais, line 133: This is good. I prefer keeping these checks separate instead of moving IsAuthorized into the isAllowed function.

FYI @bobdev-94

src/components/dialogs/PaymentErrorDialog.vue
severinbeauvais: Here's the Create UI version of this dialog. I think you can follow the same logic.
https://github.com/bcgov/business-create-ui/blob/main/src/dialogs/PaymentErrorDialog.vue

src/enums/authorizedActions.ts
severinbeauvais: Can you also alphabetize these to make them easier for a human to find? Thanks!

src/enums/index.ts
severinbeauvais: 27536 is implemented. You can now update this code (and get rid of AccountTypes).

tests/unit/EntityMenu.spec.ts
```diff
@@ -81,47 +82,12 @@ describe('Entity Menu - entities', () => {
     wrapper.destroy()
   })
 
-  it('displays entity info properly for a draft IA', async () => {
```
severinbeauvais, line 84: Right. It's no longer Filings UI who displays the dashboard with the special EntityInfo component for draft businesses.

tests/unit/auth-services.spec.ts
```diff
@@ -24,9 +25,8 @@ describe('Auth Services', () => {
 
     // call method
     const response = await AuthServices.fetchAuthorizations('', 'CP1234567')
```
severinbeauvais, line 27: Do we still need fetchAuthorizations?

src/components/AnnualReport/AGMDate.vue
```diff
@@ -54,15 +54,18 @@
...
-            <span v-if="!allowCoa && !allowCod">
+            <span
+              v-if="!allowCoa && !IsAuthorized(AuthorizedActions.ADDRESS_CHANGE_FILING) &&
+                !allowCod && !IsAuthorized(AuthorizedActions.DIRECTOR_CHANGE_FILING)"
+            >
               You cannot change your Registered Office Addresses or Directors in this Annual Report
               because your AGM predates another filing that may have conflicting changes.
             </span>
-            <span v-else-if="!allowCoa">
+            <span v-else-if="!allowCoa && !IsAuthorized(AuthorizedActions.ADDRESS_CHANGE_FILING)">
```
severinbeauvais, line 64: This is a tricky one and you've probably thought a lot about it...

Do we actually need these checks here? This is a sub-component and the checks for the parent views should suffice, I think?

I'm also trying to wrap my head around what this check means. "If you're not allowed to change addresses and you're not allowed to file an Address Change filing." Seems redundant...?

Do you have a use case where this is needed?

src/App.vue
```diff
@@ -52,48 +45,6 @@
       attach="#app"
       @close="notInGoodStandingDialog = false"
     />
-
-    <!-- Loading Dashboard spinner -->
-    <v-fade-transition>
-      <div
-        v-show="showLoadingDashboardSpinner"
```
severinbeauvais, line 59: You can delete this getter.

And also "spinnerText" below.

And any getters they call if no longer used.

src/App.vue
severinbeauvais: Reverted.

src/components/dialogs/PaymentErrorDialog.vue
```diff
@@ -109,16 +106,17 @@
...
-import { useRootStore } from '@/stores'
+import { IsAuthorized } from '@/utils'
 
 @Component({
   components: { ContactInfo }
 })
 export default class PaymentErrorDialog extends Vue {
-  // Getter to check if logged in user is Staff.
-  @Getter(useRootStore) isRoleStaff!: boolean
+  /** For Template Contact Info / message */
```
JazzarKarim, line 117: I wouldn't use JSDoc comments on inline variables. Also, there's an extra / typo here. This was done in a lot of places. If it's too much work to go to `//`, leave as is. We usually use JSDoc comments on functions and API calls for example.

src/enums/authorizedActions.ts
```diff
@@ -5,32 +5,32 @@ export enum AuthorizedActions {
     ADMIN_DISSOLUTION_FILING,
     AGM_CHG_LOCATION_FILING,
     AGM_EXTENSION_FILING,
+    ALTERATION_FILING,
```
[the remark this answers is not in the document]
severinbeauvais: I think the following aren't needed either:
- CONTINUATION_IN_FILING
- CORRECTION_FILING
- SPECIAL_RESOLUTION_FILING
- VOLUNTARY_DISSOLUTION_FILING (part of Entity Menu)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-filings-ui",
-  "version": "7.5.6",
+  "version": "7.6.0",
```
severinbeauvais, line 3: I suggest 8.0.0 since the removal of the dashboard stuff is a major change.

src/components/dialogs/PaymentErrorDialog.vue
JazzarKarim: Thanks for fixing this! Resolving.

src/interfaces/root-state-interface.ts
severinbeauvais: Unresolving.

Why is there both auth roles and keycloak roles -- are they both being used?

src/stores/rootStore.ts
```diff
@@ -435,15 +384,15 @@ export const useRootStore = defineStore('root', {
       this.stateFiling = stateFilingResponse
     },
 
+    setNoRedirect (val: boolean) {
+      this.noRedirect = val
+    },
```
severinbeauvais, line 389: I think you can delete this code since there is no longer a dashboard to display here. This was to prevent redirection to the Business Dashboard UI.

src/views/ConsentContinuationOut.vue
```diff
@@ -321,10 +321,11 @@ export default class ConsentContinuationOut extends Mixins(CommonMixin, DateMixi
...
   @Getter(useRootStore) getUserInfo!: any
-  @Getter(useRootStore) isRoleStaff!: boolean
 
   // enum for template
   readonly FilingCodes = FilingCodes
+  readonly AuthorizedActions = AuthorizedActions
+  readonly IsAuthorized = IsAuthorized
 
   // variables for Certify component
   certifiedBy = ''
```
severinbeauvais, line 331: Should there be a check in created(), below, that the current filing type is authorized? I'm thinking yes, just to prevent smart users from trying to go straight to that route (thus bypassing the dashboard).

severinbeauvais: I see you did this in some filings. I think this applies to all filings.

src/views/CourtOrder.vue
```diff
@@ -484,9 +487,8 @@ export default class CourtOrderView extends Mixins(DateMixin, FilingMixin, Commo
   }
   /** Called when component is created. */
   created (): void {
-    // Safety check to make sure Staff is filing the Continuation Out.
-    if (!this.isRoleStaff) {
-      this.resumeErrorDialog = true
+    // Safety check to make sure user has required permissions to file the Court Order.
+    if (!this.IsAuthorized(AuthorizedActions.STAFF_FILINGS)) {
```
severinbeauvais, line 491: There's a separate authorized action for this one: COURT_ORDER_FILING.

src/enums/index.ts
severinbeauvais: I see isSbcFieldOfficeStaff is updated and the subject enum is still used to type a property in an interface. Looks like nothing more is needed. Resolving.

src/App.vue
```diff
@@ -562,18 +474,8 @@ export default class App extends Mixins(
...
-  /** Fetches Keycloak roles from JWT. */
-  fetchKeycloakRoles (jwt: any): Array<string> {
-    const keycloakRoles = jwt.roles
-    if (keycloakRoles && keycloakRoles.length > 0) {
-      return keycloakRoles
-    }
-    throw new Error('Error getting Keycloak roles')
-  }
-
   storeAuthorizations (response: any): void {
```
severinbeauvais, line 477: I believe storeAuthorizations is no longer used and can be deleted.

src/mixins/allowable-actions-mixin.ts
severinbeauvais: Oops, we no longer handle detail comments on filings in this UI. Resolving.

src/routes.ts
severinbeauvais: I think we will have a loop here if the user goes to route `/`, or an invalid route which also redirects to `/`.

One possibility is to create an error page if the route is invalid (eg, default/fallback route).

Another possibility is to redirect the user to the business' dashboard.

src/views/CourtOrder.vue
severinbeauvais: Still the wrong filing but I added another comment for that so I'm resolving this one.

src/mixins/allowable-actions-mixin.ts
```diff
@@ -98,7 +97,7 @@ export default class AllowableActionsMixin extends Vue {
       }
 
       case AllowableActions.DETAIL_COMMENT: {
-        return (isBusiness && this.isRoleStaff)
+        return (isBusiness && IsAuthorized(AuthorizedActions.DETAIL_COMMENTS))
```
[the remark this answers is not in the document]
severinbeauvais: Actually, detail comments aren't even in this UI any more, so this can be deleted.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -175,7 +181,8 @@
                   <v-list-item
                     v-if="showConsentAmalgamateOut"
                     id="consent-amalgamate-out-list-item"
-                    :disabled="!isAllowed(AllowableActions.CONSENT_AMALGAMATION_OUT)"
+                    :disabled="!isAllowed(AllowableActions.CONSENT_AMALGAMATION_OUT) &&
+                      !IsAuthorized(AuthorizedActions.CONSENT_AMALGAMATION_OUT_FILING)"
```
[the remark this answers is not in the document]
severinbeauvais: @bobdev-94 You hide unauthorized actions in the dasboard, right?

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -336,7 +349,10 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
    * More actions should only display if an entity is not historical or if the Digital Credentials
    * feature is available, otherwise an empty menu will display for historical entities. */
   get areMoreActionsAvailable (): boolean {
-    return !this.isHistorical || this.isAllowed(AllowableActions.DIGITAL_CREDENTIALS)
+    return (!this.isHistorical ||
+      (this.isAllowed(AllowableActions.DIGITAL_CREDENTIALS) &&
+      this.IsAuthorized(AuthorizedActions.DIGITAL_CREDENTIALS))
+    )
```
severinbeauvais, line 355: This should check the other actions in this More Actions menu.

src/components/EntityInfo/EntityMenu.vue
```diff
@@ -367,7 +383,7 @@ export default class EntityMenu extends Mixins(AllowableActionsMixin) {
 
   /** The tooltip text for AGM Extension list item. Text is different if action item is disabled. */
   get agmExtensionToolTipText (): string {
-    if (!this.isAllowed(AllowableActions.AGM_EXTENSION)) {
+    if (!this.isAllowed(AllowableActions.AGM_EXTENSION) && !this.IsAuthorized(AuthorizedActions.AGM_EXTENSION_FILING)) {
```
severinbeauvais, line 386: You won't need this if the action is hidden when not authorized.

Ditto below.

src/components/EntityInfo/EntityMenu.vue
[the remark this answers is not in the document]
severinbeauvais: Except Staff Comments, which I don't actually see in Dev.

src/components/AnnualReport/AGMDate.vue
severinbeauvais: So, is this already handled in the Annual Report?

image

src/interfaces/root-state-interface.ts
severinbeauvais: OK, I see we're looking at Keycloak roles but storing it in authRoles array -- that's fine as it's only temporary. Resolving.

src/stores/rootStore.ts
```diff
@@ -24,7 +24,6 @@ export const useRootStore = defineStore('root', {
     currentJsDate: null,
     bootstrapFilingStatus: null,
     bootstrapFilingType: null,
-    keycloakRoles: [],
     noRedirect: false,
```
severinbeauvais, line 27: Here.

src/stores/rootStore.ts
```diff
@@ -224,33 +200,6 @@ export const useRootStore = defineStore('root', {
...
-      return (
-        this.isAmalgamationFiling ||
-        this.isContinuationInFiling ||
-        this.isIncorporationApplicationFiling ||
-        this.isRegistrationFiling
-      )
-    },
-
     /** The Name Request (may be null). */
     getNameRequest (state: RootStateIF): any {
```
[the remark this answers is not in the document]
severinbeauvais: What about in App.vue where we store the NR data?

src/stores/rootStore.ts
severinbeauvais: And also line 39 (and the interface for this data).

tests/unit/AgmExtension.spec.ts
```diff
@@ -150,7 +152,7 @@ describe('AGM Extension view', () => {
 
     // simulate valid certify data and component
     wrapper.setData({ certifiedBy: 'Full Name', isCertified: true, certifyFormValid: true })
-
+    await utils.sleep(300) // need to wait for debounce
```
severinbeauvais, line 155: Does `await flushPromises()` work here instead?

.env.example
```diff
@@ -14,6 +14,7 @@ VUE_APP_SITEMINDER_LOGOUT_URL="https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi"
 VUE_APP_BUSINESS_CREATE_URL="https://dev.create.business.bcregistry.gov.bc.ca/"
 VUE_APP_BUSINESS_EDIT_URL="https://dev.edit.business.bcregistry.gov.bc.ca/"
 VUE_APP_BUSINESS_DASH_URL="https://dev.business-dashboard.bcregistry.gov.bc.ca/"
+VUE_APP_BUSINESS_REG_DASH_URL="https://dev.business-registry-dashboard.bcregistry.gov.bc.ca/"
```
severinbeauvais, line 17: You also need to update https://github.com/bcgov/business-filings-ui/blob/main/devops/vaults.env so that the build/deploy process knows where to get this value from 1Password.

Hint:
- vault is `web-url`
- section is `business-dash`
- key is `BUSINESS_DASH_URL`

src/stores/configurationStore.ts
```diff
@@ -54,6 +58,7 @@ export const useConfigurationStore = defineStore('configuration', {
 
     getLegalApiUrl (state: ConfigurationStateIF): string {
       const configuration = state.configuration
+      console.log('configuration', configuration)
```
severinbeauvais, line 61: Remove debugging code (also line 147 below).

src/components/dialogs/PaymentErrorDialog.vue
severinbeauvais: I see you've changed this to STAFF_PAYMENT now. My preference would be to just show the buttons the same to all users, but this is OK in a pinch.

src/utils/set-base-route-and-business-id.ts
```diff
@@ -22,6 +23,9 @@ export function setBaseRouteAndBusinessId (pathname: string, processEnvBaseUrl:
     sessionStorage.setItem('TEMP_REG_NUMBER', id)
     // ensure we don't already have a Business ID in scope
     sessionStorage.removeItem('BUSINESS_ID')
+    // If user hits just root with no other info, they are sent back to BRD
+  } else if (pathname === '/' || pathname === processEnvBaseUrl) {
+    return
```
severinbeauvais, line 28: Why is this needed? Does line 30 below not handle that acceptably?

src/views/AgmExtension.vue
```diff
@@ -260,6 +272,11 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin, FilingM
 
   /** Called when component is created. */
   created (): void {
+    if (!IsAuthorized(AuthorizedActions.AGM_EXTENSION_FILING)) {
+      // user is not authorized to access AGM extensions, so route to dashboard
+      this.authErrorDialog = true
+      throw new Error('You are not authorized to complete this action.')
```
severinbeauvais, line 278: Do we need both a dialog and a console error? Could you just return here to pre-empt the rest of created()?

src/views/Dashboard.vue
severinbeauvais, line 1: I'd be tempted to keep this dummy view but rename it Default or Redirect or something to match its purpose.

src/components/EntityInfo/EntityMenu.vue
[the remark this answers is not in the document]
severinbeauvais: Please look into this: your code includes the EntityMenu, which includes the Staff Comments, but in Dev when I go to a filing such as COD, I don't see the staff comments.

Dashboard: https://dev.business-dashboard.bcregistry.gov.bc.ca/BC0888143/?accountid=3040
COD: https://dev.business.bcregistry.gov.bc.ca/BC0888143/standalone-directors?accountid=3040&businessid=BC0888143&filingId=0

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-749-g6i2jqys.web.app

.env.example
severinbeauvais: Fixed, thanks.

src/components/EntityInfo.vue
```diff
@@ -43,27 +42,18 @@
 
 <script lang="ts">
 import { Component, Emit, Vue } from 'vue-property-decorator'
-import { Getter } from 'pinia-class'
 import { NigsMessage } from '@/enums'
 import EntityDefinitions from './EntityInfo/EntityDefinitions.vue'
 import EntityHeader from './EntityInfo/EntityHeader.vue'
-import EntityMenu from './EntityInfo/EntityMenu.vue'
```
severinbeauvais, line 50: You need to remove the component from the template, too (line 20).

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: No more EntityMenu, so this comment is obsolete. Resolving.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Obsolete. Resolving.

src/components/dialogs/AddCommentDialog.vue
```diff
@@ -28,7 +28,7 @@
             id="dialog-save-button"
             text
             color="primary"
-            :disabled="!detailCommentValid || saving"
+            :disabled="!detailCommentValid || saving || !IsAuthorized(AuthorizedActions.DETAIL_COMMENTS)"
```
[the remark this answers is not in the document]
severinbeauvais: Deleted. Resolving.

src/views/CourtOrder.vue
severinbeauvais: Fixed. Resolving.

src/routes.ts
severinbeauvais: Resolving as I will start a new conversation.

src/routes.ts
```diff
@@ -16,15 +15,14 @@ import Signout from '@/views/auth/Signout.vue'
 import { DigitalCredentialRoutes } from '@/resources/DigitalCredentialRoutes'
 import { Routes } from '@/enums'
 import { FilingNames } from '@bcrs-shared-components/enums'
+import Dashboard from './views/Dashboard.vue'
 
 export default [
   {
+    // Return to BRD if the URL is hit directly
     path: '/',
```
severinbeauvais, line 23: An example of this route is https://dev.business.bcregistry.gov.bc.ca/CP1049523/, right?

Why not let the default/fallback route handle this? (see also my next comment)

src/routes.ts
```diff
@@ -238,18 +236,11 @@ export default [
...
-  //   // ref: https://cli.vuejs.org/guide/html-and-static-assets.html#prefetch
-  //   component: () => import(/* webpackChunkName: "about" */ './views/About.vue')
-  // },
-  {
-    // default/fallback route
-    path: '*',
-    redirect: '/'
+  {
+    // Return to new dashboard if we have a businessID, but no actual filing info
+    path: '/:businessId/',
```
severinbeauvais, line 241: Does this ever get used? I ask because the business id stripped out of the route (VUE_ROUTER_BASE) in set-base-route-and-business-id.ts. In other words, the router (that uses this file) should only ever see what's after the business id, ex: "standalone-directors?filingId=0".

Meanwhile, we should still have a default/fallback route to handle such cases as "https://dev.business.bcregistry.gov.bc.ca/CP1049523/not-a-valid-filing".

I know you've investigated this a bunch so let me know what you found.

src/components/EntityInfo/EntityMenu.vue
severinbeauvais: Obsolete.

src/mixins/allowable-actions-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Obsolete.

src/utils/set-base-route-and-business-id.ts
```diff
@@ -22,8 +22,12 @@ export function setBaseRouteAndBusinessId (pathname: string, processEnvBaseUrl:
     sessionStorage.setItem('TEMP_REG_NUMBER', id)
     // ensure we don't already have a Business ID in scope
     sessionStorage.removeItem('BUSINESS_ID')
+    // If user hits just root with no other info, they are sent back to BRD
   } else {
-    throw new Error('Missing or invalid Business ID or Temporary Registration Number.')
+    // Make sure we don't have anything already in scope
+    sessionStorage.removeItem('BUSINESS_ID')
+    sessionStorage.removeItem('TEMP_REG_NUMBER')
+    return
```
severinbeauvais, line 30: OK. Thanks for the comment.
Please add another comment indicating how/where this condition is handled.

src/main.ts
```diff
@@ -52,6 +52,14 @@ async function start () {
...
   setBaseRouteAndBusinessId(windowLocationPathname, processEnvBaseUrl, windowLocationOrigin) // may throw an error
 
+  // Saftey check for proper business number
+  const businessId = sessionStorage.getItem('BUSINESS_ID')
+  console.log(`Business ID: ${businessId}`)
+  if (!businessId) {
+    const businessRegistryDashboardUrl = configurationStore.getBusinessRegistryDashboardUrl
+    navigate(businessRegistryDashboardUrl)
+    return
+  }
```
severinbeauvais, line 62: 1. Would it make sense to move this into setBaseRouteAndBusinessId(), or lines 137+ below?
2. What about checking for valid T number?
3. Previously, setBaseRouteAndBusinessId threw an error, which was caught and handled on lines 137+ below. Is that still useful?
4. Is the console log for debugging? Please remove it.

src/routes.ts
```diff
@@ -16,16 +15,9 @@ import Signout from '@/views/auth/Signout.vue'
 import { DigitalCredentialRoutes } from '@/resources/DigitalCredentialRoutes'
 import { Routes } from '@/enums'
 import { FilingNames } from '@bcrs-shared-components/enums'
+import Default from './views/Default.vue'
```
severinbeauvais, line 18: Use `@` format like the other imported views.

src/routes.ts
```diff
@@ -238,18 +230,11 @@ export default [
...
   DigitalCredentialRoutes,
-  // {
-  //   path: '/about',
-  //   name: 'about',
-  //   // route level code-splitting
-  //   // this generates a separate chunk (about.[hash].js) for this route
-  //   // which is lazy-loaded when the route is visited.
-  //   // ref: https://cli.vuejs.org/guide/html-and-static-assets.html#prefetch
-  //   component: () => import(/* webpackChunkName: "about" */ './views/About.vue')
-  // },
```
severinbeauvais, line 249: I'm OK with deleting this. It was an idea way back when for lazy loading components using Webpack. This is now obsolete.

src/routes.ts
severinbeauvais: Looks good. Resolving.

src/views/AnnualReport.vue
```diff
@@ -170,7 +178,7 @@
                   <h2 id="directors-header">
                     3. Directors
                   </h2>
-                  <p v-if="allowChange('cod')">
+                  <p v-if="allowChange('cod') && IsAuthorized(AuthorizedActions.DIRECTOR_CHANGE_FILING)">
```
severinbeauvais, line 181: Whoops, I just noticed this.

Please remove this check (and also the check on lines 139 and 140). This authorized action is for Director Change filings on their own. When director changes or address changes are part of an AR, they're authorized (if the AR is authorized).

src/components/AnnualReport/AGMDate.vue
severinbeauvais: (Actually, I think these checks aren't needed for an Annual Report. See my other comment.)

src/enums/authorizedActions.ts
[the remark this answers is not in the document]
severinbeauvais: Looks OK now. Resolving.

severinbeauvais:
I'm caught up on the latest code. There are a few comments still left.

Please add a "Test Notes" comment in the ticket explaining what should be tested -- basically, all filings, and at least a regular user and REGI staff and looking at filings/components authorized or not.

src/components/AnnualReport/AGMDate.vue
severinbeauvais: Checks are removed. Resolving.

src/views/AnnualReport.vue
severinbeauvais: Checks are removed. Resolving.

severinbeauvais:
There are still a few unresolved conversations. (They may be collapsed above.)

TVWerdal:
> I'm caught up on the latest code. There are a few comments still left.
> 
> Please add a "Test Notes" comment in the ticket explaining what should be tested -- basically, all filings, and at least a regular user and REGI staff and looking at filings/components authorized or not.

This has been added here: https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/27186

Let me know if it looks like I have missed anything. Going to go over it myself again top to bottom - there have been a number of changes since some of these were last tested.

TVWerdal:
> There are still a few unresolved conversations. (They may be collapsed above.)

I think I have addressed them all now - let me know if I am missing anything.

src/views/AgmExtension.vue
```diff
@@ -260,6 +272,11 @@ export default class AgmExtension extends Mixins(CommonMixin, DateMixin, FilingM
 
   /** Called when component is created. */
   created (): void {
+    if (!IsAuthorized(AuthorizedActions.AGM_EXTENSION_FILING)) {
+      // user is not authorized to access AGM extensions, so route to dashboard
+      this.authErrorDialog = true
```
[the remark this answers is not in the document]
severinbeauvais: Ping.

src/utils/set-base-route-and-business-id.ts
severinbeauvais: I was looking for a code comment 😄 

Now that I think about it, bootstrap businesses (Txxx) never do anything in Filings UI, right? So that's a bit more code that can be cleaned up.

src/main.ts
severinbeauvais: I'm OK with keeping the code here.

What do you think about throwing an error here and letting the catch block on lines 136+ handle the navigation?

You need the import to call the store getter, right? What if you only check for a BUSINESS_ID and then throw an error? Now your code could go back into set-base-route-and-business-id...?

severinbeauvais:
Last round of reviews from me, I think, but now you have some code conflicts (sorry).

src/components/EntityInfo/EntityMenu.vue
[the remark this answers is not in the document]
TVWerdal: Okay, so should I return the staff comment stuff to the code as well?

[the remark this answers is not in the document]
severinbeauvais: 1. We don't need the entire menu component. For comparison, Create UI just puts everything in the EntityInfo component.

2. I'm OK either way. If you don't revert some of that menu code the the next ticket can do it.

Let me know what you're planning and then I'll resolve this conversation.

PR #760, opened 2025-06-24
28541 - Do not Prepopulate Folio Number field

meawong:
*Issue #:* /bcgov/entity#28541

*Description of changes:*
- Updated requirement: Do not prepopulate the account folio number (we will restore the TFN when resuming a draft filing, when applicable)

### DIFF
--- a/src/components/common/TransactionalFolioNumber.vue
+++ b/src/components/common/TransactionalFolioNumber.vue
@@ -51,9 +51,6 @@ export default class TransactionalFolioNumber extends Vue {
     folioNumberInput: any
   }
 
-  /** Account Folio Number prop. */
-  @Prop({ default: null }) readonly accountFolioNumber!: string
-
   /** Transactional Folio Number prop. */
   @Prop({ default: null }) readonly transactionalFolioNumber!: string
 
--- a/src/components/common/TransactionalFolioNumber.vue
+++ b/src/components/common/TransactionalFolioNumber.vue
@@ -62,8 +62,8 @@ export default class TransactionalFolioNumber extends Vue {
 
   /** Called when component is mounted. */
   mounted (): void {
-    // restore transactional FN if it exists, otherwise use account FN
-    this.localFolioNumber = this.transactionalFolioNumber || this.accountFolioNumber
+    // restore transactional FN from draft if it exists, otherwise leave field empty
+    this.localFolioNumber = this.transactionalFolioNumber || ''
     // Emit initial validity
     this.emitValid(this.isValid)
   }
--- a/src/views/AnnualReport.vue
+++ b/src/views/AnnualReport.vue
@@ -736,7 +736,9 @@ export default class AnnualReport extends Mixins(CommonMixin, DateMixin, FilingM
       this.certifiedBy = header.certifiedBy
 
       // restore Transactional Folio Number
-      if (filing.header.folioNumber) this.setTransactionalFolioNumber(filing.header.folioNumber)
+      if (filing.header.isTransactionalFolioNumber && filing.header.folioNumber) {
+        this.setTransactionalFolioNumber(filing.header.folioNumber)
+      }
 
       // restore Staff Payment data
       if (header.routingSlipNumber) {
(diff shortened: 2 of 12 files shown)


bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-760-ljhmgx31.web.app

JazzarKarim (approved):
LGTM 👍

ArwenQin (approved):
LGTM

src/components/common/TransactionalFolioNumber.vue
```diff
@@ -62,8 +62,8 @@ export default class TransactionalFolioNumber extends Vue {
 
   /** Called when component is mounted. */
   mounted (): void {
-    // restore transactional FN if it exists, otherwise use account FN
-    this.localFolioNumber = this.transactionalFolioNumber || this.accountFolioNumber
```
severinbeauvais, line 66: Is `accountFolioNumber` now obsolete? Please remove prop from here and parent components.

src/components/common/TransactionalFolioNumber.vue
```diff
@@ -62,8 +62,8 @@ export default class TransactionalFolioNumber extends Vue {
...
+    // restore transactional FN from draft if it exists, otherwise leave field empty
+    this.localFolioNumber = this.transactionalFolioNumber || ''
```
severinbeauvais, line 66: With this, the business folio number will never be used in the filing, right?

I think we wanted this to fall back to the BFN if the user didn't enter a TFN or clear it out. Do you have more information on this?

severinbeauvais: I see the requirements in bcgov/entity#28739.

However, I also see a conflicting requirement:

image

In a previous ticket, we determined that the API would not populate the business folio number if a transactional folio number was provided. So I don't see how this will work now.

Perhaps we only need to go over the requirements with Mihai to explain that the filing itself has to contain a BFN, a TFN, or an empty field, and therefore we must prepopulate the BFN here.

cc; @Mihai-QuickSilverDev

meawong: > Also, do ALL filings have the same code to use TFN if available else use BFN if available else use nothing?

Yes, I implemented this for all filings in Filings UI. I can also replicate this in Create UI and Edit UI.

severinbeauvais: @Mihai-QuickSilverDev Do you want the ability to override the business folio number with "blank"?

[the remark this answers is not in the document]
meawong: If a TFN was set in the draft, that value will be restored in the TFN component after resuming:

image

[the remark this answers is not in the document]
severinbeauvais: If no TFN was set in the draft but there's a business (account) FN, what does the TFN component look like?

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-760-ljhmgx31.web.app

src/views/AnnualReport.vue
```diff
@@ -737,7 +736,9 @@ export default class AnnualReport extends Mixins(CommonMixin, DateMixin, FilingM
       this.certifiedBy = header.certifiedBy
 
       // restore Transactional Folio Number
-      if (filing.header.folioNumber) this.setTransactionalFolioNumber(filing.header.folioNumber)
+      if (filing.header.isTransactionalFolioNumber && filing.header.folioNumber) {
```
meawong, line 739: I added a new flag `isTransactionalFolioNumber` (ref from Create UI: https://github.com/bcgov/business-create-ui/blob/e1f18fe6daac1a43e2507328a890f44cfac02cab/src/mixins/filing-template-mixin.ts#L1053)

This will make sure the BFN isn't restored when resuming a draft if no TFN was set (i.e. the field will be empty).

severinbeauvais (approved):
Even though AgmExtension and AgmLocationChg do not suport resuming a draft, what do you about adding the `isTransactionalFolioNumber` flag to them as well? (Just in case.)

src/components/common/TransactionalFolioNumber.vue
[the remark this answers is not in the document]
severinbeauvais: Fixed, thanks!

severinbeauvais: OK. Resolving.

PR #774, opened 2025-08-21
30248 update API error parsing

bobdev-94:
*Issue #:* /bcgov/entity#30248

*Description of changes:*

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -470,7 +470,7 @@ export default class App extends Mixins(
     ).catch(error => {
       if (error.response.status === 404 &&
         error.response.config.url.includes('addresses') &&
-        error.response?.data?.rootCause?.message.includes('address not found')) hasBAError = true
+        error.response.data?.rootCause?.message.includes('address not found')) hasBAError = true
     })
 
     if (data) {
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -470,7 +470,7 @@ export default class App extends Mixins(
     ).catch(error => {
       if (error.response.status === 404 &&
         error.response.config.url.includes('addresses') &&
-        error.response.data.message.includes('address not found')) hasBAError = true
+        error.response?.data?.rootCause?.message.includes('address not found')) hasBAError = true
```
severinbeauvais, line 473: We need to be careful when using conditional chaining because it can mask errors that we want to see.

Here, I think `response` should not use conditional chaining, because if response is falsy then lines 471 and 472 will fail also.

Is there a specific reason you added all this conditional chaining? Should `message` also use it?

bobdev-94: There is no specific reason. But to avoid `no props` error, I think it's better debug later

severinbeauvais: Please believe me when I say conditional chaining can mask errors and that's not a good thing.

I am OK with using conditional chaining here on `rootCause` and `message`. The objects before that should always be declared since Axios creates them -- they do not come from the API.

meawong (approved):
LGTM

PR #784, opened 2026-01-13
31795 - Add Canadian Postal Code Validation to COA/COD

meawong:
*Issue #:* /bcgov/entity#31795 

*Description of changes:*
- Imported BaseAddress/mixins components for missing validation updates
- Added custom postal code validation for Canadian addresses (referenced from [Create UI](https://github.com/bcgov/business-create-ui/blob/main/src/schemas/address-schemas.ts#L11)) to the office schemas referenced in COD and COA
- Added unit tests

COA Before:
image

COA After:
image

COD Before:
image

COD After:
image

### DIFF
--- a/src/components/common/Directors.vue
+++ b/src/components/common/Directors.vue
@@ -727,6 +727,7 @@
 import { Component, Emit, Mixins, Prop, Vue, Watch } from 'vue-property-decorator'
 import axios from '@/axios-auth'
 import { Getter } from 'pinia-class'
+import { validatePostalCode } from '@/validators'
 import { required, maxLength } from 'vuelidate/lib/validators'
 import { cloneDeep, isEqual } from 'lodash'
 import { BaseAddress } from '@bcrs-shared-components/base-address'
@@ -858,7 +859,8 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
     },
     postalCode: {
       required,
-      maxLength: maxLength(15)
+      maxLength: maxLength(15),
+      validatePostalCode
     },
     deliveryInstructions: {
       maxLength: maxLength(80)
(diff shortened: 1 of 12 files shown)


src/validators.ts
```diff
@@ -63,3 +63,28 @@ export function isValidEffectiveDate (date: string, separator: string): boolean
   const c = date.split(separator).join('')
   return c <= d2
 }
+
```
severinbeauvais: The above are Vuetify validators but the below are Vuelidate schema custom validators. What do you think about keeping them separate?

severinbeauvais (approved):
Looks good, and thanks for the unit tests!

What do you think about moving the directors address schema to its own file (alongside office-address-schema)?

src/validators.ts
severinbeauvais: Per my review comment, it may be better to keep the Vuelidate validators in the same folder as the schemas.

loneil (approved):

src/components/common/Directors.vue
```diff
@@ -835,37 +834,8 @@ export default class Directors extends Mixins(CommonMixin, DateMixin, DirectorMi
...
-      required,
-      maxLength: maxLength(15),
-      validatePostalCode
-    },
-    deliveryInstructions: {
-      maxLength: maxLength(80)
-    }
-  }
+  /** The Address schema containing Vuelidate rules. */
+  directorAddressSchema = directorAddressSchema
```
severinbeauvais: I usually tag these as readonly and add a comment that this is for the template.

severinbeauvais (approved):
Still LGTM :)

bcregistry-sre:
Temporary Url for review: https://business-filings-dev--pr-784-eo3ke53o.web.app

loneil (approved):

